PR #23, opened 2016-05-23
Change Javadoc to mention SyncSubscription instead of AsyncSubscription.

zshamrock:
Change to Javadoc to mention SyncSubscription instead of AsyncSubscription. Also using {@link} instead of {@code}.


mcqueary:
Good catch, thank you!

PR #27, opened 2016-06-20
Fix error example code

jensenbox:
I could be wrong but I think this is the right fix here - it works for me at least.


pires:
While the tests don't pass because of non-related issues, I can confirm this is the correct API usage.

PR #28, opened 2016-06-20
Use new method instead of property

jensenbox:


pires:
Same comments as in #27.

PR #29, opened 2016-06-22
Revert to Java 1.7 compatibility

mcqueary:


pires:
@mcqueary why Java 7 compatibility? JRE 7 is EOL since April '15.

Anyway, I'd suggest to add [`maven-compiler-plugin` source/target properties](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html).

mcqueary:
@pires Android SDK compatibility issue. And I *am* using `maven-compiler-plugin` source/target properties (they are included via nats-parent-pom).

Issue #26, opened 2016-06-16
Closing NATS connection may not close all running threads


kevinsookocheff-wf:
Hi,

I'm working on getting a simplified reproducible example. Will update this issue when that is ready. 

The general problem is that closing a NATS connection may not close all threads if there is work being done in ConnectionImpl.deliverMsgs. I looks like if Channel.get() is called from deliverMsgs it calls q.take() (Channel.java:73), which blocks until data is available. I believe that calling interrupt() in Channel.close() will allow Channel.get() to unblock and exit correctly.

Below is a thread dump from `jstack` of an app that cannot exit because of hung NATS threads.

---
```
2016-06-16 12:47:13
Full thread dump Java HotSpot(TM) 64-Bit Server VM (25.60-b23 mixed mode):

"Attach Listener" #28 daemon prio=9 os_prio=31 tid=0x00007f830a001000 nid=0x3d0b waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"DestroyJavaVM" #27 prio=5 os_prio=31 tid=0x00007f830925b000 nid=0x1703 waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"msgfeeder" #17 prio=5 os_prio=31 tid=0x00007f830b083800 nid=0x6503 waiting on condition [0x0000700001b67000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x000000076c610fd0> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
	at java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
	at io.nats.client.Channel.get(Channel.java:73)
	at io.nats.client.Channel.get(Channel.java:56)
	at io.nats.client.ConnectionImpl.deliverMsgs(ConnectionImpl.java:1404)
	at io.nats.client.AsyncSubscriptionImpl$1.run(AsyncSubscriptionImpl.java:94)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #31, opened 2016-07-13
Override unsubscribe in AsyncSubscriptionImpl

tylertreat:
Currently, AsyncSubscriptionImpl leaks the msgfeeder thread on
unsubscribe because Channel.get() is never unblocked. This
overrides unsubscribe so that the thread properly terminates.
I believe this resolves issue #26.

FYI @stevenosborne-wf


tylertreat-wf:
AFAICT the build failure is not related to this code change.

mcqueary:
@tylertreat-wf The build failure is due to some deployment stuff I put in Travis need to be fixed and exclude pull requests.

Thanks!

Issue #93, opened 2016-12-26
Scatter gather pattern support


Hronom:
What need to be done:
1. Create reply topic and subscribe
2. Send query on broadcast topic with reply address
3. Listen for responses and unsubscribe after a time limit
4. Aggregate response and return result

Thanks to the @mhaan from Slack, it looks like there is not much changes need to be done to support this feature.

sasbury:
Closing for 1.x, but open to a new PR for 2.x or possibly a doc example for nats.io.

PR #94, opened 2016-12-27
Added methods for Scatter-Gather pattern

Hronom:
Added methods for Scatter-Gather pattern basic realization #93

### DIFF
--- a/src/it/java/io/nats/client/ITBasicTest.java
+++ b/src/it/java/io/nats/client/ITBasicTest.java
@@ -44,6 +44,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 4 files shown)


ColinSullivan1:
@Hronom , Thank you for the contribution!  We apologize for not getting back to you sooner.

We generally have the Apcera supported client APIs follow our canonical client, the go client, unless there is something idiomatic with the language we feel we should support.  Following the tenet of NATS simplicity, we're very hesitant to add new APIs.

This functionality seems to cover a use case that has come up more than a few times.  That being said, this may be better served as an example pattern we can showcase elsewhere.  We've had discussions about providing a repository of examples using different messaging patterns; this could be a good candidate for that.

I'll let you know.

Thanks,
Colin

ColinSullivan1:
@Hronom , we'll demonstrate this pattern in some example code.

sasbury:
This PR has gotten out of date, if you have some ideas for the 2.x branch please update and submit those. I dislike closing peoples work, but things changed too much to use this version.

Hronom:
After 3 years... well we don't use NATS for now, so info for community: I don't have time to renew code and make pull request again, sorry.

PR #101, opened 2017-02-09
Fix flush timing

mcqueary:
Resets the `flushTimerUnit` to `MILLISECONDS` and `flushTimerInterval` to `1` for best throughput/#flush tradeoff.


kozlovic:
LGTM. This is also what I do in the NATS C client. Without this delay, C and probably Java would immediately flush resulting in poor throughput perf. The Go client is different because I suspect that that go routine is not scheduled right away.

PR #106, opened 2017-04-20
Catch IOException on output stream when force flushing socket

### LINKED ISSUE #105: SocketException on publish with forceKick=true while reconnecting
Looks like when you call publish with force flush, the "kick the flusher" is not protected with a try/catch and ends up raising a `SocketException` error to the caller. Is this behavior considered correct or incorrect? Seems like a bug to me, but I was just wondering what you guys thought.

I found this while testing failover scenarios where I have two gnatsd servers in a cluster with only one running.

1. I connect java-nats to the running gnatsd server.
2. I begin sending publish in a loop.
3. I start the second gnatsd server and stop the first server.
4. A `SocketException` is raised to the caller where I expected no error to be raised.

The offending code seems to be here: https://github.com/nats-io/java-nats/blob/b320df5d551d6df7d8f445c368218fe737f126b4/src/main/java/io/nats/client/ConnectionImpl.java#L2006-L2009. We are attempting to flush the socket's output stream when the socket is closed.

film42:
Fixes #105 

I don't code a lot in java, so please let me know if I can make this more idiomatic. Thanks!

### DIFF
--- a/src/main/java/io/nats/client/ConnectionImpl.java
+++ b/src/main/java/io/nats/client/ConnectionImpl.java
@@ -1721,7 +1721,7 @@ class ConnectionImpl implements Connection {
 
     /*
      * (non-Javadoc)
-     * 
+     *
      * @see io.nats.client.AbstractConnection#flush(int)
      */
     @Override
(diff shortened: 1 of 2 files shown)


film42:
Looks like travis is failing while trying to get your deploy key.

```
$ openssl aes-256-cbc -K $encrypted_d472159ded8f_key -iv $encrypted_d472159ded8f_iv -in .travis/keyrings/deploy_key.pem.enc -out .travis/keyrings/deploy_key.pem -d
iv undefined
```

ColinSullivan1:
@film42 , this fix looks good to me.  I'll look at Travis and see what is going on.  You contribution is much appreciated!

ColinSullivan1:
LGTM; thank you for the contribution!

Issue #114, opened 2017-05-11
Support large numbers of subscribers


ColinSullivan1:
In order to support large numbers (thousands) of asynchronous subscribers in an application, using one thread per subscriber to handle message events requires more resources than are typically available. An option should be provided that will setup a NATS connection to process all subscriber message events in a pool of threads, providing a choice of scalability over performance.

This will address #113 

Related Issues/PRs in other clients:
https://github.com/nats-io/csharp-nats/issues/122
https://github.com/nats-io/csharp-nats/pull/124
https://github.com/nats-io/cnats/pull/34

PR #115, opened 2017-05-15
Remove slf4j dependency

ColinSullivan1:
This PR is basic housekeeping to keep the client libraries trim - we do not need to log.  Throwing exceptions (and in rare cases writing to stdout/stderr) should suffice.


src/main/java/io/nats/client/NatsThread.java
```diff
@@ -33,7 +29,7 @@ public NatsThread(Runnable runnable, String name) {
         super(runnable, name + "-" + created.incrementAndGet());
         setUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
             public void uncaughtException(Thread t, Throwable e) {
-                logger.debug("UNCAUGHT in thread {}", t.getName(), e);
+               e.printStackTrace();
```
petemiron: why not throw this? instead of printing?

ColinSullivan1: The exception was swallowed/discarded to begin with, and I don't want to change behavior too much at this point.

petemiron: fair. not sure of the right answer here esp. since this is being thrown on a new thread. i guess stick with this and we'll see if we get feedback, revisit as we improve/simplify the rest of the code.

ColinSullivan1: It depends on the context/usage of the NATS thread intended by the caller, but I think most of the time we'll want to keep the thread alive.  I'm also wondering if the NATS thread class is really required.

There are a number of places in the code I'd like to revisit; I've added a few TODOs.

tylertreat: This is kinda unrelated, but one issue we had with the Java NATS client at WK was how it started a thread per subscription, which was super resource-heavy for some use cases we had. Would be nice if we could get away from managing threads ourself as much as possible and make subs more lightweight.

kozlovic: In Go one can use chan subscribers where you provide your own channel. In C, I added that feature (a global dispatcher). I believe Colin did the same for the C# client.

aricart: Am I missing something here? Looks like the whole reason for this class is so that the API knows how many were created. To address @tylertreat comment, likely we just want to make use an executor, and just run the Runnables. It could be that this value could be tweaked to get more workers processing the runnables. Not sure what the intention was.

ColinSullivan1:
@tylertreat @kozlovic , Addressing the threading constraints  it is on the list (#114) to address before the next release.  I was planning on using the best thread pool mechanism available that will work for older versions of java,  TBD.

@aricart , Yes, this appears to be the reason behind the class, but I do think we can move toward standard java threads rather than using this class.  Less is more in NATS.  :)

Issue #119, opened 2017-06-05
Code in Basic Usage section of README.md doesn't compile against version 1.0


grange74:
I was just trying out this library and noticed a few compilation issues with the code in the Basic Usage when using against version 1.0 of the library. 

Here is what i changed to get my code working:

- The "Simple Sync Subscriber" code should use the `SyncSubscription` class instead of generic `Subscription` and the `nextMsg` method is now `nextMessage`
```
// Simple Sync Subscriber
SyncSubscription sub = nc.subscribeSync("foo");
Message m = sub.nextMessage(1, TimeUnit.SECONDS);
```

The "Requests and Replies" code should take in bytes[] instead of String for the `request` and `publish` methods.
```
// Requests
msg = nc.request("help", "help me".getBytes(), 10000);

// Replies
nc.subscribe("help", m -> {
    nc.publish(m.getReplyTo(), "I can help!".getBytes());
});
```

The "Close connection" code should use `Connection` instead of `Connect`
```
// Close connection
Connection nc = Nats.connect("nats://localhost:4222");
...
nc.close();
```

sasbury:
Version 2 updated to new examples, all the code is taken from code examples and should compile. What I really want is to have auto-clipping with links, but no time for that yet ;-)

Let us know if you find any more problems!

PR #120, opened 2017-06-19
updated based on suggestions from @grange74

petemiron:
@nats-io/core I'm putting these examples into a currently private repo (java-nats-examples), where they will be compilable and testable. I'll add issues for the remainder of the java examples to be added there.


petemiron:
Fixes #119

PR #122, opened 2017-06-29
Clarify the reply example in README.md a bit

toukovk:
Parameter name reply was IMO nonoptimal since that parameter contains the request message.


tylertreat:
lgtm

PR #131, opened 2017-09-28
Configurable async subscriber threading model

timboudreau:
Addresses nats-io/java-nats#113 - configurable concurrency for async subscription dispatch, as follows:

 * Adds `Options.Builder.subscriptionDispatchPool ( ExecutorService )`
 * Adds `Options.getSubscriptionDispatchPool()`
 * Adds `Nats.PROP_SUBSCRIPTION_CONCURRENCY = "subscription.concurrency"`

with the following behavior:

1.  If no executor and subscription concurrency not set, behavior is same as before
2.  If subscription concurrency is set, `java.util.concurrent.Executors.newFixedThreadPool()` is called with the property value
3.  If an `ExecutorService` was set in the connection options that is used;  if both that and the subscription concurrency property are set, the provided `ExecutorService` wins.

All tests pass with and without the patch;  I left it on in `ConnectionImplTest`.

One thing that is not obvious is whether the `Runnable` that invokes the subscriber needs to take `mu.lock()` - to err on the side of correctness I did, but probably better liveness can be had if that can be avoided, since it seems to be used everywhere.  A better variable name than `mu` might help contributors figure out *what* it locks :-)

### DIFF
--- a/src/main/java/io/nats/client/ConnectionImpl.java
+++ b/src/main/java/io/nats/client/ConnectionImpl.java
@@ -25,7 +25,6 @@ import static io.nats.client.Nats.ERR_SECURE_CONN_REQUIRED;
 import static io.nats.client.Nats.ERR_SECURE_CONN_WANTED;
 import static io.nats.client.Nats.ERR_SLOW_CONSUMER;
 import static io.nats.client.Nats.ERR_STALE_CONNECTION;
-import static io.nats.client.Nats.ERR_TCP_FLUSH_FAILED;
 import static io.nats.client.Nats.ERR_TIMEOUT;
 import static io.nats.client.Nats.PERMISSIONS_ERR;
 import static io.nats.client.Nats.TLS_SCHEME;
@@ -56,6 +55,7 @@ import java.util.Random;
 import java.util.TimerTask;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
@@ -92,6 +92,11 @@ class ConnectionImpl implements Connection {
 
     // The buffered size of the flush "kick" channel
     protected static final int FLUSH_CHAN_SIZE = 1;
+
+    static final CompletableFuture<Void> COMPLETED = new CompletableFuture();
+    static {
+        COMPLETED.complete(null);
+    }
 
     // The interval the flusher will wait after each flush of the buffered output stream
     private long flushTimerInterval = 1;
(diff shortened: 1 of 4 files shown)


kozlovic:
@timboudreau Thank you for contributing. I have updated https://github.com/nats-io/java-nats/issues/113 with a list of requirements that were not properly stated. Apologies for that.

I see several issues with this approach (which again requirements are now stated in the issue):

- The goal of having a pool was to avoid creation of a thread per subscription. As far as I can tell from your changes, we still create the thread (waitForMsgs) for each subscription. So I don't see the point of creating a runnable to just deliver the message to that thread. 
- It does not guarantee that messages from the same subscription will be dispatched in the incoming order. This is something that the NATS client must guarantee. If a user does not mind the ordering it is possible to then dispatch from the message callback to yet another execution pool, but the library should not do that.
- With this change an incoming message is handed off to a runnable that then access the connection/sub. It is possible that the connection has been closed by then. The current code would not know that.

Aside from that, you are right about the locking and we have already updated some of the supported clients to use a specific lock for the subscription map. For the Go client, the PRs are those:
https://github.com/nats-io/go-nats/pull/285
https://github.com/nats-io/go-nats/pull/286
But this should be a separate PR.

timboudreau:
Ok, I was going off of @tylertreat 's comment from Aug 22 that the quick and dirty branch someone proposed was actually on the right track, giving the impression that its intervention was in the right place.

Looks like what's needed is something akin to a queue that maintains a mapping of subscriberId : listOfMessages where the list of messages is atomically copied and cleared when polling, and a delivery loop that polls for an object containing subscriber id and the list.

kozlovic:
Issue #114 references implementations done in C# and C. In C, I have a library level thread pool that deals with message delivery. When a subscription is created, I pick one of the pool's queue. When a message is processed (in `processMsg`), we lookup the subscription, then get the pool's queue and schedule the message to that queue. The limit check can still be done in `processMsg` but with the pool's message list instead of the subscription list. The pool's thread has the same logic than `waitForMsgs`: it waits to be notified that messages are avail. Since the sub is added to the message, we can get the sub from the dequeued message, then check sub's closed status and invoke callback, etc..

The commit https://github.com/nats-io/cnats/pull/34/commits/d1af4fa198f0f107ba232b1beb5bca9c2fb1d377 for the C client may help you understand what I am trying to describe above. Note that because it's C, there is much more work involved because of reference counting, etc..

timboudreau:
BTW, thanks for the info and c impl - it was helpful, and I wound up doing something similar in the second pull request (the only difference being not permanently tying a queue to a particular thread, but keeping a mapping of thread to subscriber id that the consumer clears in a finally block once it consumes a batch of messages, so no other thread can deliver messages for that subscription until the current batch is processed).

I timed some of the functional tests, and noticed that the result (in the newer pull request) outperforms thread-per-subscriber, which it should not - and read your blog about cleaning up locking in the c implementation.  I suspect the Java implementation could benefit from similar cleanup - I was able to do all but a tiny, rarely called corner of the subscriber pool with no locks, just atomic primitives from `java.util.concurrent.atomic`.  The locking in SubscriberImpl and ConnectionImpl is pretty draconian - nothing wrong with that as a way to write something that's guaranteed not to surprise you, but it also impacts liveness.

kozlovic:
We are closing this PR since the feature has been implemented through #135.

PR #132, opened 2017-09-29
Round 2 of pooled delivery of asynchronous subscriptions

timboudreau:
The following is a new patch to address nats-io/jnats#113 which makes the same API changes as the preceding one (ability to pass an `ExecutorService` or set subscription.concurrency property or both), with the following changes:

 * The subscription.concurrency property is used to determine how many runnables are dispatched to the thread pool to handle subscription messages.  If the property is not set, then:
   * If the passed `ExecutorService` is an instanceof `ThreadPoolExecutor` then `getCorePoolSize()` is used
   * If not, `Runtime.getRuntime().availableProcessors()` is used
 * Introduces `AsyncDispatchQueue`, which uses atomic operations on `ConcurrentHashMap` to manage the mapping of subscription ids to individual queues, and individual queues which use a straightforward linked-list structure - uses as minimal locking as I could devise and have the result be correct
   * Note this means the queue held by `SubscriptionImpl` is not used *at all* - having the cache/queue know only about longs and messages is a cleaner and much more leak-proof approach than having it fiddle with subscription implementation classes
   * See the fairly exhaustive unit tests

I believe this meets all the requirements listed in the bug;  I've attempted to comment the code fairly thoroughly since it visits a few infrequently visited corners of java.util.concurrent.

Tests pass - I've set up `ConnectionImplTest` to enable the feature by default (long term, it might be a good idea to run async subscription tests twice, once with and once without this behavior).

#### Notes

Removed a few lines from tests where the thing tested is not relevant - in particular, tests for calls to `setSlowConsumer()` for asynchronous subscriptions:  In fact, at runtime that would never have been called before or now, because it is only triggered if the `ch` argument to `ConnectionImpl.subscribe()`, a `LinkedBlockingQueue` with a limit was passed in (the default is unbounded), and the original code only sets the channel in the case of a synchronous consumer, so the passed-in queue was ignored for async subscriptions already.  So nothing is being lost that ever did anything - it should probably have always been an error to pass in a queue if you are passing in a `MessageHandler`, since the queue was and is ignored in that case.  For that matter, `AsyncSubscription` should probably not have a queue - a bigger but doable project would be to factor out an abstract superclass for `SubscriptionImpl` and AsyncSubscriptionImpl, so there aren't impedance mismatches.

Dealing with exceptions thrown by `MessageHandler`s:  The original code simply aborts the subscriber thread - no further messages will be delivered.  With a shared thread pool, this would impact all subscriptions.  What I did is pass it to the async error callback if one is present, and throw an exception in `subscribe()` if you don't have one and pooled subscription handling is on.

One other note on that:  The original code handles `InterruptedException` (which might be a hint that the thread pool is shutting down) by setting the thread's interrupted flag and exiting.  Setting the interrupted flag on a thread that is about to terminate does nothing;  and this is a thread that invokes client code - the InterruptedException could be thrown from there, in which case exiting isn't actually the right thing.  I kept compatibility with that, but a saner thing to do would be to test the thread pool's `isShutdown()` and decide to exit or not based on that.

### Testing

So, after pulling another all-nighter on this, I run it and the tests pass.  Yay!  That seems a little too good to be true, ao I instrument a few bits of code to see just how extensively the new code is being touched:  Almost not at all.  Two times in the entire suite - after altering `UnitTestUtilities` to *always* use a thread pool for delivery, and finding any other cases that involve a `MessageHandler` and altering them.

So does it work?  I don't know - it seems like there's a lot of mocks creating the illusion of test coverage here while testing far less than it appears.  What tests there are pass, but those that actually touch asynchronous subscriptions are trivial.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ hs_err_pid*
 # NATS stuff
 gnatsd.log
 *.csv
+/nbproject/
+
--- a/pom.xml
+++ b/pom.xml
@@ -268,6 +268,7 @@
                     <!-- Skips unit tests if the value of skip.unit.tests 
                         property is true -->
                     <skipTests>${skip.unit.tests}</skipTests>
+                    <trimStackTrace>false</trimStackTrace>
                     <!-- Excludes integration tests when unit tests are run. -->
                     <excludes>
                         <exclude>**/IT*.java</exclude>
@@ -306,6 +307,7 @@
                             <goal>verify</goal>
                         </goals>
                         <configuration>
+                            <trimStackTrace>false</trimStackTrace>
                             <!-- Sets the VM argument line used when integration 
                                 tests are run. -->
                             <argLine>${failsafeArgLine}</argLine>
--- a/pom.xml
+++ b/pom.xml
@@ -272,6 +272,11 @@
                     <excludes>
                         <exclude>**/IT*.java</exclude>
                     </excludes>
+                    <systemPropertyVariables>
+                        <forkNumber>${surefire.forkNumber}</forkNumber>
+                        <unit.test>true</unit.test>
+                    </systemPropertyVariables>
+                    
                 </configuration>
                 <executions>
                     <execution>
--- a/src/it/java/io/nats/client/ITAsyncDeliveryTest.java
+++ b/src/it/java/io/nats/client/ITAsyncDeliveryTest.java
@@ -72,15 +72,25 @@ public class ITAsyncDeliveryTest {
     public void tearDown() throws Exception {
     }
 
+    private static final boolean LOG = Boolean.getBoolean("test.log");
+    static void log(String s) {
+        if (LOG) {
+            System.out.println(s);
+        }
+    }
+
     @Test(timeout = 480000)
     public void testThreadPerSubscription() throws Throwable {
         boolean batched = false;
-        System.out.println("\n\n\n**************** testThreadPerSubscription *********************\n\n\n");
-        EH errors = new EH();
+        log("\n\n\n**************** testThreadPerSubscription *********************\n\n\n");
+        TestExceptionHandler errors = new TestExceptionHandler();
         try (NatsServer srv = runDefaultServer()) {
             try (ConnectionImpl conn = (ConnectionImpl) connection(batched, errors)) {
                 String msgPrefix = batched ? "batched-mode " : "thread-per-subscriber ";
+                long start = System.currentTimeMillis();
                 testAsyncDelivery(conn, msgPrefix, errors);
+                long elapsed = System.currentTimeMillis() - start;
+                log("Completed in " + elapsed + "ms");
                 errors.rethrow();
             }
         }
@@ -89,52 +99,18 @@ public class ITAsyncDeliveryTest {
     @Test(timeout = 480000)
     public void testBatchedSubscriptions() throws Throwable {
         boolean batched = true;
-        System.out.println("\n\n\n**************** testBatchedSubscriptions *********************\n\n\n");
-        EH errors = new EH();
+        log("\n\n\n**************** testBatchedSubscriptions *********************\n\n\n");
+        TestExceptionHandler errors = new TestExceptionHandler();
         try (NatsServer srv = runDefaultServer()) {
             try (ConnectionImpl conn = (ConnectionImpl) connection(batched, errors)) {
                 String msgPrefix = batched ? "batched-mode " : "thread-per-subscriber ";
+                long start = System.currentTimeMillis();
                 testAsyncDelivery(conn, msgPrefix, errors);
+                long elapsed = System.currentTimeMillis() - start;
+                log("Completed in " + elapsed + "ms");
                 errors.rethrow();
             }
         }
-    }
-
-    static class EH implements ExceptionHandler {
-
-        Throwable thrown = null;
-
-        synchronized void rethrow() throws Throwable {
-            if (thrown != null) {
-                throw thrown;
-            }
-        }
-
-        volatile Thread testThread;
-
-        void pickUpTestThread() {
-            testThread = Thread.currentThread();
-        }
-
-        @Override
-        public synchronized void onException(NATSException ex) {
-            Throwable t = ex;
-            while (t.getCause() != null) {
-                t = t.getCause();
-                if (t.getCause() == null) {
-                    break;
-                }
-            }
-            if (thrown != null) {
-                thrown.addSuppressed(t);
-            } else {
-                thrown = t;
-            }
-            if (testThread != null) {
-                testThread.interrupt();
-            }
-        }
-
     }
 
     static Connection connection(boolean batched, ExceptionHandler errors) throws IOException {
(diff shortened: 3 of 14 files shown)


timboudreau:
Minor update:  I've added a fairly exhaustive functional test with multiple async subscribers, which passes for both thread-per-subscriber and pooled, and updated the other functional tests (most did not use an `ExceptionHandler` which is required for pooled async subscriptions).

Performance with pooled subscriptions is slightly better than thread-per-subscriber, probably due to the absence of lock contention.

`ITSubscriptionTest.testAsyncErrHandler()` fails with pooled message delivery on, but that's a problem with the assumptions the test makes (it changes the limit after a batch of messages has begun to be delivered and any messages that would overflow the maximum were removed, which seems like reasonable behavior).

I see a few other failures I can't explain relating to disconnect/reconnect stats, which aren't related to code I've touched in any obvious way.  Suggestions here would be welcome.

ColinSullivan1:
@timboudreau , excellent contribution, and very thorough tests, much appreciated! 

The optimizations you have in the AsyncMsgQueue look really nice, particularly the atomic references.  I was wondering though what advantages mapping the subscriber ID to a thread would have over the approach we took in C and C# in directly assigning a shared channel (and thus a thread) to a subscriber at creation time?  In C# that approach resulted in a much smaller codebase.

timboudreau:
Simple:  With the approach you took in C, say that consumer *A* and *B* both use thread *T1*.  Messages arrive for consumers *A* and *B*.  *A* is slow.  *B*'s messages are delayed until *A*'s messages are delivered, and will be *every time* there are messages for both.

By making thread affinity temporary rather than permanent, message delivery can use all the parallelism available.  *A* may still slow down delivery of *some* messages for *some* consumers, but you don't wind up with a "sad consumer" situation where whatever consumers are unlucky enough to wind up assigned to the same thread as a slow-poke will be impacted by it every time.

kozlovic:
@timboudreau An issue I have found with your approach, and you eluded to that, is that you do not honor the max messages. This can be seen with cases where subscription is auto-unsubscribed() or simply unsubscribed (or connection closed) from the callback. I have added some tests that check for that and it would fail in your branch.

The PR #135 is simpler (there are a lot of test files changed, but that's so that they extend a based class and then remove from code). The pool is global - from your PR, I can't really tell if it is per connection or still shared.

You are right about the fact that in PR #135, it is possible that messages from a slow subscription will impact messages from other subscriptions. If the user knows that it is a possibility that a subscription is going to be slow, a dedicated connection that does not use the pool can be used for that conn/subscription (or vice versa if you want a given subscription not to be impacted by others).

From your PR it is difficult for me to understand if we are guaranteed that no message of a given subscription are not possibly dispatched by different threads. The async queue is dequeued and then messages dispatched. I see that you comment that calling batch.close() ensure that this does not happen. But then I could argue that it is still then possible that a fast subscription has its messages added to the queue and will only be dispatched by the same thread (to guarantee ordering) that may then be slowed down by previous slow subscription.

Also, I have fixed the fact that `waitForMsgs` should catch any throwable that could be thrown by user code and simply ignore and move to the next. Anytime a library invokes a user callback, it has to protect itself. I have not gone to the extent of forcing the user to specify and error callback though.
The new code also ensures that exception thrown in `onMessage()` does not cause the work thread to exit.

kozlovic:
@timboudreau Regarding tests that would fail in your branch: these are the [2 ones](https://github.com/nats-io/java-nats/pull/135/commits/3947cca68de57bb49c06dcb134c6fa11994d7d67) that I have added to PR #135 that you could adapt to your branch. When I did, they would fail...

timboudreau:
Okay, this patch is not a baby that I carried for 9 months - I care much more about having the feature than whose patch wins.

> An issue I have found with your approach, and you eluded to that, is that you do not honor the max messages.

I *do* honor max messages, with a tiny caveat:  We collect a *batch* of messages - that may be more than one, and before calling the MessageHandler, remove any messages at the tail of the list that would put cause it to deliver above the max messages or byte count.  What I don't honor is *changing the maximums value while a batch of messages is being delivered* - so in the remote corner case that something changes that value within the few milliseconds the list is being iterated and the MessageHandler is being called, you are correct.  The upside of this approach is we *lock once* on the subscriber for many messages, which has performance implications.  It seems like a worthwhile trade off, as, outside of unit tests, that's a pretty unlikely situation - real code is far more likely to set that at initialization time.

> From your PR it is difficult for me to understand if we are guaranteed that no message of a given subscription are not possibly dispatched by different threads

First, they *are* dispatched by different threads, but never two threads at once - any messages that arrive for subscriber N before the previous batch is delivered will not be picked up by any thread until the batch being delivered is closed.

Let's walk through a little code - this is `AsyncDeliveryQueue.drain()`, which is the place that would happen:

```java
    private int drain(List<? super MessageBatch> batch) {
        int result = 0;
        boolean resignal = false;
        Thread currentThread = Thread.currentThread();
        int as = arrays.size();
        final long[] totalBytes = new long[1];
        for (int i = 0; i < as; i++) {
            AtomicReferenceArray<Messages> arr = arrays.get(i);
```
This is where we start iterating subscriber ids, looking for messages we can deliver.

```java
            for (int j = 0; j < arraySize; j++) {
                long subscriberId = (long) (i * arraySize) + j;
                boolean alreadyClaimed = !threads.get(i).compareAndSet(j, 0, currentThread.getId());
```
In most cases, `alreadyClaimed` will cause two threads in this method to abort.  The fact that it's atomic guarantees that two threads will not both see a value of `false` here, *BUT* ...

```
                // Ensure that only this thread can read messages for this key, until
                // the batch for that key is closed.
                if (alreadyClaimed) {
                    // Some other thread is currently dispatching messages for
                    // this subscriber, so we won't take any
                    // But note it and signal another thread (would be nicer to have a
                    // per-long condition so we could signal the exact thread, but you can't have
                    // atomic operations across two maps without a lock)
                    Messages m = arr.get(j);
                    resignal |= m != null && !m.isEmpty();
                    continue;
                }
                Messages m = arr.get(j);
                totalBytes[0] = 0L;
```
...if it *were* possible for two threads to get here, `m.drain()` is atomically flipping the tail of the queue to null.  One thread will get the messages, the other will get an empty array.

```java
                List<Message> msgs = m == null ? Collections.<Message>emptyList() : m.drain(totalBytes);
                if (!msgs.isEmpty()) {
                    result += msgs.size();
                    batch.add(new MessageBatchImpl(subscriberId, msgs, currentThread, totalBytes[0]));
                } else {
                    threads.get(i).compareAndSet(j, currentThread.getId(), 0);
                }
            }
        }
        // Note that resignal is best-effort - we cannot guarantee that another
        // thread did not add something not reflected in messagesForSubscriber.entrySet
        // while we were iterating it.  Hence the busywait rather than full wait
        // in wait()
        if (resignal) {
            signalOneThread();
        }
        return result;
    }
```

> But then I could argue that it is still then possible that a fast subscription has its messages added to the queue and will only be dispatched by the same thread (to guarantee ordering) that may then be slowed down by previous slow subscription.

Sure.  The point here is that the same subscriber IDs are not *always* throttled by the slow ones.  Say we have slow subscriber X, which is just fast enough not to get taken out of service, and gets constant throughput at just the rate it can handle.  Behind it, put subscriber Y, which is getting, say, thousands of messages per second.  The odds of an OutOfMemoryError are considerably higher if messages for Y are always cached in memory behind messages for X, than if you distribute the latency caused by X across *all* subscribers.

ColinSullivan1:
@timboudreau , thank you for the quality work here, but we went with an alternative approach, https://github.com/nats-io/java-nats/pull/135.  While we did not go with your solution, your code quality and depth of thought here is valued.  The time you've committed to NATS is very much appreciated, and we hope you submit additional PRs in the future.

PR #135, opened 2017-10-04
[ADDED] Global Message Delivery Thread Pool

### LINKED ISSUE #114: Support large numbers of subscribers

kozlovic:
This PR adds the ability to set a library level thread pool
responsible for dispatching asynchronous subscriptions' messages.
All messages from a given subscription will be delivered in
the order they arrived.

This also maintains the contract that a subscription reaching its
auto-unsubscribe max or being unsubscribed (or the connection
being closed) in the message callback prevents subsequent delivery
of messages to that subscription (even if they were already received
by the connection).

A connection option allows the user to force the subscription
to use its own thread to dispatch its message even if the
library pool has been set.

For the new code and fixing the current behavior of `waitForMsgs` that
should catch any throwable issued by the user callback and continue 
to work.

Resolves #114

### DIFF
--- a/src/main/java/io/nats/client/ConnectionImpl.java
+++ b/src/main/java/io/nats/client/ConnectionImpl.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2015-2016 Apcera Inc. All rights reserved. This program and the accompanying
+ *  Copyright (c) 2015-2017 Apcera Inc. All rights reserved. This program and the accompanying
  *  materials are made available under the terms of the MIT License (MIT) which accompanies this
  *  distribution, and is available at http://opensource.org/licenses/MIT
  */
@@ -25,7 +25,6 @@ import static io.nats.client.Nats.ERR_SECURE_CONN_REQUIRED;
 import static io.nats.client.Nats.ERR_SECURE_CONN_WANTED;
 import static io.nats.client.Nats.ERR_SLOW_CONSUMER;
 import static io.nats.client.Nats.ERR_STALE_CONNECTION;
-import static io.nats.client.Nats.ERR_TCP_FLUSH_FAILED;
 import static io.nats.client.Nats.ERR_TIMEOUT;
 import static io.nats.client.Nats.PERMISSIONS_ERR;
 import static io.nats.client.Nats.TLS_SCHEME;
@@ -1417,7 +1416,8 @@ class ConnectionImpl implements Connection {
             }
             // Deliver the message.
             if (msg != null && (max <= 0 || delivered <= max)) {
-                mcb.onMessage(msg);
+                // Ignore any error thrown by the user
+                try { mcb.onMessage(msg); } catch (Throwable t) {}
             }
             // If we have hit the max for delivered msgs, remove sub.
             if (max > 0 && delivered >= max) {
@@ -1458,7 +1458,12 @@ class ConnectionImpl implements Connection {
             // It's possible that we end up not using the message, but that's ok.
             Message msg = new Message(parser.ps.ma, sub, data, offset, length);
 
-            sub.lock();
+            final MsgDeliveryWorker mdw = sub.getDeliveryWorker();
+            if (mdw != null) {
+                mdw.lock();
+            } else {
+                sub.lock();
+            }
             try {
                 sub.pMsgs++;
                 if (sub.pMsgs > sub.pMsgsMax) {
@@ -1474,19 +1479,27 @@ class ConnectionImpl implements Connection {
                         || (sub.pBytesLimit > 0 && sub.pBytes > sub.pBytesLimit)) {
                     handleSlowConsumer(sub, msg);
                 } else {
-                    // We use mch for everything, unlike Go client
-                    if (sub.getChannel() != null) {
-                        if (sub.getChannel().add(msg)) {
-                            sub.pCond.signal();
-                            // Clear Slow Consumer status
-                            sub.setSlowConsumer(false);
-                        } else {
-                            handleSlowConsumer(sub, msg);
+                    if (mdw != null) {
+                        mdw.postMsg(msg);
+                    } else {
+                        // We use mch for everything, unlike Go client
+                        if (sub.getChannel() != null) {
+                            if (sub.getChannel().add(msg)) {
+                                sub.pCond.signal();
+                                // Clear Slow Consumer status
+                                sub.setSlowConsumer(false);
+                            } else {
+                                handleSlowConsumer(sub, msg);
+                            }
                         }
                     }
                 }
             } finally {
-                sub.unlock();
+                if (mdw != null) {
+                    mdw.unlock();
+                } else {
+                    sub.unlock();
+                }
             }
         } finally {
             mu.unlock();
@@ -1811,18 +1824,27 @@ class ConnectionImpl implements Connection {
             }
 
             if (cb != null) {
-                sub = new AsyncSubscriptionImpl(this, subject, queue, cb);
-                // If we have an async callback, start up a sub specific Runnable to deliver the
-                // messages
-                subexec.submit(new Runnable() {
-                    public void run() {
-                        try {
-                            waitForMsgs((AsyncSubscriptionImpl) sub);
-                        } catch (InterruptedException e) {
-                            Thread.currentThread().interrupt();
+                // Check if we should use the global message delivery pool or create our own thread.
+                MsgDeliveryPool msgDlvPool = null;
+                final boolean   useDlvPool = (this.opts.useGlobalMsgDelivery &&
+                                                ((msgDlvPool = Nats.getMsgDeliveryThreadPool()) != null));
+
+                sub = new AsyncSubscriptionImpl(this, subject, queue, cb, useDlvPool);
+                if (useDlvPool) {
+                    msgDlvPool.assignDeliveryWorker(sub);
+                } else {
+                    // If we have an async callback, start up a sub specific Runnable to deliver the
+                    // messages
+                    subexec.submit(new Runnable() {
+                        public void run() {
+                            try {
+                                waitForMsgs((AsyncSubscriptionImpl) sub);
+                            } catch (InterruptedException e) {
+                                Thread.currentThread().interrupt();
+                            }
                         }
-                    }
-                });
+                    });
+                }
             } else {
                 sub = new SyncSubscriptionImpl(this, subject, queue);
                 sub.setChannel(ch);
@@ -1856,24 +1878,24 @@ class ConnectionImpl implements Connection {
 
     @Override
     public AsyncSubscription subscribe(String subject, MessageHandler cb) {
-        return subscribe(subject, null, cb);
+        return (AsyncSubscription) subscribe(subject, null, cb);
     }
 
     @Override
     public AsyncSubscription subscribe(String subj, String queue, MessageHandler cb) {
-        return (AsyncSubscriptionImpl) subscribe(subj, queue, cb, null);
+        return (AsyncSubscription) subscribe(subj, queue, cb, null);
     }
 
     @Override
     @Deprecated
     public AsyncSubscription subscribeAsync(String subject, String queue, MessageHandler cb) {
-        return (AsyncSubscriptionImpl) subscribe(subject, queue, cb, null);
+        return (AsyncSubscription) subscribe(subject, queue, cb, null);
     }
 
     @Override
     @Deprecated
     public AsyncSubscription subscribeAsync(String subj, MessageHandler cb) {
-        return subscribe(subj, null, cb);
+        return (AsyncSubscription) subscribe(subj, null, cb);
     }
 
     private void addSubscription(SubscriptionImpl sub) {
--- a/src/main/java/io/nats/client/ConnectionImpl.java
+++ b/src/main/java/io/nats/client/ConnectionImpl.java
@@ -1377,9 +1377,8 @@ class ConnectionImpl implements Connection {
      * deliver messages to asynchronous subscribers.
      *
      * @param sub the asynchronous subscriber
-     * @throws InterruptedException if the thread is interrupted
      */
-    void waitForMsgs(AsyncSubscriptionImpl sub) throws InterruptedException {
+    void waitForMsgs(AsyncSubscriptionImpl sub) {
         boolean closed;
         long delivered = 0L;
         long max;
@@ -1392,7 +1391,7 @@ class ConnectionImpl implements Connection {
             try {
                 mch = sub.getChannel();
                 while (mch.size() == 0 && !sub.isClosed()) {
-                    sub.pCond.await();
+                    try { sub.pCond.await(); } catch (InterruptedException e) {}
                 }
                 msg = mch.poll();
                 if (msg != null) {
@@ -1840,11 +1839,7 @@ class ConnectionImpl implements Connection {
                     // messages
                     subexec.submit(new Runnable() {
                         public void run() {
-                            try {
-                                waitForMsgs((AsyncSubscriptionImpl) sub);
-                            } catch (InterruptedException e) {
-                                Thread.currentThread().interrupt();
-                            }
+                            waitForMsgs((AsyncSubscriptionImpl) sub);
                         }
                     });
                 }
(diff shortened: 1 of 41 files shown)


kozlovic:
@ColinSullivan1 @tylertreat This PR looks bigger than what it is. I had to update the tests and made the integration and unit test extend a base class that implements the `BeforeClass`, etc... So lots of files touched but actually removing code since now they extend a base class.

The new file is `MsgDeliveryPool.java` and its unit test file. Other minimal changes are in `ConnectionImpl` when creating a subscription to associate to a delivery worker if needed, and when processing a message if the sub has a delivery worker, use its lock to update subscription's stats and post that message. Since we use that lock (and avoid locking sub *and* worker), the subscription's getter have been updated to lock worker if one is defined.

The library pool is set through use of library method `Nats.setMsgDeliveryThreadPoolSize()`, but I also setup the pool if a system property is set.

src/main/java/io/nats/client/MsgDeliveryPool.java
```diff
@@ -0,0 +1,174 @@
...
+            final MsgDeliveryWorker w = this.workers.get(i);
+            w.shutdown();
+        }
+        this.workers.clear();
+    }
+}
+
+class MsgDeliveryWorker extends Thread {
+    private final Lock          mu       = new ReentrantLock();
+    private final List<Message> msgs     = new LinkedList<Message>();
```
timboudreau: `java.util.LinkedList` is not thread-safe.  Unless you can guarantee that the list is both *read* and *written* from the same thread, use `CopyOnWriteArrayList` or `Collections.synchronizedList(someList)` or something [like this](https://github.com/timboudreau/util/blob/master/src/main/java/com/mastfrog/util/thread/AtomicLinkedQueue.java) that I factored out of my patch, and likely has lower write overhead than either of those.

src/main/java/io/nats/client/MsgDeliveryPool.java
```diff
@@ -0,0 +1,174 @@
...
+    private final Condition     cond     = mu.newCondition();
+    private boolean             inWait   = false;
```
timboudreau: `inWait` should be `volatile` if you are reading and writing it from different threads - two different threads can have different values in the CPU cache, so this needs to go to main memory.

src/main/java/io/nats/client/SubscriptionImpl.java
```diff
@@ -256,7 +272,9 @@ void setPendingMsgsMax(int max) {
             if (conn == null) {
                 throw new IllegalStateException(ERR_BAD_SUBSCRIPTION);
             }
+            this.dlvWorkerLock();
```
timboudreau: Is it *guaranteed* that `setDeliveryWorker()` will never be called *between* calls to `dlvWorkerLock()` and `dlvWorkerUnlock()` so it's not possible to make a call to `dlvWorker.unlock()` that never had `lock()` called?

Unlikely, but if it ever happens, debugging it will not be fun.

src/main/java/io/nats/client/SubscriptionImpl.java
```diff
@@ -51,6 +51,10 @@
     BlockingQueue<Message> mch;
     Condition pCond;
 
+    // Ideally, this should be in AsyncSubscriptionImpl, but because of
+    // locking and to reduce code duplication, leave this here.
+    private MsgDeliveryWorker dlvWorker;
+
```
timboudreau: A `SubscriptionImpl` instance will never need a `MsgDeliveryWorker` unless it is also an `AsyncSubscription` - seems like implementation code not needed by synchronous subscription instances should be there.  Really, the shared code used by both should be factored into an internal class both of them *have* rather than *are*, and `AsyncSubscriptionImpl` should not subclass `SubscriptionImpl`.  This works, but it is only going to get messier and harder to maintain as years go by.

src/main/java/io/nats/client/MsgDeliveryPool.java
kozlovic: No need since we put and remove from the list under the worker's lock.

src/main/java/io/nats/client/MsgDeliveryPool.java
kozlovic: Same as above. The inWait is used when posting the message, which is done under the worker's lock. It is just to save the need for a signal.

src/main/java/io/nats/client/SubscriptionImpl.java
kozlovic: In those cases, there is no danger since we are just capturing some state, but this would have to be changed with try/finally should it become more complex for sure.

src/main/java/io/nats/client/SubscriptionImpl.java
kozlovic: That's the one I am not happy with. I originally put the worker only in the `AsyncSubscriptionImpl` class.

But I had then to check in connection's `processMsg()` what is the instance of the subscription in order to get the delivery worker (since then the getter would only be defined in the `AsyncSubscriptionImpl` class). 

Finally, I have assumed that I should never get worker's lock -> sub's lock sequence, but only the opposite, which would then force me to duplicate all `SubscriptionImpl` getters/setter that need to grab the worker's lock (to respect the assumed locking sequence). Looking more into this, I see no reason to do the other way around, which then means that I would have to overload them all but simply get the worker's lock (if worker is assigned, which is an immutable field) and then call `super` on the parent method.

Doing the review of the above, I realized that I missed protecting subscription `closed` flag with worker lock when appropriate.

Thanks for the review, I will update the PR.

src/main/java/io/nats/client/MsgDeliveryPool.java
```diff
@@ -0,0 +1,179 @@
...
+        Message msg = null;
+        AsyncSubscriptionImpl sub = null;
+        ConnectionImpl nc = null;
+        MessageHandler mcb = null;
+        long max = 0;
+        long delivered = 0;
+
+        this.mu.lock();
+        while (true) {
+            while (this.msgs.isEmpty() && !this.shutdown) {
```
ColinSullivan1: Nice, this also accounts for a spurious wakeup in `cond.await()`.

src/main/java/io/nats/client/MsgDeliveryPool.java
```diff
@@ -0,0 +1,179 @@
...
+
+            // Capture these under lock
+            nc = (ConnectionImpl) sub.getConnection();
+            mcb = sub.getMessageHandler();
+            max = sub.max;
+
+            sub.pMsgs--;
+            sub.pBytes -= (msg.getData() == null ? 0 : msg.getData().length);
+
+            // If sub is closed, simply go back at beginning of loop.
```
ColinSullivan1: Should this happen above, after getting the subscription?

src/test/java/io/nats/client/BaseUnitTest.java
```diff
@@ -0,0 +1,41 @@
+/*
```
ColinSullivan1: This was needed.

src/main/java/io/nats/client/Nats.java
```diff
@@ -470,4 +487,63 @@ public static Options defaultOptions() {
...
+
+    /**
+     * Sets the message delivery thread pool size.
+     *
+     * By default, each asynchronous subscription has its own thread
+     * responsible to dispatch the subscription's messages.
+     * If the application needs to create a high number of subscriptions,
+     * this may not scale.
+     *
+     * Setting a message delivery thread pool size will limit the
```
ColinSullivan1: Maybe add a note the effect that size can be reset to grow, but cannot be reduced?

kozlovic:
@ColinSullivan1 I have addressed all your comments and then some. In normal situations, the pool will not shrink (because otherwise I would need to keep track of which subscription is still assigned to a pool in order to know that there are no subscription left and thread can be reclaimed). However, in tests, I do sometimes shutdown the pool and reset to different size. When size was smaller than original size, there would be an index issue. I addressed that and added unit tests to verify correct behavior if pool is shutdown from user callback.

src/test/java/io/nats/client/MsgDeliveryPoolTest.java
```diff
@@ -23,8 +25,82 @@
...
+        // Callback should be invoked twice, even if throwing exception
+        latch.await(2, TimeUnit.SECONDS);
+        worker.shutdown();
+    }
+
+    @Test
     public void testMsgDeliveryPool() throws Exception {
         MsgDeliveryPool pool = new MsgDeliveryPool();
+        // Check zero or negative size is ignored.
+        pool.setSize(-2);
```
ColinSullivan1: Should this throw an exception?

ColinSullivan1: Other than this, LGTM!

kozlovic: The user facing API is `Nats.setMsgDeliveryThreadPoolSize()` which throws an `IllegalArgumentException` if size is `<=0`. So the low level `MsgDeliveryPool.setSize()` does not need to - since it is assumed to be used correctly. The unit test calls `setSize(-2)` and verifies that size is still `0`.
I am adding in javadoc for Nats API that it can throw exception if size is <= 0.

timboudreau:
I'd say unless there is a legitimate, proven use-case for resizing the thread pool after creation - as in, somebody using the library submits a feature request for it, that you just keep it out of the API altogether (you'd probably want to rebalance the existing subscribers across the new number of threads, or it's not really going to solve much).

For that matter, in general, you guys might be better off taking more advantage of immutability to do [less locking across the board](https://timboudreau.com/blog/concurrency/read).  It seems like there is a lot of stuff in this library that's mutable after creation that probably buys very little other than maintenance headaches.

kozlovic:
@timboudreau Thing is that it is a library level pool, and so although we don't support resizing, what API would it make it obvious there is no resizing happening? I mean, a `Nats.createMsgDeliveryPool(size)` would be bad too because it has to be a singleton. I could throw an exception if it was already created though. Thoughts?

kozlovic:
The one that I would want to maybe not include is the `Nats.shutdownMsgDeliveryThreadPool()` API because there is really no reason to call this.
In C I had to have a shutdown for users that track memory leak, I need everything to be freed on exit. But here, it may not make sense.

kozlovic:
Also, should I create the pool automatically based on the system property `JNATS_MSG_DELIVERY_THREAD_POOL_SIZE`. I did that so that an application could use delivery pool without having to change a line of code, but then it may clash with the explicit use of `setMsgDeliveryThreadPoolSize()` or `createMsgDeliveryThreadPool()` (if we were to change to that). I guess if it has been created on class initialization, the explicit call could simply increase size if system prop is lower, otherwise, leave as is. It would still need to throw exception if the user call that method twice (to indicate that only one instance of this thread pool can exist).

timboudreau:
> what API would it make it obvious there is no resizing happening?
Give no way to set it except at pool construction time.  Make Nats an object you *instantiate*, not a pile of static methods, and let it have its own configuration and let the tree of objects spawned from it reference it.

> The one that I would want to maybe not include is the `Nats.shutdownMsgDeliveryThreadPool()` API because there is really no reason to call this.

I'll give you one:  Tests.  Specifically, where you want to run multiple tests in a virtual machine, and use shutdown hooks to cleanly unload absolutely everything - [here](https://github.com/timboudreau/giulius-tests) is one library that lets you do that - so you can guarantee one test does not interfere with another.

That's what was behind the idea of letting the caller pass in an `ExecutorService` in my patch - the application manages it - if they want a global pool, they can pass the same instance to everything, and if they don't, they can pass in different ones - and this library doesn't have to decide what's the one right choice for every possible caller.  Your code gets simpler, and the caller can do whatever is right for them (and, as a bonus, shut it down if that's called for).

It's true, the idea of things you can cleanly shut down is a little alien to most Java programmers - the illusion is that since you don't destruct objects, you shouldn't have to destruct anything.  But that's silly, and it's how we end up in a world of crappy software where people write scripts to bounce Apache every hour.  If you can unload it, you can test that it really doesn't leave anything behind.

Particularly in an era when dependency injection is mainstream, library-level, statically declared pools of anything is an antipattern.  It's common enough to see it still, but we shouldn't be creating more stuff like that.

src/main/java/io/nats/client/ConnectionImpl.java
```diff
@@ -1458,7 +1458,15 @@ void processMsg(byte[] data, int offset, int length) {
             // It's possible that we end up not using the message, but that's ok.
             Message msg = new Message(parser.ps.ma, sub, data, offset, length);
 
-            sub.lock();
+            MsgDeliveryWorker mdw = null;
+            if (sub instanceof AsyncSubscriptionImpl) {
+                mdw = ((AsyncSubscriptionImpl) sub).getDeliveryWorker();
+            }
+            if (mdw != null) {
```
petemiron: Is it possible to have a race on `mdw` and not be able to obtain the lock?

src/main/java/io/nats/client/ConnectionImpl.java
```diff
@@ -1811,18 +1827,27 @@ SubscriptionImpl subscribe(String subject, String queue, MessageHandler cb,
...
+                    msgDlvPool.assignDeliveryWorker((AsyncSubscriptionImpl) sub);
+                } else {
+                    // If we have an async callback, start up a sub specific Runnable to deliver the
+                    // messages
+                    subexec.submit(new Runnable() {
+                        public void run() {
+                            try {
+                                waitForMsgs((AsyncSubscriptionImpl) sub);
+                            } catch (InterruptedException e) {
+                                Thread.currentThread().interrupt();
```
petemiron: Should we do anything besides setting the interrupt flag? Is this handled elsewhere?

src/main/java/io/nats/client/MsgDeliveryPool.java
```diff
@@ -0,0 +1,182 @@
...
+    public void run() {
+        Message msg = null;
+        AsyncSubscriptionImpl sub = null;
+        ConnectionImpl nc = null;
+        MessageHandler mcb = null;
+        long max = 0;
+        long delivered = 0;
+
+        this.mu.lock();
+        while (true) {
```
petemiron: Is there a reason not to move `!this.shutdown` to the first `while(true)` loop?

src/main/java/io/nats/client/Nats.java
```diff
@@ -393,6 +397,19 @@ private Nats() {}
...
     static final String TLS_SCHEME = "tls";
 
+    protected static final String SYSTEM_ENV_MSG_DELIVERY_THREAD_POOL_SIZE = "JNATS_MSG_DELIVERY_THREAD_POOL_SIZE";
+    static private MsgDeliveryPool globalMsgDeliveryPool = null;
+
+    static {
+        final String msgDeliveryPoolSizeStr = System.getenv(SYSTEM_ENV_MSG_DELIVERY_THREAD_POOL_SIZE);
+        if (msgDeliveryPoolSizeStr != null) {
+            int poolSize = 0;
+            try { poolSize = Integer.parseInt(msgDeliveryPoolSizeStr); } catch (Exception e) {}
```
petemiron: Is this just to catch `blank`? should we do something with the exception if it's a non-blank non-int?

src/main/java/io/nats/client/ConnectionImpl.java
kozlovic: No since the worker is assigned to the subscription when the sub is being created. From that point on it is immutable.

src/main/java/io/nats/client/ConnectionImpl.java
kozlovic: I changed `waitForMsgs()` to catch any throwable that the user callback should issue. There is no need for this try/catch here, will remove.

src/main/java/io/nats/client/MsgDeliveryPool.java
kozlovic: Because I want to finish dispatching all messages before a shutdown. So shutdown flag itself is not enough condition to exit.

src/main/java/io/nats/client/Nats.java
kozlovic: I am planning to remove creation of the pool from system property and change `Nats.MsgDeliveryThreadPoolSize()` with `Nats.createMsgDeliveryThreadPool()` to remove any idea that pool could be resized. Will update the PR in a bit.

PR #136, opened 2017-10-16
wake-up any message threads (so they can terminate) when closing subs…

edam:
Currently, if you manually close a subscription with:

    sub.close()

the subscription is internally unregistered and the server is informed.  But the thread that waits on a message to come in is left blocking.  This quickly causes leaked threads (and memory) where lots of subs are manually closed.

E.g., try this:
```java
    public void main()
    {
        for( int round = 0; round < 3; round++ )
        {
            subs = new ArrayList<>();
            for( int a = 0; a < NUM; a++ ) {
                subs.add( con.subscribe( "test", m -> {} ) );
            }
            Thread.sleep( 1000 );
            report( "subscribed" );

            for( Subscription sub : subs ) {
                sub.close();
            }
            Thread.sleep( 1000 );
            report( "closed" );
        }
    }

    static void report( String at )
    {
        int threads = Thread.getAllStackTraces().size();
        System.out.println( at + " (" + threads + " threads)" );
    }
```

Note that this is not the same as issue #114, which deals with a different approach to spawning message threads.  This patch, I believe, fixes the current approach.

### DIFF
--- a/src/main/java/io/nats/client/SubscriptionImpl.java
+++ b/src/main/java/io/nats/client/SubscriptionImpl.java
@@ -128,12 +128,20 @@ abstract class SubscriptionImpl implements Subscription {
         mu.lock();
         conn = this.conn;
         mu.unlock();
-        if (conn == null) {
-            if (!ignoreInvalid) {
-                throw new IllegalStateException(ERR_BAD_SUBSCRIPTION);
+        try {
+            if (conn == null) {
+                if (!ignoreInvalid) {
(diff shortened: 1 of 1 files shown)


kozlovic:
@edam Thanks. I have already fixed this issue in PR #135 (https://github.com/nats-io/java-nats/pull/135/files#diff-6f327304322aab567c0a9d69e16a0cc9R105). Since it is probable that we merge that other PR, I will close this one now. Thanks!

Issue #146, opened 2018-01-26
NullPointerException in io.nats.client.Message#toString


skoskav:
NPE occurs in both 1.0 and master:
```
public static void main(String[] args) {
    Message message = new Message();
    System.out.println(message);
}
```

```
Exception in thread "main" java.lang.NullPointerException
	at io.nats.client.Message.getSubject(Message.java:106)
	at io.nats.client.Message.toString(Message.java:220)
	at java.lang.String.valueOf(String.java:2994)
	at java.io.PrintStream.println(PrintStream.java:821)
	at org.skoskav.jnats.Demo.main(Demo.java:30)
```

I came across it when writing JUnit tests for an `org.apache.camel.Processor` implementation. Apache Camel's mock library is unable to assert on the `Message` being tested, which just has its `data` field set:
```
    @Override
    public void process(Exchange exchange) {
        String payload = exchange.getIn().getBody(String.class);
        Message message = new Message();
        message.setData(payload.getBytes(StandardCharsets.UTF_8));
        exchange.getIn().setBody(message);
    }
```

kozlovic:
Fixed in #147

PR #147, opened 2018-01-26
[FIXED] NPE in io.nats.client.Message#getSubject

skoskav:
Adding null check to `io.nats.client.Message#getSubject` to match behavior with `getReplyTo` and make `toString` work, fixing #146


kozlovic:
LGTM. Thanks for the contribution!

PR #148, opened 2018-01-31
Removed section about logging in README.md

suhlig:


kozlovic:
Thanks, but I believe we have removed the dependency on sl4j, @ColinSullivan1 correct me if I am wrong. If that the case, then we should just remove that section from the README.md.

ColinSullivan1:
@suhlig thank you for the PR, @kozlovic is correct though - we no longer have a dependency on sl4j - if you'd like to update the PR to remove that section, we'd appreciate it!

kozlovic:
LGTM. Thanks for the contribution!

PR #149, opened 2018-02-15
handle IOExceptions while writing resubscriptions (and pending data) …

edam:
IOExceptions thrown as a result of writing to the OutputStream during a reconnect, whilst trying to write subscription (and pending) commands should not be silently ignored.  This could lead to the scenario where the library believes it has successfully recovered a connection, but where some subscriptions have not been re-established.

This scenario may also be especially likely, given that the reconnection code may run where there are networking issues/partitions .  At this time, conceivably, a backlog of pending RPC-like requests (i.e., publishes that require a response via a one-shot inbox) may have accumulated, leading to an excess of subscriptions to re-establish.


sasbury:
Too many library changes to bring this PR in, if you see something similar in 2.x please resubmit.

PR #154, opened 2018-03-20
CNCF Updates

ColinSullivan1:
* Apache 2.0 License
* Add code of conduct
* Add owners file
* Update copyright

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2015-2018 The NATS Authors
+   Copyright [yyyy] [name of copyright owner]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/LICENSE
+++ b/LICENSE
@@ -1,22 +1,201 @@
-The MIT License (MIT)
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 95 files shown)


LICENSE
```diff
@@ -1,22 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
aricart: Need to update this with "$BIRTH-2018 The NATS Authors"

OWNERS
```diff
@@ -0,0 +1,10 @@
+reviewers:
+- aricart
```
aricart: ...

LICENSE
ColinSullivan1: Not sure we do that in the license here.  These are boilerplate instructions...   An example elsewhere:  https://github.com/kubernetes/kubernetes/blob/master/LICENSE

ColinSullivan1: Let's keep as is for now, and key off of what we decide to do in gnatsd.

PR #174, opened 2018-09-06
V2.1.0

sasbury:
Support for Drain on the connection and Subscription.
Fixed a null pointer in certain situations with ping/reconnect
Added more JDKs to travis build

Bumping version number for added drain API.


src/main/java/io/nats/client/Connection.java
```diff
@@ -234,10 +234,45 @@
...
 
+    /**
+     * Drain tells the connection to process in flight messages before closing.
+     * 
+     * Drain initially drains all of the consumers, stopping incoming messages.
+     * Next, publishing is halted and a flush call is used to insure all published
+     * messages have reached the server.
+     * Finally the connection is closed.
+     * 
+     * In order to drain subscribers, an unsub is sent to the server followed by a flush.
```
ColinSullivan1: Nitpick.  Maybe say UNSUB protocol message?

src/main/java/io/nats/client/Connection.java
```diff
@@ -234,10 +234,45 @@
...
+     * 
+     * The connection will try to let all messages be drained, but when the timeout is reached
+     * the connection is closed and any outstanding dispatcher threads are interrupted.
+     * 
+     * A future is used to allow this call to be treated as synchronous or asynchronous as
+     * needed by the application. The value of the future will be true if all of the subscriptions
+     * were drained in the timeout, and false otherwise. The future is completed after the connection
+     * is closed, so any connection handler notifications will happen before the future completes.
+     * 
+     * @param timeout The time to wait for the drain to succeed, pass 0 to wait
```
ColinSullivan1: In other languages we use -1 for "forever".

src/main/java/io/nats/client/Connection.java
sasbury: fixed

src/main/java/io/nats/client/Connection.java
sasbury: since we don't support 0 time, i think this is ok. Also, changing would require a major revision bump.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -748,7 +793,7 @@ public Message request(String subject, byte[] body, Duration timeout) throws Int
         Future<Message> incoming = this.request(subject, body);
         try {
             reply = incoming.get(timeout.toNanos(), TimeUnit.NANOSECONDS);
-        } catch (ExecutionException|TimeoutException e) {
+        } catch (ExecutionException | TimeoutException e) {
```
ColinSullivan1: Did you intend to leave interrupted and cancellation exceptions to be uncaught?

sasbury: yes, interrupted exceptions should rarely be eaten as they tell the caller the thread was interrupted. Cancellation has the same meaning here for the future, so we don't want to eat it.

PR #185, opened 2018-10-19
V2.1.2

sasbury:
Added UTF8 subject option
Updated autobenchmark test a bunch to work with UTF8 and to have modes for smaller tests (ran into that on my chromebook)
Minor update for default pending limits and jar version strings.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsMessage.java
+++ b/src/main/java/io/nats/client/impl/NatsMessage.java
@@ -40,74 +40,73 @@ class NatsMessage implements Message {
         return pos;
     }
 
+    private static String PUB_SPACE = NatsConnection.OP_PUB + " ";
+    private static String SPACE = " ";
+
     // Create a message to publish
-    NatsMessage(String subject, String replyTo, byte[] data) {
+    NatsMessage(String subject, String replyTo, byte[] data, boolean utf8mode) {
         this.subject = subject;
         this.replyTo = replyTo;
         this.data = data;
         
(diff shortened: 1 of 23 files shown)


src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -40,74 +40,73 @@ static int copy(byte[] dest, int pos, String toCopy) {
...
-        StringBuilder protocolStringBuilder = new StringBuilder((2 * subject.length()) + 20); // guess a size based on replyTO = subject
-        protocolStringBuilder.append(PUB_SPACE);
-
-        protocolStringBuilder.append(subject);
-        protocolStringBuilder.append(SPACE);
-
-        if (replyTo != null) {
-            protocolStringBuilder.append(replyTo);
+        if (utf8mode) {
+            StringBuilder protocolStringBuilder = new StringBuilder((2 * subject.length()) + 20); // guess a size based on replyTO = subject
```
kozlovic: I guess it is just an estimation and size will grow as needed, but strange that you use 2*subject. You could have something like:
```
int subjectSize = subject.length() * 2;
int replySize = (replyTo != null) ? replyTo.length() * 2 : 0;
StringBuilder protocolStringBuilder = new StringBuilder(4 + subjectSize + 1 + replySize + 1);
```

sasbury: made the change.

PR #208, opened 2019-01-28
Backport to Java 7 and Android API less 26 

AcoustickSan:
On Android use [ThreeTenABP](https://github.com/JakeWharton/ThreeTenABP), on Java [ThreeTen](https://www.threeten.org/threetenbp/)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -100,18 +100,17 @@ test {
 
 javadoc {
     options.overview = 'src/main/javadoc/overview.html' // relative to source root
-    options.encoding = 'UTF-8'
     source = sourceSets.main.allJava
     title = "NATS.IO Java API"
     excludes = ['io/nats/client/impl', 'io/nats/examples']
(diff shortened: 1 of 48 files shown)


sasbury:
I think we should do this as a fork. I am worried that backporting to Java 7, when Java 8 is already old is the wrong direction for the library, but I also understand the need to deal with android fragmentation.

AcoustickSan:
Ok. Can I create new repo with android example and publish library with link for you as original?

sasbury:
If you fork the repo to create a Java 7/Android version, GitHub will link back to the original (and point the other way), but I can also add a link to the readme here so that others can find your work. I keep hoping Google will push Java 8 to every Android phone but I know that just isn't in the cards yet.

sasbury:
Closing in favor of https://github.com/SpotOnInc/nats-android. Sorry for the delays in resolving.

PR #210, opened 2019-02-12
Update NatsMessage

Arunold:
Removed unwanted code


sasbury:
Somethings up with travis, looks like an obvious clean up so I don't think it is the code. Now the question is can I remember to bring it in or move it to a branch, git will not defeat me.

sasbury:
Ok, looks like the travis issue is a bug on their side.

I created a branch for 2.4.2. Can you move this PR to that branch and I will merge it.

Issue #184, opened 2018-10-16
Compatibility with lower versions of Android


asbapat:
Hi,

I tried running the nats client on lower version of Android (API 21) and it threw a "java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/Duration;" for the class Options.Builder implemented in the code.

How can I implement this library for running nats client subscription of API Level < 26?

Crash Report:
`2018-10-16 11:21:48.656 14742-14759/com.example.aniketbapat.demoapp E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #1
    Process: com.example.aniketbapat.demoapp, PID: 14742
    java.lang.RuntimeException: An error occurred while executing doInBackground()
        at android.os.AsyncTask$3.done(AsyncTask.java:318)
        at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:354)
        at java.util.concurrent.FutureTask.setException(FutureTask.java:223)
        at java.util.concurrent.FutureTask.run(FutureTask.java:242)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
        at java.lang.Thread.run(Thread.java:762)

     Caused by: java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/Duration;
        at io.nats.client.Options.<clinit>(Options.java:80)
        at io.nats.client.Options$Builder.<init>(Options.java:390)
        at com.example.aniketbapat.demoapp.MainActivity$ConnectServer.doInBackground(MainActivity.java:70)
        at com.example.aniketbapat.demoapp.MainActivity$ConnectServer.doInBackground(MainActivity.java:60)
        at android.os.AsyncTask$2.call(AsyncTask.java:304)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243) 
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133) 
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607) 
        at java.lang.Thread.run(Thread.java:762)`

kkocel:
[...]

PR #211, opened 2019-02-12
#184 Android compatibility pre API 24

kedzie:


kedzie:
This is for issue #184

kedzie:
I replaced `CompletableFuture` everywhere with an implementation based on `CountDownLatch`.  I even replaced it in the tests/examples, which we could undo to make the PR smaller.

sasbury:
This looks like a really clean way to port to Android. There is another PR that uses a different dependency, but I like how you just implemented what you need with available JDK classes.

I am still thinking we need a new repo because this is a Major API change, it will break existing users by changing imported classes. The other reason I have been thinking of Android as a separate repo is that there are likely power/network optimizations that apply to mobile and not desktop/server.

kedzie:
Do you want to maintain it?  Start a new repo in the nats-io org will give the android project exposure, vs leaving it in my fork

sasbury:
I have a biz trip next week, but will think on how we want to handle this, we may bring it in but try to get some outside maintainer support (would you be up for something like that?) I haven't talked to the team but that would probably work better than a non-android developer maintaining.

kedzie:
I'd be happy to contribute my efforts.

sasbury:
I talked to the team, the preferred path is to have you make a fork, and then we will add it to the community library list on the site, that way people can find it. Moving forward we can decide if it should get adopted in. Does that seem reasonable? Would you be willing to have us point to your version?

kedzie:
Yes I'd be glad to have you guys point to it.

kedzie:
I will also merge my changes to master on my fork and rebrand the resume to nats android.

sasbury:
Sounds good, ping the final repo URL and I will get it on the site, and maybe add something to the readme here.

kedzie:
Hey I have a unrelated question.   on `NatsDispatcher` in the `subscribeImpl` method line 186 does ```sub = connection.createSubscription(subject, queueName, this);
            NatsSubscription actual = subscriptions.putIfAbsent(subject, sub);
            if (actual != null) {
                this.connection.unsubscribe(sub, -1); // Could happen on very bad timing
            }``` . This comment confused me because `putIfAbsent` always returns a value.  either the existing value or the newly inserted value.  So `actual != null` is **always true**.  So it will unsubscribe... and this

sasbury:
For concurrent hash map the doc says "Returns: the previous value associated with the specified key, or null if there was no mapping for the key" so it should be null unless the subscription was already there. I am looking at jdk 8, maybe it changed somewhere?

kedzie:
I updated the readme and included an Android example app.  Will have to figure out travis builds and publish the artifact somewhere.   [nats-android](https://github.com/SpotOnInc/nats-android)

kedzie:
In the Android JDK 8 it is implemented as
```
default V putIfAbsent(K key, V value) {
        V v = get(key);
        if (v == null) {
            v = put(key, value);
        }

        return v;
    }
```

kedzie:
Yes you were correct about the `putIfAbsent` only returns when as value was already there.  My mistake I will implement the same behavior.

sasbury:
Closing the PR, now that we have https://github.com/SpotOnInc/nats-android on nats.io :-)

PR #215, opened 2019-03-04
216 - NatsConnection thread pool

kedzie:


sasbury:
Thinking on this. I actually had it on my todo for a while, but realized that threads aren't recycled so the pool is just providing a place to make new threads, but we aren't using anything "pool like". What was your motivation? Maybe i just gave up too soon.

kedzie:
The idea would be to have it configurable.  So one could pass in an Executor implementation in the Options.   By default a CachedThreadPool would only offer benefit if threads were reused.  There are threads made for Draining i believe.  Or if dispatchers are created and removed.

sasbury:
I will take a look this week, trying to wrap something up.

sasbury:
Planning to pull this in today, i am going to do a few changes to support naming (which helps in debugging) and also to put the default into the options object.

PR #218, opened 2019-03-18
Do not call exception handler on IOException during connection closing/draining

yuanmwang-wf:
Fixes https://github.com/nats-io/java-nats/issues/217.

Confirmed exceptionOccurred is not triggered with the same example code listed in 217.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnectionReader.java
+++ b/src/main/java/io/nats/client/impl/NatsConnectionReader.java
@@ -143,7 +143,10 @@ class NatsConnectionReader implements Runnable {
                 }
             }
         } catch (IOException io) {
-            this.connection.handleCommunicationIssue(io);
+            // SocketException is thrown when connection is closing or draining, no need to handle
+            if (!this.connection.isClosingOrDraining()) {
+                this.connection.handleCommunicationIssue(io);
+            }
         } catch (CancellationException | ExecutionException | InterruptedException ex) {
             // Exit
         } finally {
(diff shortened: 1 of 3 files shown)


src/main/java/io/nats/client/impl/NatsConnectionReader.java
```diff
@@ -352,7 +355,7 @@ public String opFor(char[] chars, int length) {
             } else if (chars[0] == '-' && 
                         (chars[1] == 'E' || chars[1] == 'e') &&
                         (chars[2] == 'R' || chars[2] == 'r') && 
-                        (chars[3] == 'R' || chars[3] == 'R')) {
+                        (chars[3] == 'R' || chars[3] == 'r')) {
```
yuanmwang-wf: also noticed this probably needs to be lowercase :)

sasbury:
I will take a look this week, trying to wrap something up.

src/main/java/io/nats/client/impl/NatsConnectionReader.java
sasbury: very nice catch on the R->r, my brain never even thought to see that.

I started looking at the PR and realized i am handling a bunch of these states in handleCommunicationIssue already, so will fix it there:
```
            if (this.connecting || this.disconnecting || this.status == Status.CLOSED || this.isDraining()) {
```

As a result I will likely close this PR without a merge, but the issue will be fixed in 2.4.2, which I am working on as I type. When I get the change in to connection.handleCommunicationIssue, i will ping to have you try it out before I release 2.4.2

yuanmwang-wf: ok, just let me know when 2.4.2 is ready, thanks!

sasbury:
Added fix to connection code, and updated R->r :-)

fix is in the 2.4.2 branch.

PR #222, opened 2019-03-19
Fixed an icky bug in message queue caused by waiter list timing and s…

### LINKED ISSUE #220: Race condition can prevent message from being read until another message is received
I'm not sure if anyone here is very familiar with the java client, but I was doing some debugging of an intermittent issue and I think tracked it down, but I'm not sure the best way to fix it. What I think is happening is
1. thread1 - https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/impl/MessageQueue.java#L136 is called, and no messages are available in the queue.
2. thread2 - https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/impl/MessageQueue.java#L101 is called. A message is received, but no waiters are currently in the queue, so nothing is signaled.
3. thread1 - Our waiter is eventually added, and https://github.com/nats-io/java-nats/blob/master/src/main/java/io/nats/client/impl/MessageQueue.java#L154 is called as our waiter wasn't in the list when the message was received, so we weren't able to be notified and our thread is parked, and our subscriber never receives a message. The message is pulled off the queue when another message arrives to notify the waiter, but sometimes that doesn't occur for several seconds, leaving our client receiving a timeout for the first message

I can reproduce the issue with this patch and the `testPopBeforeTimeout` test. The best case scenario would be the message is received as soon as the sleep in `waitForTimeout` is over, instead the test spins for a long, long time. (Sorry for the line endings)
```
diff --git a/src/main/java/io/nats/client/impl/MessageQueue.java b/src/main/java/io/nats/client/impl/MessageQueue.java
index 95e8fce..100bb6a 100644
--- a/src/main/java/io/nats/client/impl/MessageQueue.java
+++ b/src/main/java/io/nats/client/impl/MessageQueue.java
@@ -149,6 +149,8 @@ class MessageQueue {
                     }
                 }

+                Thread.sleep(5000);^M
+^M
                 waiters.add(t);
                 if (timeoutNanos == 0) {
                     LockSupport.park();
@@ -294,4 +296,4 @@ class MessageQueue {

         this.queue.addAll(newQueue);
     }
-}
\ No newline at end of file
+}^M
diff --git a/src/test/java/io/nats/client/impl/MessageQueueTests.java b/src/test/java/io/nats/client/impl/MessageQueueTests.java
index eadd84a..07aeeb4 100644
--- a/src/test/java/io/nats/client/impl/MessageQueueTests.java
+++ b/src/test/java/io/nats/client/impl/MessageQueueTests.java
@@ -109,7 +109,7 @@ public class MessageQueueTests {
         t.start();

         // Thread timing, so could be flaky
-        NatsMessage msg = q.pop(Duration.ofMillis(200));
+        NatsMessage msg = q.pop(Duration.ofMillis(2000000));^M
         assertNotNull(msg);
     }

@@ -519,4 +519,4 @@ public class MessageQueueTests {
         q.filter((msg) -> {return true;});
         assertFalse(true);
     }
-}
\ No newline at end of file
+}^M
```

sasbury:
Fixed issue in message queue due to timing between spin wait and going into the waiters list

Fixes #220

### DIFF
--- a/src/main/java/io/nats/client/impl/MessageQueue.java
+++ b/src/main/java/io/nats/client/impl/MessageQueue.java
@@ -149,8 +149,19 @@ class MessageQueue {
                     }
                 }
 
+                // Thread.sleep(1000) <- use this to test the "isEmpty" fix below
+                // see https://github.com/nats-io/java-nats/issues/220 for discussion
+
+                // once we are in the waiters, we can be signaled, but
+                // until then we can't, so there is a possible timing bug
+                // where we aren't in waiters, aren't checking queue and
+                // miss the signalone when we are the only reader. By double
+                // checking isEmpty we insure that we either get the signal in park
+                // or we have a very short park here
                 waiters.add(t);
-                if (timeoutNanos == 0) {
+                if (!this.queue.isEmpty()) {
+                    LockSupport.parkNanos(SPIN_WAIT);
+                } else if (timeoutNanos == 0) {
                     LockSupport.park();
                 } else {
                     LockSupport.parkNanos(timeoutNanos);
--- a/src/test/java/io/nats/client/impl/MessageQueueTests.java
+++ b/src/test/java/io/nats/client/impl/MessageQueueTests.java
@@ -100,7 +100,7 @@ public class MessageQueueTests {
 
         Thread t = new Thread(() -> {
             try {
-                Thread.sleep(100);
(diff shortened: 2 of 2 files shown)


brianshannan-wf:
+1 from me, this fixes the issue I was seeing

PR #229, opened 2019-04-13
Allow supplying custom threadExecutor for Options

altmind:
There is currently no way to change NATS thread names or join NATS threads in single thread group.

This is useful because currently NATS threads dont have "nats" in thread name and `connectionName-003:20` as hardly a meaningful name.

This PR adds `Options.threadFactory()` that can override the use of DefaultThreadFactory.
I removed some stray whitespace to be closer to codeStyle and questionable comparison(not .equals?)
`(this.connectionName != null && this.connectionName != "") `

A test is included.

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -100,7 +100,7 @@ public class Options {
      * Default server ping interval. The client will send a ping to the server on this interval to insure liveness.
      * The server may send pings to the client as well, these are handled automatically by the library
      * , see {@link #getPingInterval() getPingInterval()}.
-     * 
+     *
      * <p>A value of {@code <=0} means disabled.
      *
      * <p>This property is defined as 2 minutes.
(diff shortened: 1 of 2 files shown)


altmind:
The checks seems to fail for me missing some keys. I dont know what are these and there is no CONTRIBUTING.md to inquire:
```
The command "openssl aes-256-cbc -K $encrypted_f07928735f08_key -iv $encrypted_f07928735f08_iv -in .travis/nats.travis.gpg.enc -out .travis/nats.travis.gpg -d" failed and exited with 1 during .
```

sasbury:
Please ignore the missing checks for now.

I have a method to set the executor `executor()` in the options, you can use this to set up a thread factory. The current default is:

```java
this.executor = new ThreadPoolExecutor(0, Integer.MAX_VALUE,
                                                        500L, TimeUnit.MILLISECONDS,
                                                        new SynchronousQueue<Runnable>(),
                                                        new DefaultThreadFactory(threadPrefix));
```

but you could do whatever. I do use a short timeout to prevent long delays on shutdown.

PR #236, opened 2019-05-17
Added optional back pressure in the message queue

sasbury:
This is not the full 2.5, i will not merge this as is, but wanted to get a check on the backpressure implementation. I meant to do the commit differently but my git is no good. Ignore the NatsBench.java change. It is just to have an error listener. Main change is in MessageQueue.java


sasbury:
closing for now, ivan suggested something and it it working out better

PR #237, opened 2019-05-17
Refactored MessagesQueue to use LinkedBlockingQueue

sasbury:
message queue was using a lot of custom code to handle all the performance and locking requirements. Now it uses a LinkedBlockingQueue which is much simpler. The new implementation uses a poison pill to shutdown.

By switching to a blocking queue we get back pressure on publishers. This changes the performance characteristics from Autobench but mostly to smooth them out over all message sizes. Latency and request/reply performance are nominal.

Added error handler to NatsBench.
Fixed a minor warning in raw tcp test

Prepare for 2.5.0, updated all versions

### DIFF
--- a/src/examples/java/io/nats/examples/autobench/ThrottledBenchmark.java
+++ b/src/examples/java/io/nats/examples/autobench/ThrottledBenchmark.java
@@ -70,7 +70,7 @@ public abstract class ThrottledBenchmark extends AutoBenchmark {
 
     /**
      * Throttled benchmarks use an optional target send rate. This is to make sure the subscribers are not overrun,
-     * become slow consumers. The addition of this code was motivated by the assymetry in the java library between
+     * become slow consumers. The addition of this code was motivated by the asymetry in the java library between
      * publishers and subscribers, especially if UTF-8 subjects are enabled.
      * @param nc the connection
      * @throws InterruptedException
--- a/src/examples/java/io/nats/examples/autobench/ThrottledBenchmark.java
+++ b/src/examples/java/io/nats/examples/autobench/ThrottledBenchmark.java
@@ -68,6 +68,13 @@ public abstract class ThrottledBenchmark extends AutoBenchmark {
 
     }
 
+    /**
+     * Throttled benchmarks use an optional target send rate. This is to make sure the subscribers are not overrun,
+     * become slow consumers. The addition of this code was motivated by the assymetry in the java library between
+     * publishers and subscribers, especially if UTF-8 subjects are enabled.
+     * @param nc the connection
+     * @throws InterruptedException
+     */
 	void adjustAndSleep(Connection nc) throws InterruptedException {
 
         long count = sent.incrementAndGet();
--- a/src/examples/java/io/nats/examples/benchmark/NatsBench.java
+++ b/src/examples/java/io/nats/examples/benchmark/NatsBench.java
@@ -211,7 +211,7 @@ public class NatsBench {
 
         private AtomicLong start;
 
-        PubWorker(Future<Boolean> starter, Phaser finisher, int numMsgs, int size, boolean secure, long targetPubRate) {
+        PubWorker(Future<Boolean> starter, Phaser finisher, int numMsgs, int size, boolean secure) {
             super(starter, finisher, numMsgs, size, secure);
             this.start = new AtomicLong();
         }
@@ -323,9 +323,9 @@ public class NatsBench {
                 }
                 
                 if (subCount == 0) {
-                    new Thread(new PubWorker(starter, finisher, perPubMsgs, this.size, secure, 0), "Pub-"+i).start();
+                    new Thread(new PubWorker(starter, finisher, perPubMsgs, this.size, secure), "Pub-"+i).start();
                 } else {
-                    new Thread(new PubWorker(starter, finisher, perPubMsgs, this.size, secure, 2_000_000), "Pub-"+i).start();
+                    new Thread(new PubWorker(starter, finisher, perPubMsgs, this.size, secure), "Pub-"+i).start();
                 }
                 
                 remaining -= perPubMsgs;
--- a/src/examples/java/io/nats/examples/benchmark/NatsBench.java
+++ b/src/examples/java/io/nats/examples/benchmark/NatsBench.java
@@ -105,6 +105,23 @@ public class NatsBench {
         builder.noReconnect();
         builder.connectionName("NatsBench");
         builder.servers(servers);
+        builder.errorListener(new ErrorListener(){
+            @Override
+            public void errorOccurred(Connection conn, String error) {
+                System.out.printf("An error occurred %s\n", error);
+            }
+
+            @Override
+            public void exceptionOccurred(Connection conn, Exception exp) {
+                System.out.println("An exception occurred...");
+                exp.printStackTrace();
+            }
+
+            @Override
+            public void slowConsumerDetected(Connection conn, Consumer consumer) {
+                System.out.println("Slow consumer detected");
+            }
+        });
         //builder.turnOnAdvancedStats();
 
         if (secure) {
@@ -213,10 +230,12 @@ public class NatsBench {
                     nc.publish(subject, payload);
                     sent.incrementAndGet();
                 }
-                nc.flush(Duration.ofSeconds(5));
+                nc.flush(Duration.ofSeconds(15));
                 long end = System.nanoTime();
 
                 bench.addPubSample(new Sample(numMsgs, size, start.get(), end, nc.getStatistics()));
+
+                //System.out.println(nc.getStatistics());
                 nc.close();
             } catch (Exception e) {
                 errorQueue.add(e);
@@ -265,7 +284,7 @@ public class NatsBench {
             
             // Flush small messages regularly
             if (this.size < 64 && count != 0 && count % 100_000 == 0) {
-                try {nc.flush(Duration.ofSeconds(5));}catch(Exception e){}
+                try {nc.flush(Duration.ofSeconds(15));}catch(Exception e){}
             }
         }
     }
--- a/src/examples/java/io/nats/examples/benchmark/NatsBench.java
+++ b/src/examples/java/io/nats/examples/benchmark/NatsBench.java
@@ -35,7 +35,6 @@ import java.util.concurrent.Phaser;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.locks.LockSupport;
 
 /**
  * A utility class for measuring NATS performance, similar to the version in go and node.
@@ -57,6 +56,7 @@ public class NatsBench {
     private final AtomicLong sent = new AtomicLong();
     private final AtomicLong received = new AtomicLong();
     private boolean csv = false;
+    private boolean stats = false;
 
     private Thread shutdownHook;
     private final AtomicBoolean shutdown = new AtomicBoolean(false);
@@ -72,7 +72,8 @@ public class NatsBench {
                     + "    -ns                             Number of concurrent subscribers (0)\n"
                     + "    -n                              Number of messages to publish (100,000)\n"
                     + "    -ms                             Size of the message (128)\n"
-                    + "    -csv                            Print results to stdout as csv (false)\n";
+                    + "    -csv                            Print results to stdout as csv (false)\n"
+                    + "    -stats                          Track and print out internal statistics (false)\n";
 
     public NatsBench(String[] args) throws Exception {
         if (args == null || args.length < 1) {
@@ -122,7 +123,10 @@ public class NatsBench {
                 System.out.println("Slow consumer detected");
             }
         });
-        //builder.turnOnAdvancedStats();
+
+        if (stats) {
+            builder.turnOnAdvancedStats();
+        }
 
         if (secure) {
             builder.secure();
@@ -187,7 +191,9 @@ public class NatsBench {
 
                 bench.addSubSample(new Sample(numMsgs, size, start, end, nc.getStatistics()));
 
-                //System.out.println(nc.getStatistics());
+                if (stats) {
+                    System.out.println(nc.getStatistics());
+                }
 
                 // Clean up
                 sub.unsubscribe();
@@ -204,12 +210,10 @@ public class NatsBench {
     class PubWorker extends Worker {
 
         private AtomicLong start;
-        private long targetPubRate;
 
         PubWorker(Future<Boolean> starter, Phaser finisher, int numMsgs, int size, boolean secure, long targetPubRate) {
             super(starter, finisher, numMsgs, size, secure);
             this.start = new AtomicLong();
-            this.targetPubRate = targetPubRate;
         }
 
         @Override
@@ -235,56 +239,14 @@ public class NatsBench {
 
                 bench.addPubSample(new Sample(numMsgs, size, start.get(), end, nc.getStatistics()));
 
-                //System.out.println(nc.getStatistics());
+                if (stats) {
+                    System.out.println(nc.getStatistics());
+                }
                 nc.close();
             } catch (Exception e) {
                 errorQueue.add(e);
             } finally {
                 finisher.arrive();
-            }
-        }
-
-
-        void adjustAndSleep(Connection nc) throws InterruptedException {
-
-            if (this.targetPubRate <= 0) {
-                return;
-            }
-
-            long count = sent.incrementAndGet();
-
-            if (count % 1000 != 0) { // Only sleep every 1000 message
-                return;
-            }
-
-            long now = System.nanoTime();
-            long start = this.start.get();
-            double rate = (1e9 * (double) count)/((double)(now - start));
-            double delay = (1.0/((double) this.targetPubRate));
-            double adjust = delay / 20.0; // 5%
-            
-            if (adjust == 0) {
-                adjust = 1e-9; // 1ns min
-            }
-            
-            if (rate < this.targetPubRate) {
-                delay -= adjust;
-            } else if (rate > this.targetPubRate) {
-                delay += adjust;
-            }
-            
-            if (delay < 0) {
-                delay = 0;
-            }
-
-            delay = delay * 1000; // we are doing this every 1000 messages
-            
-            long nanos = (long)(delay * 1e9);
-            LockSupport.parkNanos(nanos);
-            
-            // Flush small messages regularly
-            if (this.size < 64 && count != 0 && count % 100_000 == 0) {
-                try {nc.flush(Duration.ofSeconds(15));}catch(Exception e){}
             }
         }
     }
@@ -481,11 +443,18 @@ public class NatsBench {
                     it.remove();
                     continue;
                 case "-csv":
-                    if (!it.hasNext()) {
+                    if (it.hasNext()) {
                         usage();
                     }
                     it.remove();
                     csv = true;
+                    continue;
+                case "-stats":
+                    if (it.hasNext()) {
+                        usage();
+                    }
+                    it.remove();
+                    stats = true;
                     continue;
                 default:
                     System.err.printf("Unexpected token: '%s'\n", arg);
--- a/src/main/java/io/nats/client/impl/MessageQueue.java
+++ b/src/main/java/io/nats/client/impl/MessageQueue.java
@@ -35,14 +35,16 @@ class MessageQueue {
     private final LinkedBlockingQueue<NatsMessage> queue;
     private final Lock filterLock;
 
-    // Poison pill is a graphic, but common term for an item that breaks loops or forces some other action
+    // Poison pill is a graphic, but common term for an item that breaks loops or stop something.
     // In this class the poisonPill is used to break out of timed waits on the blocking queue.
+    // A simple == is used to check if any message in the queue is this message.
     private final NatsMessage poisonPill;
 
     /**
-     * If publishHighwaterMark is set to 0 the underlying queue can grow forever. This is used by readers
-     * to prevent the read thread from blocking. If set to a number, the publish command will block, which provides
-     * backpressure on a publisher if the writer is slow to push things onto the network.
+     * If publishHighwaterMark is set to 0 the underlying queue can grow forever (or until the max size of a linked blocking queue that is).
+     * A value of 0 is used by readers to prevent the read thread from blocking.
+     * If set to a number of messages, the publish command will block, which provides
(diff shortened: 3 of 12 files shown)


sasbury:
Ok, cleaned out the old PR, made this new one with a blocking queue. Easier to read. performance is about the same, a bit worse for pub-only on small messages, but more consistent so "better" in my opinion.

sasbury:
Benchmark run hitting demo servers:

```
 % java -cp ./build/libs/jnats-2.5.0-SNAPSHOT-examples.jar:./build/libs/jnats-2.5.0-SNAPSHOT-fat.jar io.nats.examples.benchmark.NatsBench -s nats://demo.nats.io:4222 -np 2 -ns 5 -n 10000 -ms 1024 test

Starting benchmark(s) [msgs=10000, msgsize=1024, pubs=2, subs=5]
Current memory usage is 971.26 mb / 981.50 mb / 14.22 gb free/total/max
Use ctrl-C to cancel.

Pub Only stats: 2,619 msgs/sec ~ 2.56 mb/sec
 [ 1] 2,141 msgs/sec ~ 2.09 mb/sec (5000 msgs)
 [ 2] 1,309 msgs/sec ~ 1.28 mb/sec (5000 msgs)
  min 1,309 | avg 1,725 | max 2,141 | stddev 416.00 msgs

Pub/Sub stats: 7,513 msgs/sec ~ 7.34 mb/sec
 Pub stats: 1,261 msgs/sec ~ 1.23 mb/sec
  [ 1] 2,425 msgs/sec ~ 2.37 mb/sec (5000 msgs)
  [ 2] 631 msgs/sec ~ 631.09 kb/sec (5000 msgs)
   min 631 | avg 1,528 | max 2,425 | stddev 897.00 msgs
 Sub stats: 6,257 msgs/sec ~ 6.11 mb/sec
  [ 1] 1,252 msgs/sec ~ 1.22 mb/sec (10000 msgs)
  [ 2] 1,251 msgs/sec ~ 1.22 mb/sec (10000 msgs)
  [ 3] 1,251 msgs/sec ~ 1.22 mb/sec (10000 msgs)
  [ 4] 1,251 msgs/sec ~ 1.22 mb/sec (10000 msgs)
  [ 5] 1,251 msgs/sec ~ 1.22 mb/sec (10000 msgs)
   min 1,251 | avg 1,251 | max 1,252 | stddev .45 msgs


Final memory usage is 760.77 mb / 981.50 mb / 14.22 gb free/total/max
```

sasbury:
autobenchmark run for broad perf numbers

```
 % java -cp ./build/libs/jnats-2.5.0-SNAPSHOT-examples.jar:./build/libs/jnats-2.5.0-SNAPSHOT-fat.jar io.nats.examples.autobench.NatsAutoBench nats://localhost:4222 
Connecting to NATS server at nats://localhost:4222
Running warmup
Current memory usage is 894.33 mb / 981.50 mb / 14.22 gb free/total/max
Executing tests ........................................

PubOnly 0            10,000,000          5,430,509 msg/s       0.00 b/s
PubOnly 8            10,000,000          4,796,963 msg/s     36.60 mb/s
PubOnly 32           10,000,000          5,356,735 msg/s    163.47 mb/s
PubOnly 256          10,000,000          4,918,497 msg/s      1.17 gb/s
PubOnly 512          10,000,000          4,186,566 msg/s      2.00 gb/s
PubOnly 1024          1,000,000          3,331,125 msg/s      3.18 gb/s
PubOnly 4096            500,000          1,083,602 msg/s      4.13 gb/s
PubOnly 8192            100,000            533,976 msg/s      4.07 gb/s

PubSub 0             10,000,000          2,202,389 msg/s       0.00 b/s
PubSub 8             10,000,000          2,361,216 msg/s     18.01 mb/s
PubSub 32            10,000,000          2,155,187 msg/s     65.77 mb/s
PubSub 256           10,000,000          1,476,426 msg/s    360.46 mb/s
PubSub 512           10,000,000          1,276,072 msg/s    623.08 mb/s
PubSub 1024           1,000,000          1,044,820 msg/s   1020.33 mb/s
PubSub 4096             500,000            391,480 msg/s      1.49 gb/s
PubSub 8192             100,000            211,569 msg/s      1.61 gb/s

PubDispatch 0        10,000,000          2,308,575 msg/s       0.00 b/s
PubDispatch 8        10,000,000          2,259,020 msg/s     17.23 mb/s
PubDispatch 32       10,000,000          2,215,456 msg/s     67.61 mb/s
PubDispatch 256      10,000,000          1,548,035 msg/s    377.94 mb/s
PubDispatch 512      10,000,000          1,331,753 msg/s    650.27 mb/s
PubDispatch 1024      1,000,000          1,087,749 msg/s      1.04 gb/s
PubDispatch 4096        500,000            404,253 msg/s      1.54 gb/s
PubDispatch 8192        100,000            202,155 msg/s      1.54 gb/s

ReqReply 0               20,000             11,569 msg/s       0.00 b/s
ReqReply 8               20,000             11,500 msg/s     89.85 kb/s
ReqReply 32              10,000             11,621 msg/s    363.18 kb/s
ReqReply 256             10,000             11,638 msg/s      2.84 mb/s
ReqReply 512             10,000             11,567 msg/s      5.65 mb/s
ReqReply 1024            10,000             11,490 msg/s     11.22 mb/s
ReqReply 4096            10,000             11,374 msg/s     44.43 mb/s
ReqReply 8192            10,000             11,123 msg/s     86.90 mb/s

Latency 0     5,000     36 /  47.59 / 127    +/- 0.70  (microseconds)
Latency 8     5,000     41 /  46.28 / 129    +/- 0.70  (microseconds)
Latency 32    5,000     41 /  44.99 / 100    +/- 0.65  (microseconds)
Latency 256   5,000     41 /  45.10 / 158    +/- 0.65  (microseconds)
Latency 512   5,000     41 /  45.69 / 102    +/- 0.66  (microseconds)
Latency 1024  5,000     42 /  45.43 / 105    +/- 0.66  (microseconds)
Latency 4096  5,000     43 /  46.82 / 188    +/- 0.68  (microseconds)
Latency 8192  5,000     45 /  48.36 / 116    +/- 0.70  (microseconds)

Final memory usage is 973.10 mb / 982.50 mb / 14.22 gb free/total/max
```

src/examples/java/io/nats/examples/benchmark/NatsBench.java
```diff
@@ -213,10 +230,12 @@ public void run() {
                     nc.publish(subject, payload);
                     sent.incrementAndGet();
                 }
-                nc.flush(Duration.ofSeconds(5));
+                nc.flush(Duration.ofSeconds(15));
                 long end = System.nanoTime();
 
                 bench.addPubSample(new Sample(numMsgs, size, start.get(), end, nc.getStatistics()));
+
+                //System.out.println(nc.getStatistics());
```
derekcollison: Extra debug statement?

src/examples/java/io/nats/examples/benchmark/NatsBench.java
```diff
@@ -265,7 +284,7 @@ void adjustAndSleep(Connection nc) throws InterruptedException {
             
             // Flush small messages regularly
             if (this.size < 64 && count != 0 && count % 100_000 == 0) {
-                try {nc.flush(Duration.ofSeconds(5));}catch(Exception e){}
+                try {nc.flush(Duration.ofSeconds(15));}catch(Exception e){}
```
derekcollison: We still need this now?

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -27,39 +28,30 @@
...
-    private final ConcurrentLinkedQueue<NatsMessage> queue;
-    private final ConcurrentLinkedQueue<Thread> waiters;
+    private final LinkedBlockingQueue<NatsMessage> queue;
     private final Lock filterLock;
+    private final NatsMessage poisonPill;
 
-    MessageQueue(boolean singleReaderMode) {
-        this.queue = new ConcurrentLinkedQueue<>();
+    MessageQueue(boolean singleReaderMode, int publishHighwaterMark) {
+        this.queue = publishHighwaterMark > 0 ? new LinkedBlockingQueue<NatsMessage>(publishHighwaterMark) : new LinkedBlockingQueue<NatsMessage>();
```
derekcollison: What does default do with no publishHighwaterMark?

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -27,39 +28,30 @@
...
         this.running = new AtomicInteger(RUNNING);
         this.sizeInBytes = new AtomicLong(0);
         this.length = new AtomicLong(0);
 
+        // The poisonPill is used to stop poll and accumulate when the queue is stopped
+        this.poisonPill = new NatsMessage("_poison", null, NatsConnection.EMPTY_BODY, false);
```
derekcollison: This makes me feel like we want something to crash, do we want that or is this a `pauseSignal` or `blockSignal` or something like that?

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -27,39 +28,30 @@
...
-        this.waiters = new ConcurrentLinkedQueue<>();
         this.singleThreadedReader = singleReaderMode;
+    }
 
-        String os = System.getProperty("os.name");
-        os = os != null ? os.toLowerCase() : "";
-
-        // Windows does not like spin locks, see issue #224
-        if(os.contains("windows")) {
-            this.maxSpins = 0;
```
derekcollison: Interested in how this affects windows perf. MC was running benchmark on windows, fyi.

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -76,162 +68,104 @@ boolean isDraining() {
 
     void pause() {
         this.running.set(STOPPED);
-        signalAll();
+        this.poisonTheQueue();
```
derekcollison: Better name?

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -76,162 +68,104 @@ boolean isDraining() {
...
 
-            // Semi-spin for at most MAX_SPIN_TIME
-            if (timeoutNanos > MAX_SPIN_TIME) {
-                int count = 0;
-                while (this.isRunning() && (retVal = this.queue.poll()) == null && count < this.maxSpins) {
+    void put(NatsMessage msg) {
+        try {
+            this.queue.put(msg);
+        } catch (InterruptedException ie) {
+            // ok to ignore this
```
derekcollison: Why "ok", will this not put the msg on the queue? Is this assumed to always be terminating the process?

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -76,162 +68,104 @@ boolean isDraining() {
...
+        } else {
+            long nanos = timeout.toNanos();
 
-                    count++;
-                    LockSupport.parkNanos(SPIN_WAIT);
+            if (nanos != 0) {
+                msg = this.queue.poll(nanos, TimeUnit.NANOSECONDS);
+            } else {
+                while (this.isRunning()) {
+                    msg = this.queue.poll(100, TimeUnit.DAYS);
```
derekcollison: So will this potentially block for 100 days?

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -49,10 +50,13 @@
...
         ((CompletableFuture<Boolean>)this.stopped).complete(Boolean.TRUE); // we are stopped on creation
 
-        this.sendBuffer = new byte[connection.getOptions().getBufferSize()];
+        int bufSize = connection.getOptions().getBufferSize();
+        this.sendBuffer = new byte[bufSize];
 
-        outgoing = new MessageQueue(true);
-        reconnectOutgoing = new MessageQueue(true);
+        // Limit the messages that we can push on the queue to 5x the number we would send in a single batch
+        this.maxAccumulate = 1000;
```
derekcollison: Should this be based on total pending size in bytes vs msgs?

src/examples/java/io/nats/examples/benchmark/NatsBench.java
sasbury: intentional in case someone (or me) wants to see advanced stats.

src/examples/java/io/nats/examples/benchmark/NatsBench.java
sasbury: i don't know if we need the adjust but i was having issues with the flush on a slow network so I bumped it up

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: when set to 0 we have an unbounded queue

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: poisonpill was the descriptive term used by the people that said how to stop a linked blocking queue, so I kept it.

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: no more spinning so this is no longer needed. windows looks ok with the blocking queue

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: ultimately it is supposed to block forever, so anything i put here seems reasonable. The poison pill stops the wait if it has to.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
sasbury: ultimately it is limited by what is there, but this is a fixed value to prevent trying to read too much.

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: i don't want to spin too fast, so i figured bigger is better

src/examples/java/io/nats/examples/benchmark/NatsBench.java
derekcollison: ok might add a comment above stating to see more details uncomment below.

src/examples/java/io/nats/examples/benchmark/NatsBench.java
derekcollison: I mean the flush for small messages inline in the benchmark. Client should do the right thing now yes?

src/main/java/io/nats/client/impl/MessageQueue.java
derekcollison: That ever good?

src/main/java/io/nats/client/impl/MessageQueue.java
derekcollison: Gotcha..

src/main/java/io/nats/client/impl/MessageQueue.java
derekcollison: I mean to understand why MC was seeing a much different behavior then when you ran against demo. I assume you were not on Windows.

src/main/java/io/nats/client/impl/MessageQueue.java
derekcollison: Maybe comment explaining so.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
derekcollison: Read or Write?

src/examples/java/io/nats/examples/benchmark/NatsBench.java
sasbury: this was more for the pub/sub asymmetry. When I first wrote the lib i kept getting slow consumers. That might have gone away but I am tempted to leave this just in case, especially for windows and slower machines where i have seen bigger asymmetry issues.

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: only the reader uses 0 for subscribers, so it is checked a different way.

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: i think windows was slower to lock or maybe had lower resolution on the locksupport timeout. So things just "took longer" when I tried to spin wait.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
sasbury: write

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: will add comments throughout

derekcollison: ok

src/examples/java/io/nats/examples/benchmark/NatsBench.java
sasbury: made this a flag :-)

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: added comment

src/examples/java/io/nats/examples/benchmark/NatsBench.java
sasbury: this one was not used, the one in throttled benchmark is used now, but added a comment.

sasbury:
Added a CL with the comments/updated java doc.

src/main/java/io/nats/client/impl/MessageQueue.java
derekcollison: If possible we should test and see if we see same behavior I saw at MC.

src/examples/java/io/nats/examples/autobench/ThrottledBenchmark.java
```diff
@@ -68,6 +68,13 @@ public void execute(Options connectOptions) throws InterruptedException {
 
     }
 
+    /**
+     * Throttled benchmarks use an optional target send rate. This is to make sure the subscribers are not overrun,
+     * become slow consumers. The addition of this code was motivated by the assymetry in the java library between
```
derekcollison: spelling - asymmetry

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -34,8 +34,18 @@
     private final boolean singleThreadedReader;
     private final LinkedBlockingQueue<NatsMessage> queue;
     private final Lock filterLock;
+
+    // Poison pill is a graphic, but common term for an item that breaks loops or forces some other action
+    // In this class the poisonPill is used to break out of timed waits on the blocking queue.
```
derekcollison: Consistency with PoisonPill spelling/spaces.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -53,9 +53,16 @@
         int bufSize = connection.getOptions().getBufferSize();
         this.sendBuffer = new byte[bufSize];
 
-        // Limit the messages that we can push on the queue to 5x the number we would send in a single batch
+        // The value of 1000 is arbitrary, based on performance testing, a send buffer is also configured above and
+        // that size is used first, only if many small messages are being sent on a big buffer will this value
+        // come into play.
         this.maxAccumulate = 1000;
```
derekcollison: Still think tracking pending outbound bytes is best long term and benchmark program can watch that.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -53,9 +53,16 @@
...
         this.sendBuffer = new byte[bufSize];
 
-        // Limit the messages that we can push on the queue to 5x the number we would send in a single batch
+        // The value of 1000 is arbitrary, based on performance testing, a send buffer is also configured above and
+        // that size is used first, only if many small messages are being sent on a big buffer will this value
+        // come into play.
         this.maxAccumulate = 1000;
+        
+        // We limit the messages that we can push before blocking to 5x the number we would send in a single batch
+        // Again this is arbitrary, based on testing
```
derekcollison: Different machines, OS probably yield different results.

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -27,39 +28,40 @@
...
+
+    /**
+     * If publishHighwaterMark is set to 0 the underlying queue can grow forever. This is used by readers
+     * to prevent the read thread from blocking. If set to a number, the publish command will block, which provides
+     * backpressure on a publisher if the writer is slow to push things onto the network.
+     * @param singleReaderMode allows the use of "accumulate"
+     * @param publishHighwaterMark sets a limit on the size of the underlying queue
+     */
+    MessageQueue(boolean singleReaderMode, int publishHighwaterMark) {
+        this.queue = publishHighwaterMark > 0 ? new LinkedBlockingQueue<NatsMessage>(publishHighwaterMark) : new LinkedBlockingQueue<NatsMessage>();
```
kozlovic: Well, actually since it is a linked blocking queue, there is always a capacity, which in this case looking at javadoc, if no capacity is specified it is 2^31-1 :-). Nothing to change here, just mentioning it.

src/main/java/io/nats/client/impl/MessageQueue.java
kozlovic: @sasbury I understand what you said previously about this naming being used elsewhere, but personally I have used similar approach in the C client for a library level dispatcher and I coin the term of end message or control message. By the way, not sure if your API allows it, but I create an empty message (no subject, nothing). In your specific case, I wonder if you could even initialize it here as a static, singleton, message. The code that checks on the end message would check if this is this message (as opposed to user message). It may not be possible in Java since there is no pointer, you would have to make sure that the "equal" makes sure we know this is this message and not any user message.

src/main/java/io/nats/client/impl/MessageQueue.java
sasbury: did anyone take notes?

derekcollison: I can give you the info.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
sasbury: we can't use bytes with the linked blocking queue, since we aren't the ones waiting. The old code could make that distinction but the linked queue just blocks on the size in "slots".

We could muck with it, but I then we are back to the custom implementation.

Issue #197, opened 2018-12-06
Dispatch based on SID


film42:
It would be really nice if `NatsDispatcher` used the underlying `sid` to dispatch instead of a HashMap based on the subscription name. This would allow you to subscribe to the same topic multiple times. Now, this might not make that much sense from the client's point of view, but it does make sense from a "load balance" point of view.

As a sort of "hack" I have our servers subscribe to each topic `n` times, with a several second delay between each round of subscriptions. This allows our servers to warm up slowly. It's also useful for balancing traffic across larger/smaller vms. I'm currently upgrading from the pre-1.0 client, where I could pass a shared queue into the `subscribe` method. I think the dispatcher is really neat way to abstract the problem, but I'm unable to use it, because it doesn't use the underlying sid. 

Would you be open to review a pull request that changes this? I'm guessing it would end up being a breaking change, which is probably undesirable.

Otherwise, would you be interested in extending the connection interface to support:
```
public Subscription subscribe(String subject, String queueName, MessageQueue messageQueue);
```

...so I can support similar behavior to the pre-1.0 client?

film42:
Here's an example of how the `Connection` interface could be extended to support a user provided message queue.
```java
public interface Connection extends AutoCloseable {
    public Subscription subscribe(String subject);
    public Subscription subscribe(String subject, String queueName);
    // Expose a messageQueue option when subscribing.
    public Subscription subscribe(String subject, String queueName, MessageQueue messageQueue);
}

class NatsConnection implements Connection {
    // The createSubscription method would receive the messageQueue param.
    NatsSubscription createSubscription(String subject, String queueName, NatsDispatcher dispatcher, MessageQueue messageQueue) {
        // ...
        sub = new NatsSubscription(sid, subject, queueName, this, dispatcher, messageQueue);
        // ...
    }
}

class NatsSubscription extends NatsConsumer implements Subscription {
    // The constructor would receive the message queue and set it.
    NatsSubscription(String sid, String subject, String queueName, NatsConnection connection,
                     NatsDispatcher dispatcher, MessageQueue incoming) {
        // ...
        this.dispatcher = dispatcher;
        this.incoming = incoming;
        this.usingExternalMessageQueue = false;

        if (this.incoming != null) {
            // I don't love this idea, but it could be a way of avoiding the subscription calling
            // "this.incoming.pause()" when "invalidate()" is invoked.
            this.usingExternalMessageQueue = true
        }

        // If both params are missing, we can manage our own.
[...]

PR #251, opened 2019-07-15
Add support for subscriptions with their own message handlers for Dispatcher

film42:
This is a solution that satisfies what I was looking to get out of https://github.com/nats-io/nats.java/issues/197.

This adds a few new methods to the Dispatcher interface:

```java
public Subscription subscribe(String subject, MessageHandler handler);
public Subscription subscribe(String subject, String queue, MessageHandler handler);

public Dispatcher unsubscribe(Subscription subscription);
public Dispatcher unsubscribe(Subscription subscription, int after);
```

Calling `Dispatcher#subscribe` with a message handler does not treat duplicate subscriptions as a no-op. It will instead allow duplicate subscriptions to be created and managed by the caller. With this support, the Dispatcher can be used in a similar manner to v1 of this library. Behavior does not change for existing Dispatcher methods.

Example:
```java
Dispatcher d = nc.createDispatcher((msg) -> {});

// Subscribe to different topics using our own message handlers to allow duplicate
// subscriptions to be created.

Subscription s1 =
    d.subscribe("subject", (msg) -> { printMsg(msg, "processed by first subscription"); });
d.subscribe("subject", (msg) -> { printMsg(msg, "processed by second subscription"); });
d.subscribe("done", (msg) -> { print("all done!"); });

// Publish messages.

nc.publish("subject", "1");
nc.publish("subject", "2");
nc.publish("done", "done");

// Output:
// Msg: 1, processed by first subscription
// Msg: 1, processed by second subscription
// Msg: 2, processed by first subscription
// Msg: 2, processed by second subscription
// Msg: done, all done!

// Now we will unsubscribe from the first "subject" subscription.

d.unsubscribe(s1);

// Publish messages.

nc.publish("subject", "1");
nc.publish("subject", "2");
nc.publish("done", "done");

// Output:
// Msg: 1, processed by second subscription
// Msg: 2, processed by second subscription
// Msg: done, all done!
```

I'm sure this branch could use a few more tests and a bit more polish, but feel free to start reviewing! Thanks.

cc @sasbury

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsDispatcher.java
+++ b/src/main/java/io/nats/client/impl/NatsDispatcher.java
@@ -26,7 +26,7 @@ import io.nats.client.MessageHandler;
 class NatsDispatcher extends NatsConsumer implements Dispatcher, Runnable {
 
     private MessageQueue incoming;
-    private MessageHandler handler;
+    private MessageHandler defaultHandler;
 
     private Future<Boolean> thread;
     private final AtomicBoolean running;
@@ -35,7 +35,7 @@ class NatsDispatcher extends NatsConsumer implements Dispatcher, Runnable {
 
     // We will use the subject as the key for subscriptions that use the
     // default handler.
-    private Map<String, NatsSubscription> subscriptions;
+    private Map<String, NatsSubscription> subscriptionsUsingDefaultHandler;
     // We will use the SID as the key. Sicne these subscriptions provide
     // their own handlers, we allow duplicates. There is a subtle but very
     // important difference here.
@@ -48,9 +48,9 @@ class NatsDispatcher extends NatsConsumer implements Dispatcher, Runnable {
 
     NatsDispatcher(NatsConnection conn, MessageHandler handler) {
         super(conn);
-        this.handler = handler;
+        this.defaultHandler = handler;
         this.incoming = new MessageQueue(true);
-        this.subscriptions = new ConcurrentHashMap<>();
+        this.subscriptionsUsingDefaultHandler = new ConcurrentHashMap<>();
         this.subscriptionsWithHandlers = new ConcurrentHashMap<>();
         this.subscriptionHandlers = new ConcurrentHashMap<>();
         this.running = new AtomicBoolean(false);
@@ -88,7 +88,7 @@ class NatsDispatcher extends NatsConsumer implements Dispatcher, Runnable {
                     sub.incrementDeliveredCount();
                     this.incrementDeliveredCount();
 
-                    MessageHandler currentHandler = this.handler;
+                    MessageHandler currentHandler = this.defaultHandler;
                     MessageHandler customHandler = this.subscriptionHandlers.get(sub.getSID());
                     if (customHandler != null) {
                         currentHandler = customHandler;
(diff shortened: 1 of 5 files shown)


src/main/java/io/nats/client/impl/NatsDispatcher.java
```diff
@@ -164,40 +204,68 @@ public Dispatcher subscribe(String subject, String queueName) {
         if (queueName == null || queueName.length() == 0) {
```
sasbury: Can you add the subject == null check to this one, the others have it. Or we could move that check to the impl and just check queues and handlers as necessary

src/main/java/io/nats/client/impl/NatsDispatcher.java
```diff
@@ -164,40 +204,68 @@ public Dispatcher subscribe(String subject, String queueName) {
...
     }
 
+    public Dispatcher unsubscribe(Subscription subscription) {
+        return this.unsubscribe(subscription, -1);
+    }
+
     public Dispatcher unsubscribe(String subject, int after) {
         if (!this.running.get()) {
-            throw new IllegalStateException("Dispatcfher is closed");
+            throw new IllegalStateException("Dispatcher is closed");
```
sasbury: :-P

src/main/java/io/nats/client/impl/NatsDispatcher.java
```diff
@@ -207,8 +275,32 @@ public Dispatcher unsubscribe(String subject, int after) {
...
+
+        return this;
+    }
+
+    public Dispatcher unsubscribe(Subscription subscription, int after) {
+        if (subscription.getDispatcher() != this) {
+            throw new IllegalStateException("Subscription is not managed by this Dispatcher");
+        }
+
+        if (this.subscriptionsUsingDefaultHandler.get(subscription.getSubject()) != null) {
```
sasbury: I don' think this is right. You could have "foo" under the default handler and "foo" under a custom handler. Instead you should check if the sid is in the with handler map and if that is false, then do this check.

src/main/java/io/nats/client/impl/NatsDispatcher.java
film42: It's there, I think the github collapsed diff hid it from view, but it's there.

src/main/java/io/nats/client/impl/NatsDispatcher.java
film42: Derp! Great catch.

film42: I was trying to be clever here. We do end up checking against `subscriptionsWithHandlers` a few lines down (mimicking behavior of `subscribe(String, int)`), so I think we might actually be able to remove the entire conditional.

sasbury: the problem is if i subscribe("foo") and s = subscribe("foo", mh), then try to unsubscribe(s) this will throw an illegal state. But it should be ok. I think you can just remove the check.

film42: We should add a test around that example, because I think it could pose an issue in `Dispatcher#remove`. We should be fine if we do something like:
```java
    void remove(NatsSubscription sub) {
        if (this.subscriptionsWithHandlers.remove(sub.getSID()) != null) {
            this.subscriptionHandlers.remove(sub.getSID());
        } else {
            NatsSubscription s = this.subscriptionsUsingDefaultHandler.get(sub.getSubject());
            if (s.getSID() == sub.getSID()) {
                this.subscriptionsUsingDefaultHandler.remove(sub.getSubject());
            }
        }
    }
```
I'll push a commit up with comments if you're :+1: .

sasbury: +1

src/main/java/io/nats/client/impl/NatsDispatcher.java
```diff
@@ -207,8 +285,28 @@ public Dispatcher unsubscribe(String subject, int after) {
...
+
+        if (sub != null) {
+            this.connection.unsubscribe(sub, after); // Connection will tell us when to remove from the map
+        }
+
+        return this;
+    }
+
+    public Dispatcher unsubscribe(Subscription subscription, int after) {
+        if (subscription.getDispatcher() != this) {
```
sasbury: last thing - this needs
```
if (!this.running.get()) {
            throw new IllegalStateException("Dispatcher is closed");
        }

        if (isDraining()) { // No op while draining
            return this;
        }
```
at the top like the previous method

film42: Done! Good find!

PR #254, opened 2019-07-18
Converted password/user/token to char arrays

sasbury:
This ended up being pretty invasive since it required carrying char arrays throught the other code to make sure we don't "toString" the password.
All tests are running but some of the API is now deprecated.


sasbury:
Messed up git branch stuff

PR #296, opened 2020-02-19
Update the PGP signing key

philpennock:
Key would expire in July of this year; avoid having builds fail for trying to sign with an expired key, use the updated keyring with the new expiration date in 2022.


sasbury:
I haven't poked at this in a while, is this going to break local deploys or is the encoding key the same?

philpennock:
> I haven't poked at this in a while, is this going to break local deploys or is the encoding key the same?

It shouldn't.  At most, you'll need to `gpg --refresh-key info@nats.io` to pull the public key updates in.  I carefully avoided anything which might need new private key parts.  The private key is updated here because that export includes the applicable "certificate" style data from the public keyring.

PR #301, opened 2020-04-03
modifications for race condition on close #300

### LINKED ISSUE #300: potential issue with race condition on close
potential issue with race condition on close

RichardHightower:
modifications for race condition on close #300


src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -49,7 +50,9 @@
      * @param publishHighwaterMark sets a limit on the size of the underlying queue
      */
     MessageQueue(boolean singleReaderMode, int publishHighwaterMark) {
-        this.queue = publishHighwaterMark > 0 ? new LinkedBlockingQueue<NatsMessage>(publishHighwaterMark) : new LinkedBlockingQueue<NatsMessage>();
+        this.queue = new LinkedBlockingQueue<NatsMessage>();
```
RichardHightower, line 53: so what is publishHighwaterMark for?

ColinSullivan1: It looks like a publish buffer, thats set to infinite during reconnect.  I don't believe there's an analogous buffer in the other clients.

PR #306, opened 2020-04-13
Discard messages when outgoing queue is full

wjbakker:
Fast producers may fully fill the outgoing message queue. Once this
happens, a producer will block until there is space for new messages.

This change adds the option to set the maximum number of messages in the
outgoing queue. This option can be set with property
`io.nats.client.outgoingqueue.maxmessages`, default `5000`.

Moreover, this change adds the option to discard
messages when the outgoing queue is full. This option can be set with
property `io.nats.client.outgoingqueue.discardwhenfull`, default `false`.
Clients can be notified of such event by implementing the
`messageDiscarded` callback in the `ErrorListener` interface.

Example usage:
```java
ErrorListener handler = new ErrorListener() {
    public void messageDiscarded(Connection conn, Message msg) {
        ...
    }
}

Options options = new Options.Builder().
	maxMessagesInOutgoingQueue(10_000).
	discardMessagesWhenOutgoingQueueFull().
	errorListener(handler)
	build();

Connection nc = Nats.connect(options);
```

### DIFF
--- a/src/test/java/io/nats/client/impl/ErrorListenerTests.java
+++ b/src/test/java/io/nats/client/impl/ErrorListenerTests.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
 import java.time.Duration;
+import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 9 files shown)


src/test/java/io/nats/client/impl/ErrorListenerTests.java
```diff
@@ -231,4 +232,31 @@ public void testExceptionInExceptionHandler() throws Exception {
             }
         }
     }
+
+    @Test
+    public void testDiscardedMessage() throws Exception {
```
sasbury, line 237: Can you update to shut down the test server after connect, so the outgoing queue is clearly being used, rather than having it work because of timing or the test server not actually reading.

wjbakker: Hi Stephen, thanks for the quick review. I was able to confirm that this test case is indeterministic.

I discovered a corner case on flushing the connection. The flush command pushes a ping message on the queue. If the queue is full, this ping message may trigger an IllegalStateException (existing behaviour, when it cannot be offered within 5 seconds), or may be discarded (this pull request, when it cannot be offered directly). I will look into keeping the old behaviour for internal messages.

wjbakker: Hi Stephen,

I have changed the implementation to make sure that upon a full outgoing queue:
* internal messages wait for the queue to become available
* messages from fast producers are discarded immediately 

Tests have been added.

sasbury:
merged but found another issue in the test that I think is related to pings going out, i think it is fixed, will see how the CI does

PR #318, opened 2020-06-17
Printing trace when NUID initialization takes long

matthiashanel:
Finding that one took a while ...

Signed-off-by: Matthias Hanel <mh@synadia.com>

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -156,7 +156,15 @@ class NatsConnection implements Connection {
         this.serverAuthErrors = new HashMap<>();
 
         this.nextSid = new AtomicLong(1);
+        long start = System.nanoTime();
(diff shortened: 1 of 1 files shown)


matthiashanel:
> I wonder if we should add this to the connect code when TLS is used, that is the other place people run into this.

@sasbury would you agree with this: Since the nuid code is going to be executed in either case, I assume adding a trace around ssl would basically cause the /dev/random delay to be logged twice. So maybe this trace already covers it?

Does the delay for ssl happen in upgradeToSecure or during ssl context creation?

sasbury:
I think the delay only happens the first time, i haven't traced it, but i think either can happen first so it may be worth logging on both. Upgrade to secure would be the place.

matthiashanel:
I'll add a trace for timing around upgradeToSecure as well and update this pr

matthiashanel:
For the TLS upgrade it seemed appropriate to trace all the time.

Issue #321, opened 2020-07-10
Add a simple way to connect using NATS 2.0 credentials.


ColinSullivan1:
e.g.

`Nats.Connect("localhost", my.creds);`

olicuzo:
Hi, I would like to help on this if possible, I'm getting to know the library and contributing seems to be the best way.

matthiashanel:
@olicuzo Thanks for your contribution. 
@ColinSullivan1 I'll close this due to the merged change

PR #323, opened 2020-07-14
Nats.connect v2 credentials

olicuzo:
Overloads the `Nats.connect` method by adding `AuthHandler` as an argument.

Addresses https://github.com/nats-io/nats.java/issues/321

### DIFF
--- a/src/main/java/io/nats/client/Nats.java
+++ b/src/main/java/io/nats/client/Nats.java
@@ -132,6 +132,28 @@ public class Nats {
     }
 
     /**
+     * Connect to the specified URL with the specified username, password and default options.
+     *
+     * <p>This is a synchronous call, and the connection should be ready for use on return
+     * there are network timing issues that could result in a successful connect call but
+     * the connection is invalid soon after return, where soon is in the network/thread world.
+     *
+     * <p>If the connection fails, an IOException is thrown
+     *
+     * <p>See {@link Nats#connect(Options) connect(Options)} for more information on exceptions.
+     *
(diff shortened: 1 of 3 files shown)


olicuzo:
It seems that the build fails with some config issue: `iv undefined`

src/main/java/io/nats/client/Nats.java
```diff
@@ -131,6 +131,28 @@ public static Connection connect(String url) throws IOException, InterruptedExce
...
+     *
+     * <p>This is a synchronous call, and the connection should be ready for use on return
+     * there are network timing issues that could result in a successful connect call but
+     * the connection is invalid soon after return, where soon is in the network/thread world.
+     *
+     * <p>If the connection fails, an IOException is thrown
+     *
+     * <p>See {@link Nats#connect(Options) connect(Options)} for more information on exceptions.
+     *
+     * @param url the url of the server, ie. nats://localhost:4222
```
matthiashanel, line 145: To avoid this not being used when you want multiple urls we should mention that url can be a comma separated list of servers as well.
nats://localhost:4222,nats://localhost:4223
createConnection splits around ,

olicuzo: updated javadoc

matthiashanel:
@olicuzo Thanks for submitting this PR and improving the doc. I ran the tests with your changes and they pass with flying colors. (you don't have access to a certain key, which is why they failed for you) 
@ColinSullivan1 The change looks good to me.

Issue #326, opened 2020-07-24
JetStream NEXT Not Working


brimworks:
## FIX

https://github.com/nats-io/nats.java/pull/327/files

NOTE: There is no way to "throw an exception" if we can't match up a response future, so the best option I could think of was to print an error to stdout (maybe it should be printed to stderr though?). Other ideas are welcomed.

This implementation is based on @ColinSullivan1 's response in #jetstream slack channel:

> The client generates a SID (could random, could be increasing int) and associates it with the SUB protocol. When the server matched the subscription it delivers a MSG proto with the same SID so the client can properly associate the message with a subscription.

## IDEA

Not to be "cheeky", but given that all existing clients have to change to support subject name remapping (and all clients have to switch back to "old mode" to make this work). I'd advocate for changing the subject name remapping mechanism in a way that is backwards compatible with existing clients (minus retaining the original subject name) so as to avoid these sort of surprises. Since jetstream is still in "beta", it should be possible to change this mechanism.

One idea that I have is to add a new "operation name" to the NATS protocol which indicates the subject override.

```
ORM <subject-override>
MSG <subject> <sid> [reply-to] <#bytes>\r\n[payload]\r\n
```

I believe existing clients would ignore this new operation name and would work fine, but the object name remapping would only work after upgrading the client. It would also allow for using the more efficient "new style" subscriptions.

## HISTORY

I created this PR to demonstrate what I mean: https://github.com/nats-io/nats.java/pull/325

I talked to R.I. Pienaar on #jetstream slack channel, and the expected behavior is for "pull" clients of jetstream to obtain the "next" element from the consumer stream via the "reply"... so I believe this code should work and not timeout.

PR #327, opened 2020-07-25
Subject Remapping Fix

brimworks:
Fixes: https://github.com/nats-io/nats.java/issues/326

Allow for JetStream subject remapping by tracking outstanding requests based on the SID (only works for "old style").

The `build.gradle` fix was useful when using Visual Studio Code which failed due to invalid `dependsOn` when `useSigning=false`.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -1009,8 +1009,10 @@ class NatsConnection implements Connection {
         Future<Message> incoming = this.request(subject, body);
         try {
             reply = incoming.get(timeout.toNanos(), TimeUnit.NANOSECONDS);
-        } catch (ExecutionException | TimeoutException e) {
-            reply = null;
+        } catch (TimeoutException e) {
+            incoming.cancel(true);
+        } catch (Throwable e) {
+            throw new AssertionError(e);
         }
 
         return reply;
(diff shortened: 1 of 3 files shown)


brimworks:
Unfortunately, this approach won't work in the face of a "reply timeout".

Specifically, if you write code that repeatedly polls for the `$JS.API.CONSUMER.MSG.NEXT.SUBJECT1.CONSUMER1` subject, the NATS server doesn't know that you set a 5 second timeout and thus it will eventually send a message with the session id set to the FIRST reply instance which is a "no-op" if you are on the 2nd or 3rd message poll.

brimworks:
After thinking about this more I realize that if we UNSUBscribe on reply timeout we should be fine... although a race condition does still exist, since a message might be delivered after the future has a timeout but before the UNSUBscribe is received by the nats-server.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1059,11 +1061,23 @@ public Message request(String subject, byte[] body, Duration timeout) throws Int
         String responseToken = getResponseToken(responseInbox);
         CompletableFuture<Message> future = new CompletableFuture<>();
 
-        responses.put(responseToken, future);
+        if (!oldStyle) {
```
ColinSullivan1, line 1064: nitpick, could simplify with if / else and move `statistics.incrementOutstandingRequests();` earlier.

ColinSullivan1:
First, thank you for the fix!  Much appreciated!  

There is that race condition, but imo this should be fine.  This particular case should only occur with JetStream.  Caveat - I don't thoroughly know this client, but looking through it I didn't spot any issues beyond said race condition.  I'm OK moving forward with this fix.  btw, this fix was completely green (100% success) testing locally on my machine.

@sasbury,  it'd be great if you could give a quick review of this.

ColinSullivan1:
LGTM.  Thanks!

sasbury:
the unsub 1 should also be tracked on teh server (if my memory serves) so the race condition may not be that bad. I didn't see anything leap out at me, not sure my mind wrapped around old style R/R but seems ok to me

PR #330, opened 2020-08-03
[FIX] Reconnect wait was not being honored

### LINKED ISSUE #290: Reconnect wait logic in java only waits if the current and last server tried are the same
https://github.com/nats-io/nats.java/blob/7f7fe824422b6966c8147bc5f028f228be4ed6ad/src/main/java/io/nats/client/impl/NatsConnection.java#L275-L278

This differs from go/other implementations.

### LINKED ISSUE #329: Issue in discovered server list in case of DNS entry
In the case of DNS as nats endpoint, while NATS producer/consumer discovering list the servers, Noticing  DNS entry also one of the servers discovered Servers list. This could an issue in case of if want to route the traffic from one cluster to another cluster in case of DR activity.

**Sample Log:**
``` Cnnected/ReConnected NATS Server: [tls://dns:4222, server1:4222, server2:4222,server3:4222, server4:4222, server5:4222] - Event :nats: connection opened ```

So the issue here, for example, if all the servers in the NATS clusters went down in the respective cluster. the produce/consumer is thinking DNS entry has a valid server and it keeps on trying and it is hanging there itself. and somehow producer/consumer always considering this DNS entry is healthy even though no active brokers in the clusters.

### PR DESCRIPTION
**Pull request by ColinSullivan1**
The reconnect wait option was not being honored.  Adjust the reconnect logic to honor it.

Partially resolves #329.
Resolves #290.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Timer;
 import java.util.TimerTask;
@@ -274,7 +275,9 @@ class NatsConnection implements Connection {
         boolean doubleAuthError = false;
 
         while (!isConnected() && !isClosed() && !this.isClosing()) {
-            Collection<String> serversToTry = buildServerList();
+            List<String> serversToTry = buildServerList();
+            lastServer = serversToTry.get(serversToTry.size()-1);
+            boolean shouldWait = false;
 
             for (String server : serversToTry) {
                 if (isClosed()) {
@@ -283,19 +286,17 @@ class NatsConnection implements Connection {
 
                 connectError.set(""); // reset on each loop
 
-                if (server.equals(lastServer)) {
-                    this.reconnectWaiter = new CompletableFuture<>();
-                    waitForReconnectTimeout();
-                }
-
                 if (isDisconnectingOrClosed() || this.isClosing()) {
                     break;
                 }
 
+                if (server.equals(lastServer)) {
+                    shouldWait = true;
+                } 
+
                 updateStatus(Status.RECONNECTING);
 
                 tryToConnect(server, System.nanoTime());
-                lastServer = server;
                 tries++;
 
                 if (maxTries > 0 && tries >= maxTries) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/main/java/io/nats/client/impl/NatsConnection.java [hunk 1]
```diff
@@ -283,19 +286,17 @@ void reconnect() throws InterruptedException {
...
-                    this.reconnectWaiter = new CompletableFuture<>();
-                    waitForReconnectTimeout();
-                }
-
                 if (isDisconnectingOrClosed() || this.isClosing()) {
                     break;
                 }
 
+                if (server.equals(lastServer)) {
+                    shouldWait = true;
```
matthiashanel, line 294: if we manage to connect, should shouldWait be set to false?
line 305

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -325,6 +326,11 @@ void reconnect() throws InterruptedException {
             if (maxTries > 0 && tries >= maxTries) {
                 break;
             }
+
+            if (shouldWait) {
```
matthiashanel, line 330: shouldn't this happen inside the for loop? line 318 when we did not succeed or exceed the limits?

sasbury: it is in the while loop, but i could see it going either way for sure

sasbury: might be cleaner to move it up

src/main/java/io/nats/client/impl/NatsConnection.java
sasbury: i think so, otherwise it think we wait if we happen to connect to the last server

ColinSullivan1: Yep, good catch, thanks.  I'll fix it, and move the delay up per other comments.

ColinSullivan1:
Thanks for the reviews.  I moved the check into the loop.

ColinSullivan1:
I have no clue why but can't trigger a re-run for the single flapping CI test.  Safe to merge IMO.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -283,19 +285,13 @@ void reconnect() throws InterruptedException {
...
-                }
-
                 if (isDisconnectingOrClosed() || this.isClosing()) {
                     break;
                 }
 
                 updateStatus(Status.RECONNECTING);
 
                 tryToConnect(server, System.nanoTime());
-                lastServer = server;
```
matthiashanel, line 298: Do I read this right, removing this line will bring the client in lie with other clients and fix #290

ColinSullivan1: It should, yes.

Issue #324, opened 2020-07-22
NATS client leaves dangling threads after close, if NATS server was down when drain & close were invoked


adamkotwasinski:
Hey, first and foremost thanks for creating NATS Java client.

I found that NATS connection leaves dangling non-daemon threads after `.drain` & `.close` methods have been invoked, potentially leaking resources.

Very simple example (run it with NATS server running):
```
    public static void main(final String[] args) throws Exception {

        // We connect without exceptions.
        final Connection connection = Nats.connect();

        // Now kill the server and kick off shutdown.
        System.in.read(new byte[2]);
        LOG.info("shutting down (server is dead now)");
        try {
            final CompletableFuture<Boolean> drainFuture = connection.drain(Duration.ofSeconds(5));
            LOG.info("drain: {}", drainFuture.get()); // Throws, as expected.
        }
        catch (final Exception e) {
            LOG.error("drain", e);
        }
        finally {
            connection.close(); // Returns okay.
        }

        // Main thread finishes.
        LOG.info("finished");
    }
```
what results in main thread finishing, but us having non-daemon `NATS Connection Timer` thread and one of (potentially daemon) worker threads:
<img width="325" alt="Screen Shot 2020-07-22 at 13 50 48" src="https://user-images.githubusercontent.com/17969711/88227538-9ccf3100-cc22-11ea-84c8-2f4cf4ddb42d.png">

After some time, the worker thread finally finishes, but it turns out that Connection Timer was never cleaned up:
<img width="328" alt="Screen Shot 2020-07-22 at 13 50 18" src="https://user-images.githubusercontent.com/17969711/88227628-ba9c9600-cc22-11ea-99b1-7355550c3e40.png">

I think the cause is that it's possible for `close()` to end early in https://github.com/nats-io/nats.java/blob/2.6.8/src/main/java/io/nats/client/impl/NatsConnection.java#L624 (after all, we wanted to drain, but just failed).

Changing worker threads to daemons would not help, as the timer itself is not a daemon (https://github.com/nats-io/nats.java/blob/2.6.8/src/main/java/io/nats/client/impl/NatsConnection.java#L462). And anyways, that would not be viable for users that want to initialize NATS connection multiple times over lifetime of VM (e.g. in application containers).

matthiashanel:
@adamkotwasinski , thanks for submitting this issue. I have verified it and am now weighing my options for a fix.

matthiashanel:
[...]

PR #331, opened 2020-08-05
[Fixed] close connection when flush fails during drain.

matthiashanel:
Fixeds #324

Signed-off-by: Matthias Hanel <mh@synadia.com>


matthiashanel:
@ColinSullivan1 had to change the unit test. running in gradlew there where more than one timer.

PR #343, opened 2020-09-03
README fixes

wallyqs:


ColinSullivan1 (approved):
LGTM.  If you'd like, there are places where `java-nats client` can be replaced with NATS java client, but doesn't need to be part of this PR.

wallyqs:
ok! will update as well in this PR

PR #346, opened 2020-10-09
Fix build and run tests on newer java versions.

jameshilliard:

PR #348, opened 2020-10-16
Fix before_install formatting.

jameshilliard:
Seems I botched the syntax in #346 when making `before_install` conditional on `"${TRAVIS_SECURE_ENV_VARS}" == "true"` so that pull requests from repos without the secure env vars stopped failing, this should hopefully be correct, this doesn't get run on forks or pull requests so it's a bit tricky for me to fully test.


ColinSullivan1 (approved):
LGTM for the "Fix before_install formatting" commit.

Issue #347, opened 2020-10-11
Jnats should use non-blocking NIO2 Sockets instead of legacy blocking IO.


jameshilliard:
Currently it appears Jnats is using legacy blocking socket streams for the `SocketDataPort` implementation, this should probably be migrated to use a non-blocking NIO2 [AsynchronousSocketChannel](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/AsynchronousSocketChannel.html) for improved performance/latency along with thread safety. I'll probably try and do this myself once #346 is merged.

https://github.com/nats-io/nats.java/blob/031b245f2e66b16c9153cafe0c3a42b374040597/src/main/java/io/nats/client/impl/SocketDataPort.java#L37-L41

Is there any reason that Jnats has avoided using NIO2 sockets that I should be aware of? From my initial reading of the code converting to NIO2 sockets should be fairly straight forward as Jnats seems to use the NIO based `ByteBuffer`'s already for messages which is the same buffer format NIO2 based sockets expect for reading/writing.

brimworks:
Track this need here: https://github.com/nats-io/nats.java/issues/317

Basically, we all want an async client, so let's track that need in the above issue.

Thanks!

PR #354, opened 2020-11-03
Jetstream API

ColinSullivan1:
## Feature:  The Jetstream API

Documentation will be incomplete until the API is finalized.

Resolves #352 

### TODO
- [x] Consumer Configuration
- [x] ProducerOptions
- [x] SubscriberOptions
- [x] Ack
- [x] Publish APIs
- [x] Test Harness Updates
- [x] Msg APIs
- [x] Subscribe APIs
- [x] Pull Consumer

PR #356, opened 2020-11-04
Use ByteBuffer's for serialization/deserialization.

jameshilliard:
Some more groundwork for #347.

### DIFF
--- a/src/main/java/io/nats/client/NUID.java
+++ b/src/main/java/io/nats/client/NUID.java
@@ -16,6 +16,7 @@
 package io.nats.client;
 
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.util.Random;
@@ -50,7 +51,7 @@ public final class NUID {
     private final Random prand;
 
     // Instance fields
-    char[] pre;
+    byte[] pre;
     private long seq;
     private long inc;
 
@@ -87,7 +88,7 @@ public final class NUID {
 
         seq = nextLong(prand, maxSeq);
         inc = minInc + nextLong(prand, maxInc - minInc);
-        pre = new char[preLen];
+        pre = new byte[preLen];
         for (int i = 0; i < preLen; i++) {
             pre[i] = '0';
         }
@@ -106,7 +107,16 @@ public final class NUID {
      *
      * @return the next NUID string from this instance.
      */
-    public final synchronized String next() {
+    public final String next() {
+        return StandardCharsets.US_ASCII.decode(getInstance().nextBuffer()).toString();
+    }
+
+    /**
+     * Generate the next NUID ByteBuffer from this instance.
+     *
+     * @return the next NUID ByteBuffer from this instance.
+     */
+    public final synchronized ByteBuffer nextBuffer() {
         // Increment and capture.
         seq += inc;
         if (seq >= maxSeq) {
@@ -115,16 +125,16 @@ public final class NUID {
         }
 
         // Copy prefix
-        char[] b = new char[totalLen];
+        byte[] b = new byte[totalLen];
         System.arraycopy(pre, 0, b, 0, preLen);
 
         // copy in the seq in base36.
         int i = b.length;
         for (long l = seq; i > preLen; l /= base) {
             i--;
-            b[i] = digits[(int) (l % base)];
+            b[i] = (byte) (l % base + (byte)'0');
         }
-        return new String(b);
+        return ByteBuffer.wrap(b);
     }
 
     // Resets the sequntial portion of the NUID
@@ -145,7 +155,7 @@ public final class NUID {
         srand.nextBytes(cb);
 
         for (int i = 0; i < preLen; i++) {
-            pre[i] = digits[(cb[i] & 0xFF) % base];
+            pre[i] = (byte)((cb[i] & 0xFF) % base + (byte)'0');
         }
     }
 
@@ -170,7 +180,7 @@ public final class NUID {
     /**
      * @return the pre
      */
-    char[] getPre() {
+    byte[] getPre() {
         return pre;
     }
 
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -16,21 +16,12 @@ package io.nats.client.impl;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
-import java.nio.charset.StandardCharsets;
+import java.nio.charset.*;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Timer;
-import java.util.TimerTask;
+import java.util.*;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
@@ -73,16 +64,18 @@ class NatsConnection implements Connection {
     static final byte LF = 0x0A;
     static final byte[] CRLF = { CR, LF };
 
-    static final String OP_CONNECT = "CONNECT";
-    static final String OP_INFO = "INFO";
-    static final String OP_SUB = "SUB";
-    static final String OP_PUB = "PUB";
-    static final String OP_UNSUB = "UNSUB";
-    static final String OP_MSG = "MSG";
-    static final String OP_PING = "PING";
-    static final String OP_PONG = "PONG";
-    static final String OP_OK = "+OK";
-    static final String OP_ERR = "-ERR";
+    static final ByteBuffer OP_CONNECT = ByteBuffer.wrap(new byte[] { 'C', 'O', 'N', 'N', 'E', 'C', 'T' });
+    static final ByteBuffer OP_INFO = ByteBuffer.wrap(new byte[] { 'I', 'N', 'F', 'O' });
+    static final ByteBuffer OP_SUB = ByteBuffer.wrap(new byte[] { 'S', 'U', 'B' });
+    static final ByteBuffer OP_PUB = ByteBuffer.wrap(new byte[] { 'P', 'U', 'B' });
+    static final ByteBuffer OP_UNSUB = ByteBuffer.wrap(new byte[] { 'U', 'N', 'S', 'U', 'B' });
+    static final ByteBuffer OP_MSG = ByteBuffer.wrap(new byte[] { 'M', 'S', 'G' });
+    static final ByteBuffer OP_PING = ByteBuffer.wrap(new byte[] { 'P', 'I', 'N', 'G' });
+    static final ByteBuffer OP_PING_LINE = ByteBuffer.wrap(new byte[] { 'P', 'I', 'N', 'G', '\r', '\n' });
+    static final ByteBuffer OP_PONG = ByteBuffer.wrap(new byte[] { 'P', 'O', 'N', 'G'});
+    static final ByteBuffer OP_PONG_LINE = ByteBuffer.wrap(new byte[] { 'P', 'O', 'N', 'G', '\r', '\n' });
+    static final ByteBuffer OP_OK = ByteBuffer.wrap(new byte[] { '+', 'O', 'K' });
+    static final ByteBuffer OP_ERR = ByteBuffer.wrap(new byte[] { '-', 'E', 'R', 'R' });
 
     private Options options;
 
@@ -109,13 +102,13 @@ class NatsConnection implements Connection {
 
     private AtomicReference<NatsServerInfo> serverInfo;
 
-    private Map<String, NatsSubscription> subscribers;
-    private Map<String, NatsDispatcher> dispatchers; // use a concurrent map so we get more consistent iteration
+    private Map<ByteBuffer, NatsSubscription> subscribers;
+    private Map<ByteBuffer, NatsDispatcher> dispatchers; // use a concurrent map so we get more consistent iteration
                                                      // behavior
-    private Map<String, CompletableFuture<Message>> responses;
+    private Map<ByteBuffer, CompletableFuture<Message>> responses;
     private ConcurrentLinkedDeque<CompletableFuture<Boolean>> pongQueue;
 
-    private String mainInbox;
+    private ByteBuffer mainInbox;
     private AtomicReference<NatsDispatcher> inboxDispatcher;
     private Timer timer;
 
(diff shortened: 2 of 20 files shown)


src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -73,16 +64,16 @@
...
+    static final ByteBuffer OP_CONNECT = ByteBuffer.wrap(new byte[] { 'C', 'O', 'N', 'N', 'E', 'C', 'T' });
+    static final ByteBuffer OP_INFO = ByteBuffer.wrap(new byte[] { 'I', 'N', 'F', 'O' });
+    static final ByteBuffer OP_SUB = ByteBuffer.wrap(new byte[] { 'S', 'U', 'B' });
+    static final ByteBuffer OP_PUB = ByteBuffer.wrap(new byte[] { 'P', 'U', 'B' });
+    static final ByteBuffer OP_UNSUB = ByteBuffer.wrap(new byte[] { 'U', 'N', 'S', 'U', 'B' });
+    static final ByteBuffer OP_MSG = ByteBuffer.wrap(new byte[] { 'M', 'S', 'G' });
+    static final ByteBuffer OP_PING = ByteBuffer.wrap(new byte[] { 'P', 'I', 'N', 'G' });
+    static final ByteBuffer OP_PONG = ByteBuffer.wrap(new byte[] { 'P', 'O', 'N', 'G' });
+    static final ByteBuffer OP_OK = ByteBuffer.wrap(new byte[] { '+', 'O', 'K' });
+    static final ByteBuffer OP_ERR = ByteBuffer.wrap(new byte[] { '-', 'E', 'R', 'R' });
```
scottf, line 76: Since speed is not crucial at this initialization point it might be more readable to just do 
```
static final ByteBuffer OP_ERR = ByteBuffer.wrap("-ERR".getBytes());
```

jameshilliard: I did it this way mostly since it's pretty obvious from the variables names what they are, and since I'm wasn't sure there wouldn't be any unexpected performance issues due to allocation weirdness when calling `getBytes()`. I've seen instances where bytebuffers get sized weirdly when converting from strings sometimes since Java strings are essentially UTF-16 and not ASCII/UTF-8(which is what we need here), I figured this version was more obviously correct since it wouldn't be prone to those side effects as it entirely avoids Java UTF-16 formatted strings.

scottf: Gotcha!

jameshilliard: Yeah, I've tried to fairly aggressively minimize the usage of strings wherever feasible, especially in the lower level serialization/de-serialization code as string conversions are relatively expensive operations that need to be aggressively minimized for performance. Strings are also just difficult to work with in general with lower level Java networking code since they make accurate buffer sizing difficult, hence why I'm trying to have everything immediately normalized to `ByteBuffer`'s internally as soon as feasible.

src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -34,36 +33,25 @@
     NatsMessage next; // for linked list
 
     // Create a message to publish
-    NatsMessage(String subject, String replyTo, ByteBuffer data, boolean utf8mode) {
+    NatsMessage(ByteBuffer subject, ByteBuffer replyTo, ByteBuffer data, boolean utf8mode) {
```
scottf, line 36: Since everything is already in buffers by here, can utf8mode flag be completely removed from NatsMessage?

jameshilliard: Probably, I left it there since I wasn't sure if this was a public API or not though.

src/main/java/io/nats/client/NUID.java
```diff
@@ -98,15 +99,15 @@ public NUID() {
      * @return the next NUID string from a shared global NUID instance
      */
     public static synchronized String nextGlobal() {
-        return getInstance().next();
+        return StandardCharsets.US_ASCII.decode(getInstance().next()).toString();
```
ColinSullivan1, line 102: Why the decode?

src/main/java/io/nats/client/NUID.java
```diff
@@ -98,15 +99,15 @@ public NUID() {
...
     }
 
     /**
      * Generate the next NUID string from this instance.
      *
      * @return the next NUID string from this instance.
      */
-    public final synchronized String next() {
+    public final synchronized ByteBuffer next() {
```
ColinSullivan1, line 110: This could break existing users... perhaps deprecate, and add `nextByteBuffer()`?

src/main/java/io/nats/client/Subscription.java
```diff
@@ -41,12 +42,12 @@
     /**
      * @return the subject associated with this subscription, will be non-null
      */
-    public String getSubject();
+    public ByteBuffer getSubject();
```
ColinSullivan1, line 45: This might break existing applications?  ...String -> ByteBuffer

src/main/java/io/nats/client/NUID.java
jameshilliard: That was since `getInstance().next()` was returning a `ByteBuffer`, I've refactored it to return a `String` now.

src/main/java/io/nats/client/NUID.java
jameshilliard: Changed back to return a `String`.

src/main/java/io/nats/client/Subscription.java
jameshilliard: Changed to return a `String`.

scottf:
So while I was a waiting for this to be resolved since my work directly depends on it, I took the time to build a rudimentary benchmark. My concern was the buffer change adds complexity that outweighs the benefits. I believe the test support that opinion. 

The test was new-ing 3 NatsMessage which just basically does a little string or buffer manipulation. I don't do anything with the message except create it. The  3 creates were 1) a protocol only message, 2) a publish message without utf8 mode and 3) a publish method with utf8mode. I do that because the String version has slightly different pathway with the publish modes. The buffer version does an encoding both modes. The string version uses builders as I had already written that code.

I ran those tests 2 billion times each. On my laptop. I made sure nothing else was running. I tried to write the code to time only create code. It's not perfect, but I think it supports my point. In this case, buffers actually seem slower. The raw numbers were :
```
String version:  801327 millis = 13.35545 minutes
Buffers version: 983311 millis = 16.38857 minutes
```

The buffers version of the benchmark:
```
    int stop = 2_000_000_000;
    int print = 500_000;
    long total = 0;
    for (int x = 0; x < stop; x++) {
        if ((x % print) == 0) System.out.println(x);
        String data = "" + x;
        byte[] body = (data).getBytes();
        String subject = "subj" + x;
        String replyTo = "reply" + x;

        long start = System.currentTimeMillis();
        new NatsMessage(CharBuffer.wrap(data));
        new NatsMessage(subject, replyTo, body, true);
        new NatsMessage(subject, replyTo, body, false);
        total += (System.currentTimeMillis() - start);
    }
    System.out.println("\n" + total);
}
```

The string/builders version:
```
    int stop = 2_000_000_000;
    int print = 500_000;
    long total = 0;
    for (int x = 0; x < stop; x++) {
        if ((x % print) == 0) System.out.println(x);
        String data = "" + x;
        byte[] body = (data).getBytes();
        String subject = "subj" + x;
        String replyTo = "reply" + x;

        long start = System.currentTimeMillis();
        new NatsMessage.Builder().protocol(data).build();
        new NatsMessage.Builder().subject(subject).replyTo(replyTo).data(body).utf8mode(true).build();
        new NatsMessage.Builder().subject(subject).replyTo(replyTo).data(body).build();
        total += (System.currentTimeMillis() - start);
    }
    System.out.println("\n" + total);
}
```

jameshilliard:
>  I tried to write the code to time only create code. It's not perfect, but I think it supports my point. In this case, buffers actually seem slower.

That's expected because I moved a lot of serialization/format conversions out of `NatsConnectionReader` to `NatsMessage`, so it's not at all surprising your benchmark for `NatsMessage` is slower than before, the `NatsConnectionReader` and many other parts of the code should be significantly faster however since `NatsMessage` is essentially where the string encoding/decoding now happens.

scottf:
Yes, agreed. The test isn't far from perfect and does not take into account all your changes. So maybe we should build a better benchmark. But it still seems to support not adding complexity until we know it's really worth it.

jameshilliard:
> But it still seems to support not adding complexity until we know it's really worth it.

This while somewhat more verbose does make it significantly easier to size buffers properly, this is rather important for optimizing a lot of the low level message handling code, it also allows us to eliminate redundant codepaths in `NatsConnectionReader` as we can now effectively operate on raw UTF8/ASCII bytes instead of UTF-16 strings(which are really annoying to deal with when serializing to UTF-8/ASCII since there's not an efficient way to determine the appropriate buffer sizes from UTF-16 strings).

src/main/java/io/nats/client/impl/NatsConnectionReader.java
```diff
@@ -104,24 +93,18 @@ public void run() {
...
-                    while (this.bufferPosition < bytesRead) {
+                    while (this.buffer.hasRemaining()) {
                         if (this.mode == Mode.GATHER_OP) {
                             this.gatherOp(bytesRead);
                         } else if (this.mode == Mode.GATHER_MSG_PROTO) {
-                            if (this.utf8Mode) {
-                                this.gatherProtocol(bytesRead);
-                            } else {
-                                this.gatherMessageProtocol(bytesRead);
-                            }
```
jameshilliard, line 124: This is eliminated.

src/main/java/io/nats/client/impl/NatsConnectionReader.java
```diff
@@ -300,165 +280,142 @@ void gatherMessageData(int maxPos) throws IOException {
...
-                if (this.utf8Mode) {
-                    protocolLineLength = protocolBuffer.remaining() + 4;
+                        if(subject==null || subject.limit() == 0 || sid==null || sid.limit() == 0 || lengthChars==null) {
+                            throw new IllegalStateException("Bad MSG control line, missing required fields");
+                        }
 
-                    CharBuffer buff = StandardCharsets.UTF_8.decode(protocolBuffer);
-                    protocolLength = buff.remaining();
-                    buff.get(this.msgLineChars, 0, protocolLength);
-                }
```
jameshilliard, line 402: This as well.

src/main/java/io/nats/client/impl/NatsConnectionReader.java
```diff
@@ -300,165 +280,142 @@ void gatherMessageData(int maxPos) throws IOException {
...
-                        (chars[1] == 'N' || chars[1] == 'n') && 
-                        (chars[2] == 'F' || chars[2] == 'f') &&
-                        (chars[3] == 'O' || chars[3] == 'o')) {
-                return NatsConnection.OP_INFO;
-            }  else {
-                return UNKNOWN_OP;
-            }
-        } else {
-            return UNKNOWN_OP;
-        }
```
jameshilliard, line 362: Also was able to eliminate this scary function.

scottf:
So let's pretend we fix the benchmark I wrote to consider the new code better. And lets say instead of being 3 minutes slower, the buffer version now runs X minutes _faster_.
That's X * 60 * 1000 milliseconds faster. Remember the test did 3 messages each of 2 billion rounds. So that's 6 billion messages. If we look at what happens if the new test runs 10/100/1000 minutes faster, we can see how much gain  per message we are getting.  For our use case, how is even 1_000 nanos improvement per message worth the complexity?

```
Minutes = Millis     / Total         = Gain Per  | 
Faster  = Faster     / Messages      = Message   | Fraction    | Nanos
------- = ---------- / ------------- = --------- | ----------- | ------------
   10   =    600_000 / 6_000_000_000 = 0.0001 ms | 1/10_000 ms |    100 nanos
  100   =  6_000_000 / 6_000_000_000 = 0.001 ms  | 1/1000 ms   |  1_000 nanos
 1000   = 60_000_000 / 6_000_000_000 = 0.01 ms   | 1/100 ms    | 10_000 nanos
```

jameshilliard:
> For our use case, is even 1_000 nanos improvement per message worth the complexity?

I don't really consider this to be an overall complexity increase, but more of a complexity shift that moves a large amount of the string message serialization/de-serialization complexity out of lower level layers and into higher layers, the idea is that this significantly decreases complexity around certain areas in order to make further refactoring and optimizations such as using `AsynchronousSocketChannel`'s more straight forward.

Fundamentally at their lowest level network protocols are byte oriented, not string oriented, I can see why this change may look more complex as a lot of the `ByteBuffer` API's are very low level which result in more verbosity(this is a problem with Java in general and why a lot of the higher level code we write is actually in Kotlin), however the message processing logic itself IMO ends up being significantly simpler and easier to reason about when you can avoid Java's UTF-16 strings entirely since we can split message segments and do buffer allocations directly using the raw socket bytes/ByteBuffer's without any string conversions since the parts we need to interpret are all single byte ASCII characters.

I should also add that one of the goals for this change is to make it possible to add additional public API's that allow for the nats library to be used without any string encoding/decoding operations whatsoever for certain hot-paths, in order to make that possible without a large amount of code duplication we need to move all string encoding/decoding out of the low level connection reader/writer code.

In terms of use case a lot of what I intend to use nats for is for load balanced services that act as protocol converters/proxies, which essentially ingest a nats message, do some conversion operations/filtering and forward the request on to other services that may not speak nats protocol directly. For example this would be used for gateway services that essentially would act as high performance legacy protocol conversion gateway proxies that essentially wrap legacy backend services which in many cases may not be able to be made to easily speak nats protocol directly. In order to efficiently interface with these legacy network protocols in a low latency manner we want to avoid any unnecessary string encoding operations and pass bytes/buffers directly from one network socket to the other with both protocols using non-blocking IO without doing encoding/decoding of any part of the nats message to/from strings. Using a buffer passing approach here should make it significantly easier to heavily optimize for these performance critical codepaths as Java's async non-blocking IO support is largely built around `AsynchronousSocketChannel` which natively operates on `ByteBuffer`'s.

> So let's pretend we fix the benchmark I wrote to consider the new code better.

This set of changes by itself is not really intended to significantly improve performance by itself but rather act as groundwork for further refactoring that will do so, it's probably a bit early to be micro-benching performance as further refactoring will be needed to see a significant benefit. That can be done more incrementally I think while this set of changes was much more difficult to split up as intermixing `ByteBuffer` based message processing with string based message processing is quite difficult, so I pretty much had to convert a large portion of the internals all at once.

jameshilliard:
I now have SSL working in #357 which is based off of this.

ColinSullivan1:
I wanted to give an update for this PR.  There are many PRs that will need to be merged about the same time, so would like to take a methodical approach to this.

#362 and #354 contain missing functionality that is delaying POCs of JetStream so these should be merged first.  At that point I'd like to cut a beta release to unblock these user and then look at byte buffers (this PR and #364).  

Optimization is very important to us but it's more important to to unblock our users waiting use JetStream, with a minimum of risk.  This PR will get reviewed and merged - we just need to get the headers and the Jetstream functionality in first.  I apologize in advance for the rebase that will need to happen, but from a product perspective we need to prioritize this way.

Thanks again for all of this work and please be patient as we move forward with merging these PRs.

jameshilliard:
> we just need to get the headers and the Jetstream functionality in first

Yeah, I'm going over and refactoring the headers PR a bit to make the re-base this easier at the moment.

ColinSullivan1:
FYI - #354 will likely change quite a bit.

jameshilliard:
> At that point I'd like to cut a beta release to unblock these user and then look at byte buffers (this PR and #364).

Well #364 should apply cleanly on top of #362 and also fixes some tests/bugs in #362, I would recommend just importing my patches in #364 to the #362 PR branch.

scottf:
I have done quite a bit of work on the client, used bytebuffers in some places, wrapped with ByteArrayBuilders. I have also done some profiling. We have some places in the client where we need to improve performance, but string manipulation is extremely low on the places that profile poorly. Closing this PR

PR #358, opened 2020-11-07
Use OS Default SecureRandom.

jameshilliard:
The bug causing `/dev/random` to be used on Linux is due to the `SHA1PRNG`(use of which is generally discouraged) instance of `SecureRandom` being used, use the OS default instance to fix this issue which should properly select `/dev/urandom` on Linux and have sane defaults on other systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,12 +109,6 @@ If you need the absolute latest, before it propagates to maven central, you can 
 
 If you are using the 1.x version of java-nats and don't want to upgrade to 2.0.0 please use ranges in your POM file, java-nats-streaming 1.x is using [1.1, 1.9.9) for this.
 
-### Linux Platform Note
-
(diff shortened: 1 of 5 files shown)


scottf:
@jameshilliard So I was looking at this today and we want to include it. Is it possible for you to redo it against the most recent master? At a minimum, we want to use the default SecureRandom like you suggest.

Also, could you verify that this is still an issue with JDK's 8 or later. If it still is an issue, could you update the comment with references as those bugs refer to much older JDKs

If it's no longer an issue, we won't need the comment in the readme and the build.gradle doesn't need the jvm arg, but we still want to change to using the default SecureRandom.

Also, you will notice that I changed NUID to only create a SecureRandom once and re-use it. This change made creation of a new NUID instance take only 0.02 milliseconds on average versus it used to take 1.6 millis on average (on my machine ; )

I was thinking to pull out the randoms as static to a support [package] class so it can be re-used everywhere. We don't so much need security as we need the larger set of possible numbers that the secure random gives us over regular random.

jameshilliard:
> So I was looking at this today and we want to include it. Is it possible for you to redo it against the most recent master? At a minimum, we want to use the default SecureRandom like you suggest.

Rebased

> Also, could you verify that this is still an issue with JDK's 8 or later. If it still is an issue when using the default SecureRandom, could you update the comment with references as those bugs refer to much older JDKs

The bug is present when using the `SHA1PRNG` but not the default `SecureRandom`, with this change the bug should be fixed.

> Also, you will notice that I changed NUID to only create a SecureRandom once and re-use it. This change made creation of a new NUID instance take only 0.02 milliseconds on average versus it used to take 1.6 millis on average (on my machine ; )

Great.

scottf (approved):
LGTM

PR #359, opened 2020-11-08
Headers object implementation and NatsMessage.Builder

scottf:
also, Unit tests, UTF-8 test data improvement


src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -13,57 +13,89 @@
...
+
+    // Kind.REGULAR : just these fields
     private String subject;
     private String replyTo;
     private ByteBuffer data;
+    private boolean utf8mode;
+    private Headers headers;
+
+    // Kind.INCOMING : subject, replyTo, data and these fields
+    private String sid;
```
jameshilliard, line 38: I think it makes sense to get rid of strings as the internal data structures in `NatsMessage` and instead only convert to strings on demand when required by external consumers so that there's no string conversion overhead unless actually required by consumers. I've done this partially in #356. In addition I think we may want to move more parsing functionality from `NatsConnectionReader` into `NatsMessage` in general, in #356 I modified `NatsMessage` so that it is used to construct the complete outbound message line buffer required to write the message to the socket so that all message segment serialization functionality can be removed from `NatsConnectionWriter` I also have it so that the primary `NatsMessage` message buffer is also used as the backing buffer for the more specific buffers for the various message segments, this allows for directly calling equals operations on the segment buffers. Something along these lines may make sense for handling inbound messages we well, essentially where we use a single primary buffer with other referencing buffers, although they could also be kept separate for inbound as inbound messages do not ever need to be serialized back to full message lines like outbound messages need to be.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -0,0 +1,217 @@
...
+				throw new IllegalArgumentException(KEY_INVALID_CHARACTER + c);
+			}
+		});
+	}
+
+	private boolean checkValue(String val) {
+		// Generally more permissive than HTTP.  Allow only printable
+		// characters and include tab (0x9) to cover what's allowed
+		// in quoted strings and comments.
+		// null is just ignored
```
jameshilliard, line 191: Since this operates on single ASCII bytes we should be able to avoid strings/chars here I think.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -0,0 +1,217 @@
...
+
+import java.util.*;
+
+public class Headers {
+	private static final String KEY_CANNOT_BE_EMPTY_OR_NULL = "Header key cannot be null.";
+	private static final String KEY_INVALID_CHARACTER = "Header key has invalid character: ";
+	private static final String VALUE_INVALID_CHARACTERS = "Header value has invalid character: ";
+
+	private final Map<String, List<String>> headerMap = new HashMap<>();
+	private final Map<String, String> valueCsvCache = new HashMap<>();
```
jameshilliard, line 24: It should be possible to use a `ByteBuffer` based hashmap here like I've done in #356 which makes message serialization simpler in general as Java strings(which are UTF-16 based) do not properly map to the wire protocol(which is ASCII/UTF-8 based).

PR #362, opened 2020-11-14
Headers (HPUB) support

scottf:
File by file notes:
<pre>
main/java/io/nats/client/
    Connection.java - interface change to support publish and request with headers
    Message.java    - interface change to headers
    Options.java    - changes to suport headers in the connect string
	
main/java/io/nats/client/impl/
    ByteArrayBuilder.java     - (BAB) utility class to simplying using ByteBuffers
    Headers.java              - object to manage the header map 
    NatsConnection.java       - allow publish and request with headers. Use BAB to build simple protocol messages. 
    NatsConnectionReader.java - read HMSG
    NatsConnectionWriter.java - write HPUB
    NatsMessage.java          - Headers support. constructor cleanup. builder for normal messages (ctor is fine for incoming and protocol)

examples/java/io/nats/examples/
    ExampleArgs.java - utility to read command line args to support header inputs
    NatsPub.java     - support headers, refactor to use example args, additional output
    NatsReply.java   - "
    NatsReq.java     - "
    NatsSub.java     - "

test/java/io/nats/client/
    NKeyTests.java       - extract utf-8 test strings to data file 
    OptionsTests.java    - connect protocol has "headers": true
    PublishTests.java    - publish api takes headers
    SubscriberTests.java - extract utf-8 test strings to data file 
 
test/java/io/nats/client/impl/
    ByteArrayBuilderTests.java - test BAB utility that it auto expands and the data is correct
    HeadersTests.java          - test the headers object
    MessageProtocolCreationBenchmark.java - NatsMessage no longer contstructs with CharBuffer, takes String or byte[] 
    MessageQueueBenchmark.java            - " 
    MessageQueueTests.java                - "
    NatsMessageTests.java                 - "

test/java/io/nats/client/utils/
    ResourceUtils.java - utility with method to read a text file from the resource folder

test/resources/
    utf8-test-strings.txt - test data
</pre>


src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -0,0 +1,346 @@
...
+	private static String CRLF = "\r\n";
+	private static String HEADER_PREFIX = "NATS/1.0\r\n";
+
+	private static final String KEY_CANNOT_BE_EMPTY_OR_NULL = "Header key cannot be null.";
+	private static final String KEY_INVALID_CHARACTER = "Header key has invalid character: ";
+	private static final String VALUE_INVALID_CHARACTERS = "Header value has invalid character: ";
+
+	private static final boolean normalizeKey = false;
+
+	private Map<String, List<String>> headerMap = new HashMap<>();
```
jameshilliard, line 33: This can probably be changed to:
```
private Map<ByteBuffer List<ByteBuffer>> headerMap = new HashMap<>();
```
From what I can tell the headers are ASCII which you can operate on without any string conversions generally.

ColinSullivan1: Let's stick with String for now to be consistent with other languages (e.g. .NET uses string), and then take a pass at optimization afterword.  We might end up supporting UTF-8 in headers, TBD...

jameshilliard: I'm mostly just referring to the internal data structures since Java doesn't have support for native UTF-8 or ASCII strings, public API's can still accept/return strings and convert to/from an internal ByteBuffer format.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -891,18 +884,15 @@ void unsubscribe(NatsSubscription sub, int after) {
     }
 
     void sendUnsub(NatsSubscription sub, int after) {
-        String sid = sub.getSID();
-        CharBuffer protocolBuilder = CharBuffer.allocate(this.options.getMaxControlLine());
-        protocolBuilder.append(OP_UNSUB);
-        protocolBuilder.append(" ");
-        protocolBuilder.append(sid);
+        // allocate the proto length + 19 + 10 (sid is a long, 19 bytes max, after is an int 10 bytes max)
+        ByteArrayBuilder bab = new ByteArrayBuilder(OP_UNSUB_PROTOCOL_LEN + 29)
```
ColinSullivan1, line 888: Nice...

src/main/java/io/nats/client/impl/NatsConnectionReader.java
```diff
@@ -251,6 +260,40 @@ void gatherProtocol(int maxPos) throws IOException {
...
+                }
+
+                if (msgHeadersPosition == msgHeaders.length) {
+                    incoming.setHeaders(msgHeaders);
+                    msgHeaders = null;
+                    msgHeadersPosition = -1;
+                    this.mode = Mode.GATHER_DATA;
+                    break;
+                } else {
+                    throw new IllegalStateException("Bad socket data, headers do not match excpected length");
```
ColinSullivan1, line 289: typo

src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -13,195 +13,159 @@
...
 import io.nats.client.Connection;
 import io.nats.client.Message;
 import io.nats.client.Subscription;
 
+import static java.nio.charset.StandardCharsets.US_ASCII;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 class NatsMessage implements Message {
-    private String sid;
+    private static byte[] PUB_BYTES = "PUB ".getBytes(US_ASCII);
```
ColinSullivan1, line 24: imo the protocol bytes and lengths should be defined in NatsConnection with the others.

src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -220,19 +185,94 @@ public Connection getConnection() {
...
+    // The builder is only supports building normal publish/request messages,
+    // as an option for client use developers instead of the normal constructor
+    public static class Builder {
+        String subject;
+        String replyTo;
+        Headers headers;
+        byte[] data;
+        boolean utf8mode;
+
+        public Builder subject(final String subject) {
```
ColinSullivan1, line 249: We'll want comments on these - I think the gradle build will fail on the javadoc task w/o them.

ColinSullivan1 (changes requested):
Overall really good!  A few nitpicks, and as discussed on our call, publishing a message with headers vs passing headers into the publish API would be more inline with the other clients.

src/main/java/io/nats/client/impl/Headers.java
scottf: A couple things here. If we use ByteBuffer as a key, I would have to write a hashcode and equals, because they don't exist (which means they default to the Object's). As you mentioned, developers are going to want to work in Strings and if we provide a string api with ByteBuffers underneath, we would have to convert every time we put it into the map.
@jameshilliard Can you look at the 2 things, first the Headers class, specifically the method getSerialized. So when the publish code asks for the headers, it asks for them serialized, which I give directly in a byte[], which is generated using ByteBuffers under the covers in ByteArrayBuilder. The second thing, could you look at ByteArrayBuilder, I'd like some comment on how I'm calculating expanding the buffer.

src/main/java/io/nats/client/impl/NatsConnectionReader.java
scottf: not mine ;) but fixed.

src/main/java/io/nats/client/impl/NatsMessage.java
scottf: I think they should be pulled out into a common constants file. I do not like getting them from NatsConnection because it makes a circular dependency, which I know the compiler handles fine. I'll fix stuff like this on the next PR

src/main/java/io/nats/client/impl/NatsMessage.java
scottf: yes will add. Not tonight though. Is there a gradle command I should be running before commit?

src/examples/java/io/nats/examples/ExampleArgs.java
```diff
@@ -0,0 +1,106 @@
+// Copyright 2018 The NATS Authors
```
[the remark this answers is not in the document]
scottf: fixed

src/main/java/io/nats/client/impl/ByteArrayBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+package io.nats.client.impl;
+
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+
+import static java.nio.charset.StandardCharsets.US_ASCII;
+
+public class ByteArrayBuilder {
```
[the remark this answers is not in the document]
scottf: not sure. do we specifically want to hide it? It could be a useful utility for a dev

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -0,0 +1,274 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package io.nats.client.impl;
+
+import java.util.*;
+
+import static java.nio.charset.StandardCharsets.US_ASCII;
+
+public class Headers {
```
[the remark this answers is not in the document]
scottf: do you mean an iterator? Either way, I'll add it next round. I will add comments regarding thread safety.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -0,0 +1,274 @@
...
+	private byte[] serialized;
+
+	public Headers() {
+		headerMap = new HashMap<>();
+	}
+
+	public Headers(byte[] serialized) {
+		this();
+		if (serialized != null && serialized.length > 0) {
+			String[] split = new String(serialized, US_ASCII).split("\\Q\r\n\\E");
```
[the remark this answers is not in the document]
scottf: Ok, let's discuss so I know what you mean.

src/main/java/io/nats/client/Connection.java
```diff
@@ -105,33 +107,54 @@
...
+     * where the sender creates a byte array immediately before calling publish.
+     *
+     * See {@link #publish(String, String, byte[]) publish()} for more details on
+     * publish during reconnect.
+     *
+     * @param subject the subject to send the message to
+     * @param body the message body
+     * @throws IllegalStateException if the reconnect buffer is exceeded
+     */
+    public void publish(String subject, Headers headers, byte[] body);
```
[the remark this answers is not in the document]
scottf: yes, changed

src/main/java/io/nats/client/impl/ByteArrayBuilder.java
```diff
@@ -0,0 +1,120 @@
+// Copyright 2015-2020 The NATS Authors
```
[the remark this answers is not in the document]
scottf: fixed

src/main/java/io/nats/client/Message.java
```diff
@@ -33,6 +35,16 @@
 	 */
 	public String getReplyTo();
 
+	/**
+	 * @return the full serizlialized headers as byte array from the message
+	 */
+	public byte[] getSerializedHeader();
```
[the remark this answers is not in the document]
scottf: fixed

src/main/java/io/nats/client/impl/Headers.java
jameshilliard: > If we use ByteBuffer as a key, I would have to write a hashcode and equals, because they don't exist (which means they default to the Object's). 

They both exist, see [equals](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#equals-java.lang.Object-) and [hashCode](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#hashCode--). Note that #356 makes use of these features for the hashmaps and buffer matching.

> As you mentioned, developers are going to want to work in Strings and if we provide a string api with ByteBuffers underneath, we would have to convert every time we put it into the map.

Sure, but they eventually will have to be serialized to a bytes like type at least once at a minimum, IMO better to do that up front and use a normalized `ByteBuffer` format internally.

src/main/java/io/nats/client/impl/ByteArrayBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+        return this;
+    }
+
+    public ByteArrayBuilder append(CharBuffer cb, Charset charset) {
+        append(cb.toString().getBytes(charset));
+        return this;
+    }
+
+    public ByteArrayBuilder append(byte[] bytes) {
+        ensureCapacity(bytes.length);
```
jameshilliard, line 104: We should avoid having to do this as allocations are relatively expensive.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -0,0 +1,274 @@
...
+	public byte[] getSerialized() {
+		if (serialized == null) {
+			ByteArrayBuilder bab = new ByteArrayBuilder()
+					.append(VERSION_BYTES, VERSION_BYTES_LEN);
+			for (String key : headerMap.keySet()) {
+				for (String value : values(key)) {
+					bab.append(key);
+					bab.append(COLON_BYTES, COLON_BYTES_LEN);
+					bab.append(value);
+					bab.append(CRLF_BYTES, CRLF_BYTES_LEN);
```
jameshilliard, line 208: These append operations do not look efficient as they appear to introduce a lot of allocation overhead, you really should be pre-allocating the exact amount of bytes necessary rather than expanding the buffer repeatedly. This should be fairly straight forward if you replace the strings in `headerMap` with `ByteBuffer`'s.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -0,0 +1,274 @@
...
+	 *         -or- if any value contains invalid characters
+	 */
+	public Headers add(String key, Collection<String> values) {
+		_add(key, values);
+		return this;
+	}
+
+	// the add delegate
+	private void _add(String key, Collection<String> values) {
+		if (values != null) {
```
jameshilliard, line 87: You should have this function convert all `String`'s to `ByteBuffer`'s before inserting into `headerMap`. Accessors that need to return strings should do the reverse.

src/main/java/io/nats/client/impl/Headers.java
jameshilliard: > The second thing, could you look at ByteArrayBuilder, I'd like some comment on how I'm calculating expanding the buffer.

This class to me seems like a hack to add inefficient string conversion/resize operations to `ByteBuffer`'s, these operations doesn't seem necessary as long as you ensure each of the individual header `Strings` are already converted to `ByteBuffer`'s by the time they need to be serialized as you can then do a single up front allocation by checking `ByteBuffer` sizes using `remaining()`. Header serialization should probably be done in `NatsMessage`, maybe by passing a headers object to it, this way you don't have to do any separate buffer allocations for the serialized headers. See how I do the full message serialization with a single allocation [here](https://github.com/nats-io/nats.java/pull/356/files#diff-519f371d84d8e0274587b137b1f2fcfcaa9ab8afb0897e54bcc4418e0b2f7244R52) and then just create shared references to it for the field specific slice buffers

jameshilliard: Basically the way I've been refactoring this is by first ensuring that everything is already converted to a `ByteBuffer` by the time you need to calculate a buffer size, as pre-calculating a buffer size based on a string is not efficient and difficult as Java does not support ASCII/UTF-8 encoded strings natively.

jameshilliard: Here's what I mean: https://github.com/nats-io/nats.java/pull/364/commits/f833b99d51166821afcd544d29a0f32a84d94ace

jameshilliard: Notice that I'm using the "`serialized`" `ByteBuffer` to back the `headerMap` as well to reduce allocations.

jameshilliard: Notice that I only do a single buffer allocation, the one in `getSerialized` which allocates the exact buffer size needed for header serialization up front, even this single allocation can probably be eliminated by writing directly to a `NatsMessage` buffer.

src/main/java/io/nats/client/impl/NatsMessage.java
ColinSullivan1: I do `./gradlew build`

src/main/java/io/nats/client/impl/ByteArrayBuilder.java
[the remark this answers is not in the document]
ColinSullivan1: Either way...

src/main/java/io/nats/client/impl/Headers.java
ColinSullivan1: Could track size when adding/putting/removing headers and allocate the exact amount here. tbh, I'm not sure how much difference that'd make.  For maximum performance in an application you'd attempt to keep headers static and re-use messages (thus only one initial allocation and serialization).

jameshilliard: Take a look how I did it [here](https://github.com/nats-io/nats.java/commit/f833b99d51166821afcd544d29a0f32a84d94ace).

src/main/java/io/nats/client/impl/Headers.java
scottf: Not sure how I missed the equals, but you are right it's there. Can you benchmark the two implementations?

Also regarding the ByteArrayBuffer, we'll have to agree to disagree. There are plenty of string conversions for the keys and values to use them map, so I'm not seeing a huge benefit there. I also think in practical use, there will be minimal resizing since the original allocation can be set and the default probably covers a ton of cases, for instance 32 or 64 bytes for a protocol message, and then maybe one resize to hold the entire data block. I'm sure there are optimization as well when we factor in usage, but that's outside of the scope of getting headers working.

The headers version with byte buffers is 100 lines longer (379 from 275), mostly for the serialization. The NatsMessage code where I use ByteArrayBuilder is much more compact / readable and saves a lot of thinking ahead to figure out the lengths ahead of time.

I feel like this code is more readable, more maintainable and fast enough - for a client. In fact my addition of using byte buffers was already to respect your position on using them. I suppose for the server, when every single nano counts I would more wholeheartedly support this level of effort (which is why it's written in go,) but again for the client?

src/main/java/io/nats/client/impl/Headers.java
scottf: I considered some tracking of the length of data added and removed, wasn't sure if it was worth the effort. We can add some optimization next round.

src/main/java/io/nats/client/impl/NatsMessage.java
scottf: the javadoc excludes `excludes = ['io/nats/client/impl', 'io/nats/examples']` but I'll still add comments

src/main/java/io/nats/client/impl/Headers.java
jameshilliard: > Can you benchmark the two implementations?

When I did the initial refactoring of `NatsMessage` to use `ByteBuffer`'s avoiding unnecessary buffer allocations had a significant performance benefit.

> There are plenty of string conversions for the keys and values to use them map, so I'm not seeing a huge benefit there.

The benefit is that you should not be expanding the `ByteBuffer` as that requires expensive reallocations, string conversions should all be done prior to buffer allocation, rather than repeatedly reallocating and expanding the buffer.

> The NatsMessage code where I use ByteArrayBuilder is much more compact / readable and saves a lot of thinking ahead to figure out the lengths ahead of time.

I noticed a number of rather subtle buffer sizing issues with how you were serializing buffers, I'm trying to sidestep these issues entirely.

> I feel like this code is more readable

It's less verbose maybe but has subtle edge cases and performance issues.

> I suppose for the server, when every single nano counts I would more wholeheartedly support this level of effort (which is why it's written in go,) but again for the client?

I'm planning to use this for a protocol converting proxy gateway server in Java.

I went ahead and refactored out the usage of `ByteArrayBuilder` here already and fixed the failing tests: https://github.com/nats-io/nats.java/pull/364/commits/92e4c6c9515c3e6ee728fd40b2605aff093ccb61

jameshilliard: Here's some benchmarks of the `NatsMessage` protocol serialization codepath before and after removing `ByteArrayBuilder` usage:
Before:
```
### Total time to create 50,000,000 non-utf8 messages for sending was 9,724 ms
	194.491932 ns/op
	5,141,601.44 op/sec

### Total time to create 50,000,000 utf8 messages for sending was 10,006 ms
	200.134972 ns/op
	4,996,627.964 op/sec
```
After:
```
### Total time to create 50,000,000 non-utf8 messages for sending was 8,437 ms
	168.749313 ns/op
	5,925,950.062 op/sec

### Total time to create 50,000,000 utf8 messages for sending was 8,598 ms
	171.961319 ns/op
	5,815,261.283 op/sec
```

src/main/java/io/nats/client/impl/Headers.java
jameshilliard: You can just do it like [this](https://github.com/nats-io/nats.java/commit/92e4c6c9515c3e6ee728fd40b2605aff093ccb61#diff-5224933708c6d2f3fc83858e058e54cb3aff6e55ae4e679444ea05936c9e19a6R295-R304) which doesn't require serialization/de-serialization to compute the length.

src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -13,195 +13,184 @@
...
-        protocolBytes.put(subject.getBytes(charset));
-        protocolBytes.put((byte)' ');
+        totLen = hdrLen + dataLen;
 
-        if (replyTo != null) {
-            protocolBytes.put(replyTo.getBytes(charset));
-            protocolBytes.put((byte)' ');
+        // initialize the builder with a reasonable length, preventing resize in 99.9% of the cases
+        // 32 for misc + subject length doubled in case of utf8 mode + replyToLen + totLen (hdrLen + dataLen)
+        ByteArrayBuilder bab = new ByteArrayBuilder(32 + (subject.length() * 2) + replyToLen + totLen);
```
jameshilliard, line 94: This has the potential to cause subtle breakage if you don't precisely allocate the correct buffer size.

jameshilliard:
Rebased #364 on top of latest changes here.

src/main/java/io/nats/client/impl/NatsConnection.java
jameshilliard: This doesn't allocate the correct buffer size which causes unnecessary allocations/copying.

src/main/java/io/nats/client/impl/ByteArrayBuilder.java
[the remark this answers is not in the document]
jameshilliard: I think we can just get rid of this class, it seems to be unnecessary and inefficient. If you do the string->bytebuffer conversions prior to serialization(maybe in something like the `NatsMessage` `Builder`) you then don't have to resize the serialized buffers at all as computing the buffer sizes is straightforward for bytebuffers unlike strings, but sizing the buffers properly is somewhat function specific so this might not be able to be fully generalized.

src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -13,195 +13,184 @@
...
+public class NatsMessage implements Message {
+
+    public enum Kind {REGULAR, PROTOCOL, INCOMING}
+
+    // Kind.REGULAR : just these fields
     private String subject;
     private String replyTo;
-    private ByteBuffer data;
-    private ByteBuffer protocolBytes;
+    private byte[] data;
```
jameshilliard, line 33: Why were these changed?

src/main/java/io/nats/client/impl/Headers.java
jameshilliard: > In fact my addition of using byte buffers was already to respect your position on using them.

You have to be careful with how you use `ByteBuffer`'s, I'm primarily using them to reduce buffer copy/allocation operations and improve compatibility with the buffer passing asynchronous socket API's by natively operating on `ByteBuffers` within serialization/de-serialization classes. The approach you've used in `ByteArrayBuilder` to encapsulate `ByteBuffer` operations goes counter to those goals as you're just converting them in the end to regular `byte[]` arrays and are also greatly increasing copying/allocations via resizing.

jameshilliard: Basically you want to look at each `ByteBuffer.allocate()` as a c style char buffer `malloc()`. The `ByteBuffer` class is just a thin abstraction over a c style char buffer and some pointers from my understanding.

scottf: I'm not under allocating / resizing too often, I've already addressed that. So my takeaway from this is to reduce buffer copy. For instance when serializing the headers and other uses of ByteArrayBuilder, return or allow access to the ByteBuffer to allow the copy to come from the ByteBuffer's byte[] instead of the copy to come from the byte[] created from the ByteBuffer.
In my opinion, it's not just less verbose, it's significantly simpler. As in all software there are trade offs, and, again, my opinion is that this trade off is well worth it for less than 2 second loss over 50 million passes for the message. For our regular use case, at least for today when the priority is getting functional headers, this is more than acceptable, I cannot speak to your specific use case.

src/main/java/io/nats/client/impl/NatsConnection.java
scottf: `OP_UNSUB_PROTOCOL_LEN + 29` is the maximum number of bytes needed as noted in the comment just above the line of code. The builder code actually allocates in chunks of 32 bytes, so 64 bytes is probably allocated here.

jameshilliard: Allocating too large a buffer just like too small causes performance issues as both result in extra buffer copying.

src/main/java/io/nats/client/impl/Headers.java
jameshilliard: > I'm not under allocating / resizing too often, I've already addressed that.

You shouldn't be resizing those buffers at all if you allocate the correct amount up front, which is straight forward to do.

> In my opinion, it's not just less verbose, it's significantly simpler.

Not disagreeing with that, but that's not the issue, the issue is that it creates a significant performance regression by not sizing correctly up front.

> As in all software there are trade offs, and, again, my opinion is that this trade off is well worth it for less than 2 second loss over 50 million passes for the message.

Well it all adds up, the more places `ByteArrayBuilder` gets used instead of up front correctly sized buffers the worse performance will get.

> For our regular use case, at least for today when the priority is getting functional headers, this is more than acceptable, I cannot speak to your specific use case.

I have headers with up front correctly sized buffers working in #364(which should apply cleanly on top of this PR) so not sure why it would make sense to use the slower version.

jameshilliard:
I did some more bench-marking to see how much of a performance impact the extra allocations created by the use of `ByteArrayBuilder` have overall, they seem to have a significant overall negative performance impact(in some cases around 50% slower) for some workloads while others have nearly no impact.

With #364:
> Task :NatsAutoBench.main()
Connecting to NATS server at nats://localhost:4222
Running warmup
Current memory usage is 990.90 mb / 1.02 gb / 16.00 gb free/total/max
Executing tests ........................................

PubOnly 0            10,000,000          4,899,764 msg/s       0.00 b/s
PubOnly 8            10,000,000          4,762,806 msg/s     36.34 mb/s
PubOnly 32           10,000,000          4,510,598 msg/s    137.65 mb/s
PubOnly 256          10,000,000          4,425,957 msg/s      1.06 gb/s
PubOnly 512          10,000,000          4,073,548 msg/s      1.94 gb/s
PubOnly 1024          1,000,000          3,019,998 msg/s      2.88 gb/s
PubOnly 4096            500,000            899,662 msg/s      3.43 gb/s
PubOnly 8192            100,000            464,286 msg/s      3.54 gb/s

PubSub 0             10,000,000          1,425,461 msg/s       0.00 b/s
PubSub 8             10,000,000          1,461,065 msg/s     11.15 mb/s
PubSub 32            10,000,000          1,284,064 msg/s     39.19 mb/s
PubSub 256           10,000,000          1,088,432 msg/s    265.73 mb/s
PubSub 512           10,000,000          1,022,268 msg/s    499.15 mb/s
PubSub 1024           1,000,000            867,574 msg/s    847.24 mb/s
PubSub 4096             500,000            368,595 msg/s      1.41 gb/s
PubSub 8192             100,000            203,657 msg/s      1.55 gb/s

PubDispatch 0        10,000,000          1,690,515 msg/s       0.00 b/s
PubDispatch 8        10,000,000          1,405,512 msg/s     10.72 mb/s
PubDispatch 32       10,000,000          1,514,133 msg/s     46.21 mb/s
PubDispatch 256      10,000,000          1,203,895 msg/s    293.92 mb/s
PubDispatch 512      10,000,000          1,170,347 msg/s    571.46 mb/s
PubDispatch 1024      1,000,000          1,061,729 msg/s      1.01 gb/s
PubDispatch 4096        500,000            368,722 msg/s      1.41 gb/s
PubDispatch 8192        100,000            157,165 msg/s      1.20 gb/s

ReqReply 0               20,000              9,140 msg/s       0.00 b/s
ReqReply 8               20,000              9,287 msg/s     72.56 kb/s
ReqReply 32              10,000              8,840 msg/s    276.25 kb/s
ReqReply 256             10,000              8,744 msg/s      2.13 mb/s
ReqReply 512             10,000              9,962 msg/s      4.86 mb/s
ReqReply 1024            10,000              9,571 msg/s      9.35 mb/s
ReqReply 4096            10,000              9,553 msg/s     37.32 mb/s
ReqReply 8192            10,000              8,316 msg/s     64.97 mb/s

Latency 0     5,000     40 /  50.76 / 171    +/- 0.75  (microseconds)
Latency 8     5,000     40 /  50.36 / 208    +/- 0.75  (microseconds)
Latency 32    5,000     38 /  49.37 / 205    +/- 0.73  (microseconds)
Latency 256   5,000     40 /  49.80 / 253    +/- 0.74  (microseconds)
Latency 512   5,000     40 /  50.50 / 261    +/- 0.75  (microseconds)
Latency 1024  5,000     39 /  50.60 / 190    +/- 0.75  (microseconds)
Latency 4096  5,000     42 /  51.67 / 196    +/- 0.76  (microseconds)
Latency 8192  5,000     44 /  53.90 / 878    +/- 0.82  (microseconds)

Final memory usage is 67.86 mb / 80.00 mb / 16.00 gb free/total/max

BUILD SUCCESSFUL in 2m 25s
2 actionable tasks: 1 executed, 1 up-to-date
9:14:28 AM: Task execution finished 'NatsAutoBench.main()'.


Without #364:

> Task :NatsAutoBench.main()
Connecting to NATS server at nats://localhost:4222
Running warmup
Current memory usage is 980.49 mb / 1.02 gb / 16.00 gb free/total/max
Executing tests ........................................

PubOnly 0            10,000,000          4,391,731 msg/s       0.00 b/s
PubOnly 8            10,000,000          4,379,475 msg/s     33.41 mb/s
PubOnly 32           10,000,000          4,620,284 msg/s    141.00 mb/s
PubOnly 256          10,000,000          3,628,144 msg/s    885.78 mb/s
PubOnly 512          10,000,000          3,305,155 msg/s      1.58 gb/s
PubOnly 1024          1,000,000          1,996,703 msg/s      1.90 gb/s
PubOnly 4096            500,000            697,794 msg/s      2.66 gb/s
PubOnly 8192            100,000            396,452 msg/s      3.02 gb/s

PubSub 0             10,000,000          1,508,454 msg/s       0.00 b/s
PubSub 8             10,000,000          1,260,994 msg/s      9.62 mb/s
PubSub 32            10,000,000          1,236,728 msg/s     37.74 mb/s
PubSub 256           10,000,000          1,058,462 msg/s    258.41 mb/s
PubSub 512           10,000,000            966,510 msg/s    471.93 mb/s
PubSub 1024           1,000,000            718,667 msg/s    701.82 mb/s
PubSub 4096             500,000            268,319 msg/s      1.02 gb/s
PubSub 8192             100,000            128,888 msg/s   1006.94 mb/s

PubDispatch 0        10,000,000          1,482,108 msg/s       0.00 b/s
PubDispatch 8        10,000,000          1,488,553 msg/s     11.36 mb/s
PubDispatch 32       10,000,000          1,400,385 msg/s     42.74 mb/s
PubDispatch 256      10,000,000          1,128,562 msg/s    275.53 mb/s
PubDispatch 512      10,000,000            994,440 msg/s    485.57 mb/s
PubDispatch 1024      1,000,000            764,356 msg/s    746.44 mb/s
PubDispatch 4096        500,000            307,758 msg/s      1.17 gb/s
PubDispatch 8192        100,000            165,218 msg/s      1.26 gb/s

ReqReply 0               20,000              9,294 msg/s       0.00 b/s
ReqReply 8               20,000              9,351 msg/s     73.06 kb/s
ReqReply 32              10,000             10,109 msg/s    315.91 kb/s
ReqReply 256             10,000              9,859 msg/s      2.41 mb/s
ReqReply 512             10,000             10,017 msg/s      4.89 mb/s
ReqReply 1024            10,000              9,650 msg/s      9.42 mb/s
ReqReply 4096            10,000              9,789 msg/s     38.24 mb/s
ReqReply 8192            10,000              8,961 msg/s     70.01 mb/s

Latency 0     5,000     40 /  51.13 / 187    +/- 0.77  (microseconds)
Latency 8     5,000     42 /  53.88 / 268    +/- 0.82  (microseconds)
Latency 32    5,000     41 /  50.75 / 147    +/- 0.77  (microseconds)
Latency 256   5,000     39 /  51.31 / 171    +/- 0.77  (microseconds)
Latency 512   5,000     40 /  53.18 / 249    +/- 0.80  (microseconds)
Latency 1024  5,000     41 /  52.35 / 385    +/- 0.79  (microseconds)
Latency 4096  5,000     41 /  53.61 / 823    +/- 0.81  (microseconds)
Latency 8192  5,000     43 /  55.03 / 846    +/- 0.85  (microseconds)

Final memory usage is 67.79 mb / 80.00 mb / 16.00 gb free/total/max

BUILD SUCCESSFUL in 2m 35s
2 actionable tasks: 2 executed
9:17:59 AM: Task execution finished 'NatsAutoBench.main()'.

src/main/java/io/nats/client/impl/Headers.java
jameshilliard: > As in all software there are trade offs, and, again, my opinion is that this trade off is well worth it for less than 2 second loss over 50 million passes for the message.

So it seems it's actually a much larger difference than I even expected, in a number of the autobench benchmarks I'm showing using correctly upfront sized buffers as ~50% faster than the `ByteArrayBuilder` variant which is a larger difference than I'm seeing in just the message serialization benchmarks by themselves, not sure why that is but an overall 50% performance gain by correctly sizing the buffers up front seems worth some extra complexity to me.

jameshilliard:
@scottf I've rebased #364 on your latest changes, I think it came out significantly cleaner this time as I was able to adapt your `Token` class to do `ByteBuffer` slicing without significant logic changes.

src/main/java/io/nats/client/impl/Headers.java
jameshilliard: > In my opinion, it's not just less verbose, it's significantly simpler. 

Your latest changes made it much easier for me to adapt the internals of `Headers` to use `ByteBuffer`'s cleanly while still doing precise buffer allocations and zero-copy parsing of headers.

jameshilliard:
Potential fix for the `nats-server` install failure:
https://github.com/nats-io/nats.java/blob/fd9b4b318fffaecda0235d30aebb4f8b6d7fd661/.travis.yml#L15-L20
https://github.com/nats-io/nats.java/blob/fd9b4b318fffaecda0235d30aebb4f8b6d7fd661/.travis.yml#L53-L55

install_deps.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
ColinSullivan1, line 1: If you could use this instead I believe it'll solve the CI problems.

```bash
#!/bin/sh

echo "Downloading go."
wget --quiet https://golang.org/dl/go1.15.5.linux-amd64.tar.gz -O tmp.tar.gz

echo "Installing go."
tar -C `pwd` -xzf tmp.tar.gz
export GOROOT=`pwd`/go
export PATH=$GOROOT/bin:$PATH

echo "Getting the nats-server"

curdir=`pwd`
mkdir -p $GOROOT/src/github.com/nats-io/nats-server
git clone https://github.com/nats-io/nats-server.git $GOROOT/src/github.com/nats-io/nats-server
cd $GOROOT/src/github.com/nats-io/nats-server
GO111MODULE=on go install -v ./...

cd $curdir

# nats-server should be in `pwd`/go/bin
mkdir nats-server
cp $GOROOT/bin/nats-server nats-server

echo "NATS server version:"
`pwd`/nats-server/nats-server --version



```

jameshilliard: Isn't `eval "$(GIMME_GO_VERSION=1.15 gimme)" ` the recommended way to install a specific go version on travis?

ColinSullivan1 (approved):
LGTM!

install_deps.sh
ColinSullivan1: Thanks! I'll check that out for the JetStream PR.

PR #364, opened 2020-11-17
Refactor Headers to use ByteBuffer's internally.

jameshilliard:
#362 with `Headers` refactored to use `ByteBuffer`'s.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsMessage.java
+++ b/src/main/java/io/nats/client/impl/NatsMessage.java
@@ -19,6 +19,7 @@ import io.nats.client.Subscription;
 import io.nats.client.support.IncomingHeadersProcessor;
 import io.nats.client.support.Status;
 
+import java.nio.ByteBuffer;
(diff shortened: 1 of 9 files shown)


src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -220,19 +305,149 @@ public Connection getConnection() {
...
+                "\n  totLen=" + totLen +
+                "\n  subscription=" + subscription +
+                "\n  next=" + next;
+    }
+
+    /**
+     * The builder is for building normal publish/request messages,
+     * as an option for client use developers instead of the normal constructor
+     */
+    public static class Builder {
```
jameshilliard, line 369: You could have this class do `String` -> `ByteBuffer` conversions prior to `NatsMessage` instance initialization.

scottf:
This will not be merged. Thank you though for the effort.

PR #369, opened 2020-12-07
Add dispatcher example

ColinSullivan1:
Also added additional comments.

Signed-off-by: Colin Sullivan <colin@synadia.com>

### DIFF
--- a/src/examples/java/io/nats/examples/NatsJsPub.java
+++ b/src/examples/java/io/nats/examples/NatsJsPub.java
@@ -34,22 +34,6 @@ public class NatsJsPub {
             + "\nSet the environment variable NATS_CREDS to use JWT/NKey authentication by setting a file containing your user creds.\n"
             + "\nUse the URL for user/pass/token authentication.\n";
 
-    public static void createTestStream(JetStream js, String streamName, String subject)
-            throws TimeoutException, InterruptedException {
-
-        // Create a stream, here will use a file storage type, and one subject,
-        // the passed subject.
-        StreamConfiguration sc = StreamConfiguration.builder().
-            name(streamName).
-            storageType(StorageType.File).
-            subjects(new String[] { subject }).
-            build();
-        
-        // Add or use an existing stream.
-        StreamInfo si = js.addStream(sc);
-        System.out.printf("Using stream %s on subject %s created at %s.\n",
-           streamName, subject, si.getCreateTime().toLocalTime().toString());
-    }
     public static void main(String[] args) {
         ExampleArgs exArgs = ExampleUtils.readPublishArgs(args, usageString);
 
@@ -65,7 +49,7 @@ public class NatsJsPub {
 
             // if a stream name is not provided, attempt to create a test stream.
             if (exArgs.stream == null) {
-                createTestStream(js, "test-stream", exArgs.subject); 
+                ExampleUtils.createTestStream(js, "test-stream", exArgs.subject); 
             }
 
             // create a typical NATS message
--- a/src/examples/java/io/nats/examples/NatsJsPub.java
+++ b/src/examples/java/io/nats/examples/NatsJsPub.java
@@ -28,7 +28,7 @@ import java.util.concurrent.TimeoutException;
 
 public class NatsJsPub {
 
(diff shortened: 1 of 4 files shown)


src/examples/java/io/nats/examples/NatsJsSubCb.java
```diff
@@ -0,0 +1,104 @@
...
+                    // If you don't disable auto ack mode NATS will automatically
+                    // ack for you.  Otherwise, you'll need to ack, e.g.
+                    // msg.ack();                    
+                }
+                msgLatch.countDown();
+            };
+
+            // Build our subscription options.  We'll create a durable subscription named
+            // "sub-example".
+            SubscribeOptions so = SubscribeOptions.builder().durable("sub-example").build();
```
matthiashanel, line 92: shouldn't sub-example be an option to pass in, defaulting to sub-example when not set?

src/examples/java/io/nats/examples/NatsJsSubCb.java
```diff
@@ -0,0 +1,104 @@
...
+                if (msg.isJetStream()) {
+                    MetaData meta = msg.metaData();
+                    System.out.printf("  Stream Name: %s\n", meta.getStream());
+                    System.out.printf("  Stream Seq:  %d\n",  meta.streamSequence());
+                    System.out.printf("  Consumer Name: %s\n", meta.getConsumer());
+                    System.out.printf("  Consumer Seq:  %d\n", meta.consumerSequence());
+
+                    // If you don't disable auto ack mode NATS will automatically
+                    // ack for you.  Otherwise, you'll need to ack, e.g.
+                    // msg.ack();
```
matthiashanel, line 85: make that an option?

src/examples/java/io/nats/examples/NatsJsSub.java
```diff
@@ -40,9 +40,15 @@ public static void main(String[] args) {
 
         try (Connection nc = Nats.connect(ExampleUtils.createExampleOptions(exArgs.server, true))) {
 
+            // Create our jetstream context to receive jetstream 
+            // messages.
             JetStream js = nc.jetStream();
 
+            // Build our subscription options.  We'll create a durable subscription named
+            // "sub-example".            
             SubscribeOptions so = SubscribeOptions.builder().durable("sub-example").build();
```
matthiashanel, line 49: same as for next review

src/examples/java/io/nats/examples/NatsJsPub.java
```diff
@@ -28,7 +28,7 @@
 
 public class NatsJsPub {
 
-    static final String usageString = "\nUsage: java NatsJsPub [-s server] [-h headerKey:headerValue]* <subject> <message>\n"
+    static final String usageString = "\nUsage: java NatsJsPub [-s server] [--stream name] [-h headerKey:headerValue]* <subject> <message>\n"
```
matthiashanel, line 31: add stream name to the other examples

matthiashanel (changes requested):

matthiashanel (commented):
LGTM

src/examples/java/io/nats/examples/NatsJsSubCb.java
ColinSullivan1: I'd like to keep this really simple - different ack modes will make this example messy for users to copy, and ack modes need to correspond to consumer configuration.  Documentation with a few new examples and with corresponding configurations will be better imo.  We had a lot of options in Stan, and the example code isn't all that usable imo.

src/examples/java/io/nats/examples/NatsJsSubCb.java
ColinSullivan1: I'll leave this hardcoded for now, otherwise one could conceivably create a configuration that doesn't match the code and run into trouble - e.g. ack modes, where consumer expects explicit, but the client does something else).  There may be other examples we can add later with different usage.

src/examples/java/io/nats/examples/NatsJsPub.java
ColinSullivan1: done, and added stream creation in the example subs so they can be run first.

PR #371, opened 2020-12-09
Run tests in github actions.

jameshilliard:
Travis is becoming increasingly unusable for open source projects as they no longer offer credits for open source builds. In addition they seem to have been getting more unreliable in general.

travis

### DIFF
--- a/.github/workflows/jnats.yml
+++ b/.github/workflows/jnats.yml
@@ -1,0 +1,38 @@
+name: "Run Java Nats Client Tests"
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [ 1.8, 9, 10, 11, 12, 13, 14, 15, 16 ]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Setup go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15.x
+      - name: Install NATS
+        run: go get github.com/nats-io/nats-server/v2@master
+        env:
+          GO111MODULE: on
+      - name: Assemble
+        run: ./gradlew assemble --info
+        env:
+          TERM: dumb
+      - name: Check
+        run: ./gradlew check --info
+        env:
+          TERM: dumb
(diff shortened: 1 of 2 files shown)


scottf:
@jameshilliard Can you take your changes to the docs for ByteArrayBuilder into it's own PR? Also regarding the actions, I'm getting around to this, but for one I'm not going to do a matrix build on push / pr, although I'll consider it in the nightly integration test I'm building.

jameshilliard:
> Can you take your changes to the docs for ByteArrayBuilder into it's own PR?

Fixed in #460 will re-base this once that's merged as this depends on that change.

> Also regarding the actions, I'm getting around to this, but for one I'm not going to do a matrix build on push / pr, although I'll consider it in the nightly integration test I'm building.

What's the issue with the matrix? Using a matrix runs the tests in parallel so I don't think it should slow things down.

> Also recently I did a major overhaul to the travis script because it was not deploying properly, well because gradle versus travis.

Not sure what's going on there...the gradle version should be independent of travis AFAIU.

> As I move to actions I'll consider moving to gradle 6, but I'm not sure about 7. Is there anything significant in 7 to be aware of?

Needed for Java 16 support.

scottf:
> > Also recently I did a major overhaul to the travis script because it was not deploying properly, well because gradle versus travis.
> 
> Not sure what's going on there...the gradle version should be independent of travis AFAIU.

```
Native file system watching is not available for this operating system.
net.rubygrapefruit.platform.NativeIntegrationLinkageException: Native library 'libnative-platform-file-events.so' could not be loaded for Linux amd64.
```
Versions 6 of gradle has the above problem on Travis. Because of it, gradle cannot determine that something is not modified and ends up re-running tasks. This is generally not a problem, except in the case of signing artifacts when differing signatures caused the deploy to sonatype to fail.

I had to use the exact version of Gradle that travis supplies as it's default as part of the java language pack. Not only that, some versions of that language pack would fail during install of the Java OpenJdk 8. Java failed for the java language pack.

Anyway, we are moving to actions.

jameshilliard:
> I had to use the exact version of Gradle that travis supplies as it's default as part of the java language pack.

Can't you just have travis execute without the wrapper so that it uses the installed gradle?

scottf:
> > I had to use the exact version of Gradle that travis supplies as it's default as part of the java language pack.
> 
> Can't you just have travis execute without the wrapper so that it uses the installed gradle?

I thought the whole point of the wrapper is to make sure you get the version you want? And what's the difference if I make the project match the travis gradle or use the travis gradle?

Anyway, let's not worry about this as we are going to move to github actions pretty soon.

jameshilliard:
> I thought the whole point of the wrapper is to make sure you get the version you want?

Well if the version you want is "the exact version of Gradle that travis supplies" then I don't see why you would want/need to use the wrapper.

> And what's the difference if I make the project match the travis gradle or use the travis gradle?

Well using an ancient wrapper means the wrapper only works on ancient java versions essentially.

scottf:
@jameshilliard Closing this b/c https://github.com/nats-io/nats.java/pull/493 moves completely to GitHub Actions and gets rid of travis. Gradle has been upgraded to 6.8.3, that's the best I can do right now. When I have cycles I will look into upgrading to 7

PR #372, opened 2020-12-09
Jetstream Updates

ColinSullivan1:
* A minor update to stream defaults
* Add an attach example

### DIFF
--- a/src/examples/java/io/nats/examples/NatsJsSubAttach.java
+++ b/src/examples/java/io/nats/examples/NatsJsSubAttach.java
@@ -1,0 +1,96 @@
+// Copyright 2015-2018 The NATS Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


src/examples/java/io/nats/examples/NatsJsSubAttach.java
```diff
@@ -0,0 +1,96 @@
...
+
+                System.out.printf("  Subject: %s\n  Data: %s\n",
+                        msg.getSubject(),
+                        new String(msg.getData(), StandardCharsets.UTF_8));
+
+                // This check may not be necessary for this example depending
+                // on how the consumer has been setup.  When a deliver subject
+                // is set on a consumer, messages can be received from applications
+                // that are NATS producers and from streams in NATS servers.
+                if (msg.isJetStream()) {
```
matthiashanel, line 79: This is not what I'd expect. This is literally a subscription using the js object. How is it somehow possible that a non js message is being received here? And if it is should it even be delivered?

ColinSullivan1: This is based on the consumer delivery subject (this does not apply to a pull consumer).  If the JS consumer has a delivery subject that overlaps with a subject being published to by a regular NATS publisher, you can (and should) get a mix of core NATS messages and Jetstream messages.

matthiashanel (approved):
LGTM

PR #374, opened 2020-12-13
jitter and randomized discovered servers for reconnect

scottf:


src/main/java/io/nats/client/Options.java
```diff
@@ -1601,34 +1603,31 @@ public boolean isDiscardMessagesWhenOutgoingQueueFull() {
...
 
-        if (uri.getHost() != null && uri.getHost() != "") {
+        if (uri.getHost() != null && uri.getHost().length() > 0) {
             if (uri.getPort() == -1) {
-                uri = new URI(uri.getScheme(), 
+                uri = new URI(uri.getScheme(),
                                 uri.getUserInfo(), 
                                 uri.getHost(),
-                                4222,
+                                DEFAULT_PORT,
```
scottf, line 1630: This constant existed but wasn't used.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -177,8 +163,9 @@ void connect(boolean reconnectOnConnect) throws InterruptedException, IOExceptio
         Collection<String> serversToTry = buildServerList();
         for (String serverURI : serversToTry) {
             if (isClosed()) {
-                break;
+                break; // goes to statement after end connect server loop
```
scottf, line 166: I added some comments to this code as I learned / understood what was going on

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -248,57 +233,58 @@ void reconnect() throws InterruptedException {
 
         while (!isConnected() && !isClosed() && !this.isClosing()) {
             List<String> serversToTry = buildServerList();
-            lastServer = serversToTry.get(serversToTry.size() - 1);
+            String lastServer = serversToTry.get(serversToTry.size() - 1);
```
scottf, line 236: again, just added some comments

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -333,26 +319,28 @@ void reconnect() throws InterruptedException {
...
+        if (trace) {
+            _trace(message);
         }
     }
 
-    long timeCheck(boolean trace, long endNanos, String format, Object... args) throws TimeoutException {
+    private void _trace(String message) {
+        String timeStr = DateTimeFormatter.ISO_TIME.format(LocalDateTime.now());
+        System.out.println("[" + timeStr + "] connect trace: " + message);
+    }
```
scottf, line 335: I used underbar as in indicator for a private delegate. Other functions call (delegate) to this because they have common functionality, like the entire trace message code appeared twice.
In this case, I think it's reasonable for optimization. Not sure about printf, but print and printlin both are synchronized calls so you want to make as few as possible.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1524,30 +1512,72 @@ public long getMaxPayload() {
...
-        options.getServers().stream().forEach(x -> {
-            String uri = x.toString();
-            if (!check.contains(uri)) {
-                servers.add(uri);
-                check.add(uri);
-            }
-        });
+        return getServersList();
+    }
+
```
scottf, line 1517: refactored so I could re-use the code. This method returns a generic collection, but was made from a list. At the same time it wanted an actual set so there would not be dupes, I already mentioned that I normalized so dupes would actually be found. But I think it was a list because ordering is maintained in a list, versus a set, where the order is not guaranteed. So I left it as a list.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1524,30 +1512,72 @@ public long getMaxPayload() {
...
+        List<String> servers = shuffleWithoutCurrent( addDiscoveredServers(new ArrayList<>()) );
+
+        // configured get shuffled
+        servers.addAll( shuffleWithoutCurrent( addConfiguredServers(new ArrayList<>()) ) );
+
+        // because we still want it in the list
+        if (currentServer != null) {
+            servers.add(currentServer);
+        }
+
```
scottf, line 1537: get the discovered servers and shuffle them (without the current). separately get the configured (in options) and shuffle them (also w/o current) then add the configured and then the current to the discovered list and return

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1666,7 +1695,7 @@ void waitFor(Duration timeout, Predicate<Void> test) throws InterruptedException
 
     void waitForReconnectTimeout() {
         Duration waitTime = options.getReconnectWait();
-        long currentWaitNanos = (waitTime != null) ? waitTime.toNanos() : -1;
+        long currentWaitNanos = (waitTime == null) ? -1 : ThreadLocalRandom.current().nextLong(waitTime.toNanos());
```
scottf, line 1698: Jitter!

ColinSullivan1: Nice!

ColinSullivan1 (approved):
LGTM.  All ready to merge?

ColinSullivan1:
I triggered a re-run of CI.  If the test still fails, there may be a flapper there.  It was a reconnect test...

PR #391, opened 2020-12-28
reconnect-jitter-handler-serverinfo-tests

scottf:


src/main/java/io/nats/client/ServerInfo.java
```diff
@@ -0,0 +1,56 @@
...
+    int getProtocolVersion();
+    boolean isHeadersSupported();
+    boolean isAuthRequired();
+    boolean isTLSRequired();
+    long getMaxPayload();
+    String[] getConnectURLs();
+    byte[] getNonce();
+    boolean isJetStreamAvailable();
+    int getClientId();
+    String getClientIp();
```
ColinSullivan1, line 53: Decided on the string...

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1508,38 +1517,89 @@ public long getMaxPayload() {
...
+
+    private List<String> getReconnectServerList() {
+        if (options.isNoRandomize()) {
+            return getServersList();
+        }
+
+        // prioritize discovered over configured
+        List<String> servers = new ArrayList<>();
+        addDiscoveredServers(servers);
+        shuffleWithoutCurrent(servers);
```
ColinSullivan1, line 1538: Couldn't shuffle be called once after adding discovered and then configured?

ColinSullivan1 (commented):
Looks good - just a few questions...

src/main/java/io/nats/client/ServerInfo.java
scottf: Agreed. The InetAddress has static methods that the dev can call directly to find out info you were describing.

src/main/java/io/nats/client/impl/NatsConnection.java
scottf: The thought here was that discovered servers are the most viable and contain the list of configured servers, although possible in a ip form instead of a dns name form. In fact I'm not even sure of the value of adding the configured servers in the list once the discovered are populated.

ColinSullivan1: Discovered servers should be only servers that weren't originally configured.  For example, if I have servers A, B, and C configured in the client then added D and E, the client would have A,B,C as configured servers and D and E as a discovered servers.  Randomization should be applied to the whole list (sans current during reconnect), so each server will be weighted equally in probability w/r/t reconnect order.  Someday we'll have classes of servers for regional preference (e.g. try this set first, then if all those fail, try another set), but that's TBD at the moment.

There is a chance for duplicates based on IP vs hostname, but server configuration allows ops to handle that.

ColinSullivan1 (approved):
LGTM - Thanks!

PR #394, opened 2020-12-30
get the validation right

scottf:
NatsMesasage: 
. . . . subject is required, but no other checks

NatsJetStream.addConsumer:
. . . . subject: No subject check, it's not required
. . . . stream:, required, alphanum or dash `^[a-zA-Z0-9-]+$`

NatsJetStream.subscribe:
. . . . subject: not required but if provided cannot contain whitepsace
. . . . queue: required and cannot contain whitespace


ColinSullivan1 (approved):
LGTM.  Thanks!

PR #400, opened 2021-01-27
Split Jetstream management api from regular api

scottf:
Separate the JetStream Management API from the Publish/Subscribe portion


src/main/java/io/nats/client/JetStreamManagement.java
```diff
@@ -0,0 +1,436 @@
...
+     * publish during reconnect.
+     *
+     * @param subject the subject to send the message to
+     * @param body the message body
+     * @return The publish acknowledgement
+     * @throws IOException covers various communication issues with the NATS
+     *         server such as timeout or interruption
+     * @throws JetStreamApiException the request had an error related to the data
+     */
+    PublishAck publish(String subject, byte[] body) throws IOException, JetStreamApiException;
```
ColinSullivan1, line 131: Don't think these should be needed here...

src/main/java/io/nats/client/JetStreamManagementOptions.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private JetStreamManagementOptions(String prefix, Duration requestTimeout) {
+        super(prefix, requestTimeout, false);
+    }
+
+    /**
+     * SubscribeOptions can be created using a Builder. The builder supports chaining and will
+     * create a default set of options if no methods are calls.
+     */
+    public static class Builder {
```
ColinSullivan1, line 34: Can you subclass from JetstreamOptions.Builder, or if not, encapsulate?

src/main/java/io/nats/client/SubscribeOptions.java
```diff
@@ -104,17 +103,26 @@ public static Builder builder() {
         return new Builder();
     }
 
+    // TODO
```
ColinSullivan1, line 106: OK at this point, but not sure if you wanted to leave this in for this PR.

src/main/java/io/nats/client/JetStreamManagementOptions.java
scottf: hmmm, maybe

PR #414, opened 2021-02-23
pull simplify

scottf:


src/examples/java/io/nats/examples/NatsJsPullSubBase.java
```diff
@@ -13,9 +13,7 @@
 
 package io.nats.examples;
 
-import io.nats.client.JetStream;
-import io.nats.client.JetStreamSubscription;
-import io.nats.client.Message;
+import io.nats.client.*;
```
scottf, line 16: I would usually not "refactor" for examples but I wanted to ensure that I used the same exact functionality for each example. If you really don't like this I can copy paste into each example.

src/examples/java/io/nats/examples/NatsJsPullSubBase.java
```diff
@@ -25,50 +23,119 @@
...
+
+    public static List<Message> readMessagesAck(JetStreamSubscription sub, boolean noisy) throws InterruptedException {
         List<Message> messages = new ArrayList<>();
         Message msg = sub.nextMessage(Duration.ofSeconds(1));
         boolean first = true;
         while (msg != null) {
             if (first) {
                 first = false;
-                System.out.print("Read/Ack ->");
+                if (noisy) {
```
ColinSullivan1, line 81: Nitpick...  Most other places in NATS we use `verbose`.

src/test/java/io/nats/client/impl/JetStreamPullTests.java
```diff
@@ -328,48 +220,189 @@ public void expireFuture_AckModeNext_ExpireModeLeave() throws Exception {
...
+
+            sub.pull(1);
+
+            Message message = sub.nextMessage(Duration.ofSeconds(1));
+            assertNotNull(message);
+            String data = new String(message.getData());
+            assertEquals("NAK1", data);
+            message.nak();
+
+            // not 100 % sure what this message is
```
ColinSullivan1, line 247: Is this comment still valid - Could `nak()` be doing a request/reply?

ColinSullivan1 (approved):
LGTM, one question on `nak()` test and a nitpick.

src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -458,11 +463,14 @@ void setData(byte[] data) {
 
         NatsMessage getMessage() {
             NatsMessage message;
-            if (NatsJetStreamMessage.isJetStream(replyTo)) {
+            if (replyTo != null && replyTo.startsWith("$JS")) {
```
ColinSullivan1, line 466: Check the prefix?

src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -505,4 +513,22 @@ boolean isProtocol() {
             return true;
         }
     }
+
+    static class StatusMessage extends InternalMessage {
```
ColinSullivan1, line 517: Might be handy to provide a toString() override here.

ColinSullivan1 (approved):
LGTM - the prefix can be addressed in a later PR.

PR #415, opened 2021-03-04
custom prefix

scottf:


src/main/java/io/nats/client/JetStreamOptions.java
```diff
@@ -63,7 +64,7 @@ public static Builder builder() {
      * @return the configuration
      */
     public static JetStreamOptions defaultOptions() {
-        return new JetStreamOptions(null, null);
+        return DEFAULT_JS_OPTIONS;
```
ColinSullivan1, line 67: If someone gets the defaults, modifies them, then calls `defaultOptions()` again they would have modified options...  Not sure this is desired behavior.

ColinSullivan1: Oh, I see...   there are no setters, only a builder so once created options are immutable.  nevermind.

src/main/java/io/nats/client/JetStreamSubscription.java
```diff
@@ -56,9 +56,12 @@
...
 
     /**
+     * Fetch (pull) the list
      *
      * @param batchSize the size of the batch
      * @param timeout the timeout
+     *
+     * @return the list of messages
      */
     List<Message> fetch(int batchSize, Duration timeout);
```
ColinSullivan1, line 66: Did you intend to include fetch in this PR?

src/main/java/io/nats/client/impl/JsPrefixManager.java
```diff
@@ -0,0 +1,49 @@
+package io.nats.client.impl;
```
ColinSullivan1, line 1: Copyright...

src/main/java/io/nats/client/impl/JsPrefixManager.java
```diff
@@ -0,0 +1,49 @@
...
+        return prefix;
+    }
+
+    public static boolean isJsMessage(final String replyTo) {
+        if (replyTo != null) {
+            if (replyTo.startsWith(JS_PREFIX)) {
+                return true;
+            }
+
+            synchronized (JS_PREFIXES) {
```
ColinSullivan1, line 38: Wish there was a way to avoid this... this is in the fastpath, but I guess need to do this when there are multiple prefixes.  

Could use a synchronized list, get the count (synchronized) and then iterate through up to the count since prefixes will never change order or be deleted once created.

Can optimize later, if need be....

src/main/java/io/nats/client/impl/NatsMessage.java
```diff
@@ -462,14 +461,14 @@ void setData(byte[] data) {
...
-            if (replyTo != null && replyTo.startsWith("$JS")) {
-                message = new NatsJetStreamMessage();
-            }
-            else if (status != null) {
+            NatsMessage message = null;
+            if (status != null) {
                 message = new StatusMessage();
             }
-            else {
+            else if (JsPrefixManager.isJsMessage(replyTo)) {
```
ColinSullivan1, line 468: nitpick... `hasPrefix` might be a better name.   Up to you...

src/main/java/io/nats/client/impl/JsPrefixManager.java
scottf: I am using a synchronized list. In the javadoc itself, it says if you iterate the list, you must synchronize it. I suppose I could make a read only copy after every update.

src/main/java/io/nats/client/impl/JsPrefixManager.java
scottf: fixing

src/main/java/io/nats/client/JetStreamSubscription.java
scottf: I fixed the javadoc, this was already in master

src/main/java/io/nats/client/support/Validator.java
```diff
@@ -203,12 +203,13 @@ public static boolean containsDotWildGt(String s) {
...
     }
 
-    public static boolean containsWildGt(String s) {
+    public static boolean containsWildGtDollar(String s) {
         if (s != null) {
             for (int i = 0; i < s.length(); i++) {
                 switch (s.charAt(i)) {
                     case '*':
                     case '>':
+                    case '$':
```
ColinSullivan1, line 212: Should check for space and tab too (had to double-check myself):

https://docs.nats.io/nats-protocol/nats-protocol#protocol-conventions

src/main/java/io/nats/client/impl/JsPrefixManager.java
scottf: for js messages from the server, the majority I think, this is going to be $JS.ACK, so won't even go into the loop checking.

scottf: Good idea, implementing

ColinSullivan1 (changes requested):
Overall LGTM after a minor change to check for spaces/tabs in the prefix.  Technically they'd work but isn't adherent to our publish specs.

src/main/java/io/nats/client/impl/JsPrefixManager.java
scottf: let's leave this for now. I think it might be more expensive to know the count and get(index) b/c each get index will do a synchronized so this will just have one for the entire list instead of each item.

src/main/java/io/nats/client/impl/NatsMessage.java
scottf: changed

src/main/java/io/nats/client/support/Validator.java
scottf: fixed

ColinSullivan1 (approved):
LGTM.  Thanks!

PR #426, opened 2021-03-12
Adds websocket client support, resolves #399

### LINKED ISSUE #399: Websocket Support - Java
I'm working on a project which requires connections to be established over HTTP (potentially over an HTTP proxy). To that end, I was planning to write a `DataPort` implementation which checks if the `serverURI` has a scheme of "ws" (or "wss" if secure is desired) and then performs an HTTP connection (potentially respecting any desire to use a proxy).

Is there any pre-existing code which might already implement this?

If not, is there others which would benefit from this code being contributed back to this project?

Thanks,
-Brian

brimworks:
Due to requirements of a project that I'm working on, I need the java client library to use the websocket port. These changes add that support.


scottf:
@brimworks Thanks for the PR, adding websocket support to the client is absolutely something want to do. Right now we are focusing on JetStream which will be released as beta in the next couple weeks. This will be the priority right after our JetStream beta is released and then I'll have the bandwidth to look into this.

brimworks:
@scottf Yes, I definitely understand. Getting a NATS 2.2 released is definitely higher priority for your team. I tried to make this change as minimum as possible and no added 3rd party dependencies. I can also work on getting more code coverage (high bar!).

Thanks!

brimworks:
I've added more tests, and added support for HTTP request intercepting. This was necessary for our use-case, since we are tunneling through Cloudflare, and Cloudflare Access requires us to set a cookie for the cloudflare auth.

scottf:
@brimworks I'm finally getting around to this. can you rebase? 

Regarding HttpHeaders, I'd like you to leverage io.nats.client.impl.Headers maybe by extending the existing Headers class. I don't want to whole separate classes.

Can you also update the copyright to
```
// Copyright 2020 The NATS Authors
```

brimworks:
Sounds good. I'll update appropriately soon. As a side-note, it sounds like we may need to support proxying via HTTP CONNECT verb... so this "request interceptor" mechanism is probably not sufficient for supporting proxying.

Is HTTP proxies something that you wanted to support in the NATS code-base? Or do you think it would be better to modify the "request interceptor" abstraction so it could work for HTTP proxying?

brimworks:
So, I just noticed ` io.nats.client.impl.Headers` is in an "implementation" package, but the HttpRequest which needs to expose that is NOT in an implementation package. Do you want me to move that out of the implementation subdirectory?

scottf:
> So, I just noticed ` io.nats.client.impl.Headers` is in an "implementation" package, but the HttpRequest which needs to expose that is NOT in an implementation package. Do you want me to move that out of the implementation subdirectory?

We must maintain backward compatibility so Headers cannot be moved. It is public so it is extendable and is already available to users directly from the NatsMessage anyway. Can you please clarify what you meant by "HttpRequest which needs to expose" Probably HttpRequest should be in impl. I think HttpRequest probably should go in the impl package, not the client. 

Can you please tell me more about HttpRequest and how it's going to be used. Is this exposed to the user? I need to learn more about websockets, I'd appreciate you pointing me to some reading material.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -502,17 +502,35 @@ void checkVersionRequirements() throws IOException {
         }
     }
 
-    void upgradeToSecureIfNeeded() throws IOException {
+    private static boolean isWebsocket(String serverURI) {
```
scottf, line 505: Can this be moved into the options class or some support class?

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1548,7 +1566,7 @@ private void addDiscoveredServers(List<String> servers) {
             for (String uri : info.getConnectURLs()) {
                 try {
                     // call to createURIForServer is to parse (normalize)
-                    addNoDupes(servers, options.createURIForServer(uri).toString());
+                    addNoDupes(servers, options.createURIForServer(uri, isWebsocket(this.currentServer)).toString());
```
scottf, line 1569: see comment regarding moving isWebsocket

src/main/java/io/nats/client/impl/SocketDataPort.java
```diff
@@ -24,14 +26,28 @@
...
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * This class is not theadsafe.  Caller must ensure thread safety.
  */
 public class SocketDataPort implements DataPort {
+    private static final int MAX_LINE_LEN = 8192;
```
scottf, line 47: I'd be interested in implementing some sort of factory here to get an implementation based on the protocol, instead of just making this class bigger. May not be practical, just something to consider.

src/main/java/io/nats/client/impl/WebsocketFrameHeader.java
```diff
@@ -0,0 +1,264 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package io.nats.client.impl;
+
+public class WebsocketFrameHeader {
```
scottf, line 16: can WebsocketFrameHeader, WebsocketInputStream and WebsocketOutputStream be moved into the support package

brimworks:
>  Can you please clarify what you meant by "HttpRequest which needs to expose" Probably HttpRequest should be in impl. I think HttpRequest probably should go in the impl package, not the client.
>
> Can you please tell me more about HttpRequest and how it's going to be used. Is this exposed to the user? I need to learn more about websockets, I'd appreciate you pointing me to some reading material.

The only reason I added the HttpRequest object was to support "filtering". We needed it to add a cloudflare token header for our test environments. However, this "filtering" abstraction is not sufficient to support HTTP proxying, so I do think it is worth revisiting that abstraction so that HTTP proxying can be supported (perhaps via a separate add-on?).

Regarding websocket documentation, I'd say the RFC does a pretty good job of explaining it all in-depth:

https://datatracker.ietf.org/doc/html/rfc6455

I think the MDN article is good at giving a high-level overview though:

https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers

src/main/java/io/nats/client/impl/SocketDataPort.java
brimworks: I think that would be a MUCH better design... I just wanted to provide the minimal changes necessary to get websocket support. More specifically, if the Options class allowed one to add a `DataPort` factory that is keyed off of the "schema", then supporting new transports (such as websockets) might be as simple as defining a new `DataPort` factory.

However, the current `DataPort` abstraction could probably use some work. For example, instead of having the `connect()` method, that should probably be a method on the `DataPort` factory, and instead of passing in a `NatsConnection` it should probably pass in an `Options` object.

On a completely separate side-note, has anyone considered changing the library to use "reactive streams"? ...in which case the code would need to be substantially modified (although the "heart" of the code is really just the WebsocketFrameHeader and the initial hand-shake which is currently in the SocketDataPort.

brimworks:
@scottf note that I can't get my new tests to pass, since this code needs to be fixed:

https://github.com/nats-io/java-nats-server-runner/blob/main/src/main/java/nats/io/NatsServerRunner.java#L264

Specifically, the redundant `--port` option causes a port collision with the ws.port (aka the websocket port specified in the configuration file). Would you like to file a pull request against the `java-nats-server-runner` to fix that problem? Or is there a better way to fix it?

scottf:
> Would you like to file a pull request against the `java-nats-server-runner` to fix that problem? 
Yes, please go ahead and make a pr.

brimworks:
Sorry about my slow turn around. Work has gotten pretty busy lately. I'll see about posting the `java-nats-server-runner` PR so I can get this PR building again soon.

brimworks:
Unfortunately, I'm just not familiar enough with gradle in order to figure out how to build the `java-nats-server-runner` repository locally and then consume the build artifacts from `nats.java`. I tried to run `gradle publishToMavenLocal` in the `java-nats-server-runner` repository, which placed the artifacts in `~/.m2/repository/io/nats/jnats-server-runner/1.0.6` as expected... but when I build `nats.java` it doesn't seem to pickup my changes (since I just added a throw statement to `NatsServerRunner.java` constructor, and that throw statement is never reached).

I believe the fix I need is as simple as this:

```diff
diff --git src/main/java/nats/io/NatsServerRunner.java src/main/java/nats/io/NatsServerRunner.java
index 63f2938..04ce182 100644
--- src/main/java/nats/io/NatsServerRunner.java
+++ src/main/java/nats/io/NatsServerRunner.java
@@ -261,8 +261,6 @@ public class NatsServerRunner implements AutoCloseable {

                     if (portMatcher.find()) {
                         line = line.replace(portMatcher.group(1), String.valueOf(_port));
-                        cmd.add("--port");
-                        cmd.add(String.valueOf(_port));
                     }

                     writer.write(line);
```

...but I just can't figure out how to test this change locally. Do you have any pointers?

I also tried modifying the `build.gradle` in `nats.java` with this change to make it pickup my local installed version:

```
diff --git build.gradle build.gradle
index 0fbce34..cc41d88 100644
--- build.gradle
+++ build.gradle
@@ -72,6 +72,7 @@ java {
 }

 repositories {
+    mavenLocal()
     jcenter()
 }
```

...but still it didn't pickup my change.

Sorry that I'm a gradle newbie :(. I can keep trying to figure this out on my own, but I just don't have the time right now for this.

brimworks:
Finally, I figured out how to test locally, and I have this PR out: https://github.com/nats-io/java-nats-server-runner/pull/13

brimworks:
I've updated the PR per your suggestions, except for the major refactor of SocketDataPort. I need to add support for HTTP proxy servers as a fast-follow PR, and perhaps that can be when SocketDataPort is refactored?

brimworks:
Obviously the build will fail until the java-nats-server-runner PR is merged.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.1.1-bin.zip
```
scottf, line 3: I need to move from travis to github actions b/c our travis build only works on 5.1.1

brimworks:
This PR is replaced by this PR: https://github.com/nats-io/nats.java/pull/502 ...which depends on the DataPort refactor PR: https://github.com/nats-io/nats.java/pull/494

brimworks:
Scrapping the nio based PRs, it was 4x more complex, and re-opening this PR.

brimworks:
Replaced by https://github.com/nats-io/nats.java/pull/513  since coveralls isn't working properly when building outside of the nats-io account.

PR #437, opened 2021-03-24
2.10 refactor and stream enhancements

scottf:
A1. Most of the changes in this PR are due to repackaging. Api bean objects were moved from the impl package, or from being static inner classes into the api package. This had an affect of changing imports. Also, a few interfaces were removed when they represented a bean, because there will only ever be one concrete implementation. Those beans were named as the interface name and placed in the api package.

A2. There were some additional organization, moving support classes from the impl package to the support package.

The end result of the repackaging ends up organizing things more properly by function. Care was taken to not disrupt packaging  of features present before 2.9, so 2.8 users should have no problem dropping 2.10 in. 2.9 users will have to review their imports, but their ide should be able to assist. Since I mentioned an ide, IntelliJ did all the heavy lifting on repackaging. For instance, it automatically updated imports when a class was moved. No tests were broken.

B. There were some improvements to support classes, specifically ByteArrayBuilder has improved functionality intended to support performance improvements. These were initially done for a different branch, but I considered them important enough to make sure they did not get lost in case we throw away those performance branches. I also made an important fix to the Json utils where it handled finding an object, specifically when there where two fields with the same name. 

C. Mirror and stream support was added. Several api beans where added to support this, 2 unit tests were copied from the nats server (translated from go, but functionally identical) Tests were updated to provide coverage for the beans added to support this feature.


src/examples/java/io/nats/examples/ExampleUtils.java
```diff
@@ -132,4 +132,27 @@ private static void usage(String usageString) {
         System.out.println(usageString);
         System.exit(-1);
     }
+
+    public static String uniqueEnough() {
```
ColinSullivan1, line 136: Why not use a NUID?  `NUID.nextGlobal();`

src/examples/java/io/nats/examples/NatsJsPullSubIterate.java
```diff
@@ -18,6 +18,9 @@
 import java.time.Duration;
 import java.util.Iterator;
 
+import static io.nats.examples.NatsJsUtils.createStreamThrowWhenExists;
+import static io.nats.examples.NatsJsUtils.publishDontWait;
```
ColinSullivan1, line 22: Is this used?

scottf: example helpers.

src/examples/java/io/nats/examples/ExampleUtils.java
scottf: I specifically wanted a short, readable but unique enough piece of information. Nuid is almost 3 times as long as I wanted as well and is mixed case. Similar reason to not just using the currentTimeMillis

src/main/java/io/nats/client/JetStreamOptions.java
```diff
@@ -1,111 +1,111 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package io.nats.client;
+
+import io.nats.client.support.JsPrefixManager;
```
scottf, line 16: I don't know why it's showing the entire class as changed. This is the only change.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -11,136 +11,44 @@
...
-    private String filterSubject;
+    private final String filterSubject;
     private final String sampleFrequency;
     private final long rateLimit;
-    private long maxAckPending;
+    private final long maxAckPending;
 
     // for the response from the server
-    ConsumerConfiguration(String json) {
+    public ConsumerConfiguration(String json) {
```
ColinSullivan1, line 51: Probably document that this isn't used in an application.

src/main/java/io/nats/client/api/ConsumerCreateRequest.java
```diff
@@ -0,0 +1,57 @@
...
+package io.nats.client.api;
+
+import io.nats.client.support.JsonSerializable;
+import io.nats.client.support.JsonUtils;
+
+import static io.nats.client.support.ApiConstants.CONFIG;
+import static io.nats.client.support.ApiConstants.STREAM_NAME;
+import static io.nats.client.support.JsonUtils.*;
+
+public class ConsumerCreateRequest implements JsonSerializable {
```
ColinSullivan1, line 23: Doc that this is an internal class, or add a directive to remove from documentation...

src/main/java/io/nats/client/api/External.java
```diff
@@ -0,0 +1,84 @@
...
+import io.nats.client.support.JsonSerializable;
+import io.nats.client.support.JsonUtils;
+
+import static io.nats.client.support.ApiConstants.*;
+import static io.nats.client.support.JsonUtils.*;
+
+/**
+ * External Information
+ */
+public class External implements JsonSerializable {
```
ColinSullivan1, line 25: I'm wondering if this needs a more descriptive name.  Not sure what it is at first glance...

src/main/java/io/nats/client/api/ServerInfo.java
```diff
@@ -1,365 +1,345 @@
...
+                i++;
+            }
+            sb.append(ch);
+        }
+        return sb.toString();
+    }
+
+    @Override
+    public String toString() {
+        return "NatsServerInfo{" +
```
ColinSullivan1, line 323: `NatsServerInfo` -> `ServerInfo`

ColinSullivan1 (changes requested):
As discussed offline, perhaps more doc and hiding/documenting APIs that shouldn't be used.

PR #446, opened 2021-04-05
improve cleanup of async responses

scottf:
also some unit tests on jetstream accounts

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -754,8 +754,7 @@ class NatsConnection implements Connection {
 
         NatsMessage nm = new NatsMessage(subject, replyTo, new Headers(headers), data, utf8mode);
 
-        Connection.Status stat = this.status;
-        if ((stat == Status.RECONNECTING || stat == Status.DISCONNECTED)
+        if ((this.status == Status.RECONNECTING || this.status == Status.DISCONNECTED)
                 && !this.writer.canQueue(nm, options.getReconnectBufferSize())) {
             throw new IllegalStateException(
                     "Unable to queue any more messages during reconnect, max buffer is " + options.getReconnectBufferSize());
@@ -875,10 +874,10 @@ class NatsConnection implements Connection {
             throw new IllegalStateException("Connection is Draining");
         }
 
+        NatsSubscription sub;
         long sidL = nextSid.getAndIncrement();
         String sid = String.valueOf(sidL);
 
-        NatsSubscription sub;
         if (isJetStream) {
             sub = new NatsJetStreamSubscription(sid, subject, queueName, this, dispatcher);
         } else {
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -944,22 +944,24 @@ class NatsConnection implements Connection {
     void cleanResponses(boolean closing) {
         ArrayList<String> toRemove = new ArrayList<>();
 
-        responsesAwaiting.forEach((token, f) -> {
+        responsesAwaiting.forEach((key, future) -> {
             boolean remove = false;
-            if (f.isDone()) {
+            if (future.hasExceededTimeout()) {
                 remove = true;
+                future.cancelTimedOut();
             }
             else if (closing) {
                 remove = true;
-                f.cancelClosing();
+                future.cancelClosing();
             }
-            else if (f.hasExceededTimeout()) {
+            else if (future.isDone()) {
+                // done should have already been removed, not sure if
+                // this even needs checking, but it won't hurt
                 remove = true;
-                f.cancelTimedOut();
             }
 
             if (remove) {
-                toRemove.add(token);
+                toRemove.add(key);
                 statistics.decrementOutstandingRequests();
             }
         });
@@ -968,10 +970,10 @@ class NatsConnection implements Connection {
             responsesAwaiting.remove(token);
         }
 
-        toRemove.clear();
-        responsesRespondedTo.forEach((token, f) -> {
-            if (f.hasExceededTimeout()) {
-                toRemove.add(token);
+        toRemove.clear(); // just reuse this
+        responsesRespondedTo.forEach((key, future) -> {
+            if (future.hasExceededTimeout()) {
+                toRemove.add(key);
             }
         });
 
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -56,7 +56,6 @@ class NatsConnection implements Connection {
                                                     // connecting
 
     private Status status;
-    private final AtomicLong statusId;
     private final ReentrantLock statusLock;
     private final Condition statusChanged;
 
@@ -109,7 +108,6 @@ class NatsConnection implements Connection {
         this.statusLock = new ReentrantLock();
         this.statusChanged = this.statusLock.newCondition();
         this.status = Status.DISCONNECTED;
-        this.statusId = new AtomicLong();
         this.reconnectWaiter = new CompletableFuture<>();
         this.reconnectWaiter.complete(Boolean.TRUE);
 
@@ -941,7 +939,7 @@ class NatsConnection implements Connection {
         return responseInbox.substring(len);
     }
 
-    void cleanResponses(boolean cancelIfRunning) {
+    void cleanResponses(boolean closing) {
         ArrayList<String> toRemove = new ArrayList<>();
 
         responses.forEach((token, f) -> {
@@ -949,13 +947,13 @@ class NatsConnection implements Connection {
             if (f.isDone()) {
                 remove = true;
             }
-            else if (f.isExpired(statusId.get()) || cancelIfRunning) {
+            else if (closing) {
                 remove = true;
-                try {
-                    f.cancel(true); // does nothing if already done
-                } catch (CancellationException e) {
-                    // Expected
-                }
+                f.cancelClosing();
+            }
+            else if (f.isOrphaned()) {
+                remove = true;
+                f.cancelOrphaned();
             }
 
             if (remove) {
@@ -1000,7 +998,7 @@ class NatsConnection implements Connection {
         return requestFutureInternal(message.getSubject(), message.getHeaders(), message.getData(), message.isUtf8mode(), null, true);
     }
 
-    CompletableFuture<Message> requestFutureInternal(String subject, Headers headers, byte[] data, boolean utf8mode, Duration timeoutIfKnown, boolean cancelOn503) {
+    CompletableFuture<Message> requestFutureInternal(String subject, Headers headers, byte[] data, boolean utf8mode, Duration orphanedTimeout, boolean cancelOn503) {
         checkPayloadSize(data);
 
         if (isClosed()) {
@@ -1023,7 +1021,7 @@ class NatsConnection implements Connection {
         boolean oldStyle = options.isOldRequestStyle();
         String responseInbox = oldStyle ? createInbox() : createResponseInbox(this.mainInbox);
         String responseToken = getResponseToken(responseInbox);
-        NatsRequestCompletableFuture future = new NatsRequestCompletableFuture(cancelOn503, timeoutIfKnown, statusId.get());
+        NatsRequestCompletableFuture future = new NatsRequestCompletableFuture(cancelOn503, orphanedTimeout);
 
         if (!oldStyle) {
             responses.put(responseToken, future);
@@ -1577,8 +1575,6 @@ class NatsConnection implements Connection {
             statusChanged.signalAll();
             statusLock.unlock();
         }
-
-        statusId.incrementAndGet();
 
         if (this.status == Status.DISCONNECTED) {
             processConnectionEvent(Events.DISCONNECTED);
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -18,6 +18,7 @@ import io.nats.client.ConnectionListener.Events;
 import io.nats.client.api.ServerInfo;
 import io.nats.client.impl.NatsMessage.ProtocolMessage;
 import io.nats.client.support.ByteArrayBuilder;
+import io.nats.client.support.NatsRequestCompletableFuture;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
@@ -72,7 +73,7 @@ class NatsConnection implements Connection {
     private Map<String, NatsSubscription> subscribers;
     private Map<String, NatsDispatcher> dispatchers; // use a concurrent map so we get more consistent iteration
                                                      // behavior
-    private Map<String, CompletableFuture<Message>> responses;
+    private Map<String, NatsRequestCompletableFuture> responses;
     private ConcurrentLinkedDeque<CompletableFuture<Boolean>> pongQueue;
 
     private String mainInbox;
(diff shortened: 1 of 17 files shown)


src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -947,8 +948,10 @@ String getResponseToken(String responseInbox) {
     void cleanResponses(boolean cancelIfRunning) {
         ArrayList<String> toRemove = new ArrayList<>();
 
+        long cleanMillis = this.options.getRequestCleanupInterval().toMillis();
+
         responses.forEach((token, f) -> {
-            if (f.isDone() || cancelIfRunning) {
+            if (f.isDone() || f.isOlderThan(cleanMillis) || cancelIfRunning) {
```
scottf, line 954: options.getRequestCleanupInterval() (line 951) defaults to 5 seconds. It seems to me that if we haven't gotten a response in 5 seconds, it's fair to say it's not coming back. I put this in b/c this could be a memory leak, because the future might never be done (isDone, line 954)

ColinSullivan1: Does this mean the lesser of RequestCleanupInterval or original request timeout?  Let's say I had a 30s ack timeout on a JS publish... wouldn't this expire in 5?

scottf: I guess I could set it to something like 5 or 10 minutes. We just have to clear these eventually.

ColinSullivan1: How about setting it to the timeout used (user provided or default) + 250ms or something, and hold that value in the future itself?

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -55,44 +56,45 @@
...
 
-    private Map<String, NatsSubscription> subscribers;
-    private Map<String, NatsDispatcher> dispatchers; // use a concurrent map so we get more consistent iteration
+    private final Map<String, NatsSubscription> subscribers;
+    private final Map<String, NatsDispatcher> dispatchers; // use a concurrent map so we get more consistent iteration
                                                      // behavior
-    private Map<String, CompletableFuture<Message>> responses;
-    private ConcurrentLinkedDeque<CompletableFuture<Boolean>> pongQueue;
+    private final Map<String, NatsRequestCompletableFuture> responsesAwaiting;
+    private final Map<String, NatsRequestCompletableFuture> responsesRespondedTo;
```
scottf, line 77: responses changed to responsesAwaiting and responsesRespondedTo. Awaiting is the same as the original. Responded To gives a places to track after the first response

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -374,7 +377,7 @@ void tryToConnect(String serverURI, long now) {
             timeoutNanos = timeCheck(trace, end, "waiting for writer");
             this.writer.stop().get(timeoutNanos, TimeUnit.NANOSECONDS);
 
-            timeoutNanos = timeCheck(trace, end, "cleaning pong queue");
+            timeCheck(trace, end, "cleaning pong queue");
```
scottf, line 380: return value was not being used

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -387,20 +390,18 @@ void tryToConnect(String serverURI, long now) {
...
+            Callable<Object> connectTask = () -> {
+                readInitialInfo();
+                checkVersionRequirements();
+                long start = System.nanoTime();
+                upgradeToSecureIfNeeded();
+                if (trace && options.isTLSRequired()) {
+                    // If the time appears too long it might be related to
+                    // https://github.com/nats-io/nats.java#linux-platform-note
+                    timeTrace(true, "TLS upgrade took: %.3f (s)",
+                            ((double) (System.nanoTime() - start)) / 1_000_000_000.0);
```
scottf, line 402: simpler lambda form

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -545,19 +546,18 @@ void handleCommunicationIssue(Exception io) {
...
-            } else {
-                this.disconnecting = true;
-                this.exceptionDuringConnectChange = null;
-                wasConnected = (this.status == Status.CONNECTED);
-                statusChanged.signalAll();
             }
+            this.disconnecting = true;
+            this.exceptionDuringConnectChange = null;
+            wasConnected = (this.status == Status.CONNECTED);
+            statusChanged.signalAll();
```
scottf, line 560: removed unnecessary else - there is a return as part of the if block

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -944,39 +941,58 @@ String getResponseToken(String responseInbox) {
         return responseInbox.substring(len);
     }
 
-    void cleanResponses(boolean cancelIfRunning) {
+    void cleanResponses(boolean closing) {
```
scottf, line 944: change to this function takes advantage of NatsRequestCompletableFuture for better granularity and recognition of timeout

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -944,39 +941,58 @@ String getResponseToken(String responseInbox) {
...
         });
 
         for (String token : toRemove) {
-            responses.remove(token);
+            responsesAwaiting.remove(token);
+        }
+
+        toRemove.clear();
+        responsesRespondedTo.forEach((token, f) -> {
+            if (f.hasExceededTimeout()) {
```
scottf, line 973: now that there is responsesAwaiting and responsesRespondedTo, handle both in cleanup. responsesRespondedTo gets cleaned up after timeout

src/main/java/io/nats/client/support/NatsRequestCompletableFuture.java
```diff
@@ -0,0 +1,54 @@
+package io.nats.client.support;
+
+import io.nats.client.Message;
+import io.nats.client.Options;
+
+import java.time.Duration;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+
+public class NatsRequestCompletableFuture extends CompletableFuture<Message> {
```
scottf, line 10: renamed and extracted (no longer a static inner class)

src/main/java/io/nats/client/impl/NatsStatistics.java
```diff
@@ -29,6 +29,7 @@
     private AtomicLong outstandingRequests;
     private AtomicLong requestsSent;
     private AtomicLong repliesReceived;
+    private AtomicLong orphanRepliesReceived;
```
scottf, line 32: track these orphaned replies

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1018,10 +1036,10 @@ Message request(String subject, Headers headers, byte[] data, boolean utf8mode,
         boolean oldStyle = options.isOldRequestStyle();
         String responseInbox = oldStyle ? createInbox() : createResponseInbox(this.mainInbox);
         String responseToken = getResponseToken(responseInbox);
-        CompletableFuture<Message> future = new ExtendedCompletableFuture(regular);
+        NatsRequestCompletableFuture future = new NatsRequestCompletableFuture(cancelOn503, futureTimeout);
```
scottf, line 1039: Fully using NatsRequestCompletableFuture

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1045,29 +1063,28 @@ Message request(String subject, Headers headers, byte[] data, boolean utf8mode,
...
                 f.complete(msg);
             }
             statistics.incrementRepliesReceived();
-        } else if (!oldStyle && !subject.startsWith(mainInbox)) {
+        }
+        else if (responsesRespondedTo.get(key) != null) {
+                statistics.incrementRepliesReceived();
+        }
+        else if (!oldStyle && !subject.startsWith(mainInbox)) {
+            statistics.incrementOrphanRepliesReceived();
```
scottf, line 1087: tracking replies that weren't the first, counting orphans

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1045,29 +1063,28 @@ Message request(String subject, Headers headers, byte[] data, boolean utf8mode,
...
         String token = getResponseToken(subject);
-        ExtendedCompletableFuture f = (ExtendedCompletableFuture)
-                (oldStyle ? responses.remove(msg.getSID()) : responses.remove(token));
+        String key = oldStyle ? msg.getSID() : token;
+        NatsRequestCompletableFuture f = responsesAwaiting.remove(key);
         if (f != null) {
+            responsesRespondedTo.put(key, f);
             statistics.decrementOutstandingRequests();
-            if (f.regular && msg.isStatusMessage() && msg.getStatus().getCode() == 503) {
+            if (msg.isStatusMessage() && msg.getStatus().getCode() == 503 && f.isCancelOn503()) {
```
scottf, line 1075: more efficient order

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1045,29 +1063,28 @@ Message request(String subject, Headers headers, byte[] data, boolean utf8mode,
...
-    }
-
     void deliverReply(Message msg) {
         boolean oldStyle = options.isOldRequestStyle();
         String subject = msg.getSubject();
         String token = getResponseToken(subject);
-        ExtendedCompletableFuture f = (ExtendedCompletableFuture)
-                (oldStyle ? responses.remove(msg.getSID()) : responses.remove(token));
+        String key = oldStyle ? msg.getSID() : token;
+        NatsRequestCompletableFuture f = responsesAwaiting.remove(key);
```
scottf, line 1071: I need the key in multiple places, so extract a variable

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1191,13 +1208,13 @@ void sendConnect(String serverURI) throws IOException {
         int max = this.options.getMaxPingsOut();
 
         if (!isConnectedOrConnecting()) {
-            CompletableFuture<Boolean> retVal = new CompletableFuture<Boolean>();
+            CompletableFuture<Boolean> retVal = new CompletableFuture<>();
```
scottf, line 1211: Correct way to specify generic

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1354,10 +1371,11 @@ void deliverMessage(NatsMessage msg) {
...
             }
 
-        } else {
-            // Drop messages we don't have a subscriber for (could be extras on an
-            // auto-unsub for example)
         }
+//        else {
+//            // Drop messages we don't have a subscriber for (could be extras on an
+//            // auto-unsub for example)
+//        }
```
scottf, line 1378: just commented out do nothing code

src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -291,48 +291,47 @@ public PublishAck publish(Message message, PublishOptions options) throws IOExce
...
     }
 
-    private CompletableFuture<PublishAck> publishAsync(String subject, Headers headers, byte[] data, boolean utf8mode, PublishOptions options) {
+    private PublishAck publishSyncInternal(String subject, Headers headers, byte[] data, boolean utf8mode, PublishOptions options) throws IOException, JetStreamApiException {
+        Duration timeout = options == null ? requestTimeout : options.getStreamTimeout();
         Headers merged = mergePublishOptions(headers, options);
-        CompletableFuture<Message> future = conn.request(subject, merged, data, utf8mode, false);
+
+        Message resp = makeRequestResponseRequired(subject, merged, data, utf8mode, timeout, false);
+        return processPublishResponse(resp, options);
```
scottf, line 302: I was handling sync messages different than async messages, when they should have been treated the same as far as 503 handling. See processPublishResponse

src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -291,48 +291,47 @@ public PublishAck publish(Message message, PublishOptions options) throws IOExce
...
     }
 
-    private PublishAck publishSync(String subject, Headers headers, byte[] data, boolean utf8mode, PublishOptions options) throws IOException, JetStreamApiException {
-        Duration timeout = options == null ? requestTimeout : options.getStreamTimeout();
-        Headers merged = mergePublishOptions(headers, options);
-        Message resp = makeRequestResponseRequired(subject, merged, data, utf8mode, timeout);
-        return processAck(resp, options);
-    }
+    private PublishAck processPublishResponse(Message resp, PublishOptions options) throws IOException, JetStreamApiException {
+        if (resp.isStatusMessage()) {
```
scottf, line 320: aysnc was already doing this. Refactored so sync would do the same.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -754,8 +754,7 @@ private void publishInternal(String subject, String replyTo, Headers headers, by
 
         NatsMessage nm = new NatsMessage(subject, replyTo, new Headers(headers), data, utf8mode);
 
-        Connection.Status stat = this.status;
-        if ((stat == Status.RECONNECTING || stat == Status.DISCONNECTED)
+        if ((this.status == Status.RECONNECTING || this.status == Status.DISCONNECTED)
```
ColinSullivan1: I think I see why the original author did that - to reduce issues around thread safety.  Longer term, the status checks need to be protected or atomic.  At the least, status should be volatile.

src/main/java/io/nats/client/support/NatsRequestCompletableFuture.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final long SAFE_TO_CONSIDER_ORPHANED = Duration.ofMinutes(10).toMillis();
+    private static final long TIMEOUT_PADDING = Options.DEFAULT_CONNECTION_TIMEOUT.toMillis(); // currently 2 seconds
+
+    private final boolean cancelOn503;
+    private final long timeOutAfter;
+    private boolean wasCancelledClosing;
+    private boolean wasCancelledTimedOut;
+
+    public NatsRequestCompletableFuture(boolean cancelOn503, Duration timeout) {
```
ColinSullivan1, line 19: Probably needs a comment, at least for cancelOn503

src/test/java/io/nats/client/impl/RequestTests.java
```diff
@@ -139,13 +141,41 @@ public void testMultipleRequest() throws IOException, ExecutionException, Timeou
...
+    @Test
+    public void testMultipleReplies() throws Exception {
+        Options.Builder builder = new Options.Builder().turnOnAdvancedStats();
+        runInServer(builder, nc -> {
+            AtomicInteger replies = new AtomicInteger();
+            MessageHandler handler = (msg) -> { replies.incrementAndGet(); nc.publish(msg.getReplyTo(), null); };
+            Dispatcher d1 = nc.createDispatcher(handler);
+            Dispatcher d2 = nc.createDispatcher(handler);
+            Dispatcher d3 = nc.createDispatcher(handler);
+            Dispatcher d4 = nc.createDispatcher(msg -> { sleep(5000); handler.onMessage(msg); });
```
ColinSullivan1, line 153: If you wanted to you could probably reduce the amount of time waiting in the test and wait on a latch in this dispatcher to make this more deterministic - could see this becoming a flapper on stressed systems.

ColinSullivan1 (approved):
LGTM.  A few suggestions...  Longer term, I think a pass at locking would be good.

src/main/java/io/nats/client/impl/NatsConnection.java
scottf: reverted

src/test/java/io/nats/client/impl/RequestTests.java
scottf: Yeah, it's not optimal, but also it's hard to latch onto a different thread. I think if it flaps, we should actually just give it more sleep time.

PR #455, opened 2021-04-29
340 tls opentls multiple servers in url string / array

scottf:
when there were multiple servers in the options (either comma delimited string or added by builder, or array, etc.) the ssl context was not created if it was not already supplied as it was set to handle only when there was exactly 1 server

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -1337,19 +1337,21 @@ public class Options {
             
             if (servers.size() == 0) {
                 server(DEFAULT_URL);
-            } else if (servers.size() == 1) { // Allow some URI based configs
-                URI serverURI = servers.get(0);
-
-                if (TLS_PROTOCOL.equals(serverURI.getScheme()) && this.sslContext == null)
-                {
-                    try {
-                        this.sslContext = SSLContext.getDefault();
-                    } catch (NoSuchAlgorithmException e) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/nats/client/Options.java
```diff
@@ -1337,19 +1337,21 @@ public Options build() throws IllegalStateException {
...
-                if (TLS_PROTOCOL.equals(serverURI.getScheme()) && this.sslContext == null)
-                {
-                    try {
-                        this.sslContext = SSLContext.getDefault();
-                    } catch (NoSuchAlgorithmException e) {
-                        throw new IllegalStateException("Unable to create default SSL context", e);
+            }
+            else if (sslContext == null) {
+                for (URI serverURI : servers) {
+                    if (TLS_PROTOCOL.equals(serverURI.getScheme())) {
```
ColinSullivan1, line 1343: Wondering if there should be a check someplace for a mix of TLS types - native vs openSSL should throw.  Could get into the situation where some URLS are openTLSContext and others are native ssl context.

ColinSullivan1 (approved):
LGTM.

src/main/java/io/nats/client/Options.java
scottf: There is always the ability for the user to setup the sslContext themselves. I guess if this comes up we'll address it

PR #456, opened 2021-04-29
switched to jnats-server-runner library

scottf:
* switched to jnats-server-runner library
* moved some print util to make more available for general use


ColinSullivan1 (approved):
LGTM.

PR #467, opened 2021-05-12
duplicates and orphans

scottf:
The fix adds handlers to the ErrorListener that allows the user to listen for duplicate request replies and late (orphan replies)


scottf:
@ColinSullivan1 I removed the ErrorListener changes but added granularity to the statistics.

ColinSullivan1 (approved):
LGTM.

PR #468, opened 2021-05-13
js multi improvements to help re-use

scottf:
1. multi - make stats and args more accessible in case user wants to customize args or use stats for their own purposes 
2. multi - remove payload generation from timing in case they want to override
3. multi - ability to report to file and extract totals
4. auto bench - latency reporting more readable


ColinSullivan1 (approved):
LGTM.

PR #471, opened 2021-05-17
more flow control

scottf:
making sure unit test and example exercise use of both pending message and byte limits

### DIFF
--- a/src/examples/java/io/nats/examples/jetstream/NatsJsPushSubFlowControl.java
+++ b/src/examples/java/io/nats/examples/jetstream/NatsJsPushSubFlowControl.java
@@ -67,9 +67,8 @@ public class NatsJsPushSubFlowControl {
                 nc.flush(Duration.ofSeconds(5));
 
                 // ------------------------------------------------------------------------------------------
-                // Flow Control limit is set using pending limits
-                // you can mix and match pending messages with total bytes or rely on one
-                // or the other.
+                // Flow Control limit is set using pending limits. You can mix and match pending messages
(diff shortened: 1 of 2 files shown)


src/examples/java/io/nats/examples/jetstream/NatsJsPushSubFlowControl.java
```diff
@@ -65,7 +65,14 @@ public static void main(String[] args) {
...
                 nc.flush(Duration.ofSeconds(5));
-                sub.setPendingLimits(Consumer.DEFAULT_MAX_MESSAGES, 1024);
+
+                // flow control limit is set using pending limits
+                // you can mix and match pending messages with total bytes or rely on one
+                // or the other.
+                // The first version sets pending to 500 message with 1K total bytes
+                // The second version (commented out) sets pending to 1 message with 500K total bytes
+                sub.setPendingLimits(500, 1024);
+                // sub.setPendingLimits(1, 500000);
```
ColinSullivan1, line 75: imo, a pending limit of one message isn't something we'd want a user to copy.  Maybe (50, 1024*500)?

ColinSullivan1 (commented):
LGTM, would suggest changing the 2nd example before merging.

src/examples/java/io/nats/examples/jetstream/NatsJsPushSubFlowControl.java
scottf: ok I'm just going to add more comments. It's hard to reproduce flow control, extreme boundaries are the best

PR #473, opened 2021-05-18
performance improvements

scottf:


src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -337,22 +337,25 @@ public int serializedLength() {
...
-				}
-			}
-			bab.append(CRLF_BYTES);
-			serialized = bab.toByteArray();
+			serialized = appendSerialized(new ByteArrayBuilder(dataLength + NON_DATA_BYTES)).toByteArray();
 		}
 		return serialized;
 	}
 
+	public ByteArrayBuilder appendSerialized(ByteArrayBuilder bab) {
```
ColinSullivan1, line 345: I wonder if it'd further help to keep an instance of the ByteArrayBuilder in ThreadLocal to prevent creating a new object every call?  wdyt?

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -89,139 +91,125 @@ void start(Future<DataPort> dataPortFuture) {
...
-    }
-
-    synchronized void sendMessageBatch(NatsMessage msg, DataPort dataPort, NatsStatistics stats)
-            throws IOException {
-
-        int sendPosition = 0;
 
-        while (msg != null) {
-            long size = msg.getSizeInBytes();
+        // TODO ???
```
ColinSullivan1, line 95: Did you intend to leave this in?

ColinSullivan1 (approved):
LGTM.

src/main/java/io/nats/client/impl/Headers.java
ColinSullivan1: Per conversation, it looks like this is reused.

PR #491, opened 2021-06-18
Retract deliver subject

scottf:
The example demonstrating deliver subject being used for queueing does not work and is removed.


ColinSullivan1 (approved):
LGTM.

PR #494, opened 2021-06-21
DataPort -> NatsChannel Refactor

brimworks:
This refactor prepares for laying sub-transports "under" NATS. Specifically, websockets (and HTTP proxies) would all be sub-transports.

To accomplish this, I have created a new `NatsChannel` abstraction which replaces the `DataPort` abstraction so that implementations of `NatsChannel` can be layered on top of each other (using the decorator design pattern). This is similar to how Java IO layers InputStream/BufferedInputStream/Reader on top of each other.

To support this, I added an `isSecure()` method to `NatsChannel` which replaces the need for the `upgradeToSecure()` method... more specifically, if an upgrade to secure is made, and the current `NatsChannel.isSecure()` returns false, then `NatsConnection` will now create a `TLSNatsChannel` which wraps the lower-level `NatsChannel`.

This allows us to re-use the `TLSNatsChannel` for our websocket implementation, and the websocket implementation can just implement `NatsChannel`.

Decisions made:
* After talking to @scottf about possibly making `NatsChannel` async, the decision was made to NOT make the code more complicated by making it async.
* The following APIs are not backwards compatible:
  * `Options.getDataPortType()` removed
  * `Options.buildDataPort()` removed
  * Custom implementations of `DataPort` will never have the `DataPort.upgradeToSecure()` method called.
  * Custom implementations of `DataPort` will never have the `DataPort.flush()` method called.
  * Custom implementations of `DataPort.connect()` will have a "fake" `NatsConnection` passed in and the only "useful" information in that object will be the `Options`.

Open Items:
* I can't help but notice that the `NatsConnectionWriter.flushBuffer()` doesn't work (even prior to my change) as expected. I believe the "proper" way to handle flushing the queue is to enqueue a "fake" message that indicates that a flush is desired, and then wait for that message to be flushed. More specifically, the previous code was calling `SocketOutputStream.flush()` which is known to be an empty implementation: https://bugs.java.com/bugdatabase/view_bug.do;jsessionid=61e3ecc1489150ffffffffcc5ccaf3f797827?bug_id=4358695 Therefore, I did NOT include the `flush()` call in the new `NatsChannel` abstraction.
* How to handle connect_urls in the INFO message that have no protocol (in preparation for websocket support)?

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -13,22 +13,32 @@
 
 package io.nats.client;
 
+import io.nats.client.impl.AdaptDataPortToNatsChannelFactory;
 import io.nats.client.impl.DataPort;
+import io.nats.client.impl.DefaultNatsChannelFactory;
+import io.nats.client.channels.NatsChannelFactory;
 import io.nats.client.impl.SocketDataPort;
 import io.nats.client.support.SSLUtils;
 
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+
 import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.CharBuffer;
+import java.security.GeneralSecurityException;
 import java.security.NoSuchAlgorithmException;
 import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
 
 import static io.nats.client.support.NatsConstants.*;
+import static io.nats.client.support.SneakyThrow.sneakyThrow;
+import static io.nats.client.support.URIUtils.withDefaultPort;
 
 /**
  * The Options class specifies the connection options for a new NATs connection, including the default options.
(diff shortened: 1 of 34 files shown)


src/main/java/io/nats/client/Options.java
```diff
@@ -225,8 +229,16 @@
...
+     * 
+     * Replaced with {@link Builder#natsChannelFactory(NatsChannelFactory) natsChannelFactory}.
      */
+    @Deprecated
     public static final String PROP_DATA_PORT_TYPE = PFX + "dataport.type";
+    /**
+     * Property used to configure a builder from a Properties object. {@value}, see
+     * {@link Builder#natsChannelFactory(NatsChannelFactory) natsChannelFactory}.
+     */
+    public static final String PROP_NATS_CHANNEL_FACTORY = PFX + "channel.factory";
```
scottf, line 241: Shouldn't the factory be fixed and it uses some other property to figure out what it generates? Maybe an example of what this value is

brimworks:
Thanks for the nice chat on the phone @scottf , just to recap:

* Move the channels into a separate channels package.
* Clean-up any PR comments.
* Rebase my websocket PR on top of this PR.
* No async for now, although I'll put up a separate PR just for demo purposes.

Thanks!
-Brian

src/main/java/io/nats/client/impl/NatsChannel.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Shutdown the reader side of the channel.
+     */
+    void shutdownInput() throws IOException;
+
+    /**
+     * Flush the writer side of the channel.
+     */
+    void flushOutput() throws IOException;
```
brimworks, line 26: When I investigated what flushOutput() does, it seems to do nothing. Specifically, it seems that the flushOutput() should be enqueuing a message with a CountDownLatch() which is counted down when the message is processed and the flushOutput() should await() on that (which has nothing to do with the DataPort or NatsChannel).

src/main/java/io/nats/client/channels/NatsChannel.java
```diff
@@ -0,0 +1,26 @@
...
+     */
+    boolean isSecure();
+
+    /**
+     * Shutdown the reader side of the channel.
+     * 
+     * @throws IOException if an IO error occurs.
+     */
+    void shutdownInput() throws IOException;
+}
```
brimworks, line 26: I think I need one more API to support transforming the `connect_urls` of the INFO message so they contain the appropriate protocol. Otherwise, when reconnecting to these other servers, we may accidentally connect over the wrong protocol.

brimworks:
Scrapping the nio based solution. It ends up being 4x more code changes.

PR #498, opened 2021-06-22
Work in progress, demonstrates using Asynchronous version of NatsChannel

brimworks:


brimworks:
It sounds like async is not planned in the near future, so closing this.

PR #499, opened 2021-06-23
cluster name required in test config

scottf:
The config for a cluster needs a name. This test was somehow working, but was a flapper. Not sure why it worked sometimes without a cluster name, but should always work now.


brimworks (approved):
Thanks for fixing the flapper 👍

PR #500, opened 2021-06-25
Here is a bug in example code

dragonflysun2019:
public static ExampleArgs expectSubjectQueueAndMsgCount(String[] args, String usageString) {
        ExampleArgs ea = new ExampleArgs(args, ExampleArgs.Trail.QUEUE_AND_COUNT, usageString);
        if (ea.containedUnknown || ea.msgCount < 1) {
            usage(usageString);
        }
        return ea;
    }


dragonflysun2019:
Here is a bug in the code example

scottf:
@dragonflysun2019 Is this a bug or not?

PR #502, opened 2021-06-28
Add Websocket support using NatsChannel abstraction, resolves #399

### LINKED ISSUE #399: Websocket Support - Java

brimworks:
This replaces this PR: https://github.com/nats-io/nats.java/pull/426


brimworks:
Scrapping the nio based solution. It ends up being 4x more code changes.

PR #504, opened 2021-06-29
Fix "flapping" ErrorListenerTests.testDiscardedMessageFastProducer test

brimworks:
so it is reliable. The root cause of this flip is a race condition for
when the poisonPill of MessageQueue is enqueued (by the pause() method)
and if the NatsConnectionWriter.run() loop terminates before or after
processing of that poisonPill.


build.gradle
```diff
@@ -41,6 +41,7 @@ repositories {
 dependencies {
     implementation 'net.i2p.crypto:eddsa:0.3.0'
     testImplementation 'org.junit.jupiter:junit-jupiter:5.7.0'
+    testImplementation 'org.hamcrest:hamcrest:2.2'
```
scottf, line 44: can you do this without hamcrest. There is a strict policy to not use any libraries, even for test

scottf (changes requested):

build.gradle
brimworks: Ya no problem. I guess I figured that hamcrest is on the same "level" as junit in terms of ubiquitous usage.

scottf (approved):

PR #512, opened 2021-07-07
use internal error codes

scottf:
A unit test failed on the nightly because the server changed the text of an error. This fix is to use the new err_code field supplied in the json of api error response, instead of using the text.


matthiashanel (approved):
LGTM

PR #521, opened 2021-08-11
max_ack_pending set by server if not provided

scottf:
during subscribe, we used to set max_ack_pending if it was not supplied, but we set it to pendingMessageLimit which defaults to 64k. The server default for max_ack_pending is 20K. Seems like the right behavior to let the server set this value.


ColinSullivan1 (approved):
LGTM - please coordinate with @wallyqs and @aricart for client parity.  Thanks!

PR #522, opened 2021-08-12
Fix issue where Timeout exception is thrown if error is returned by nats-server before a PONG

brimworks:
The test case demonstrates the scenario, but here is a quick explanation:

1. The `Options.connectionTimeout()` must be set.
2. The reconnect logic must be triggered.
3. The nats-server MUST return successfully reconnect and do the INFO/CONNECT request/reply.
4. The nats-server MUST return an `-ERR` response BEFORE the `PONG`. In real-life this would likely be some sort of auth error, but maybe others (like if `nats-server` is in lame-duck mode?)

When this happens, the reconnect code which waits for a PONG response needs to be interrupted when the connect error gets set.


scottf:
Am I reading this right, it looks like you just exchanging one exception for another.

brimworks:
I am not at my computer, but as I recall, we get a different exception AND an artificial delay that is not necessary. This delay is problematic since we need a high connect timeout when a device at the edge is on a congested network. The artificially large delay then occurs almost every time the JWT token expires, and during this time the device appears unreachable from NATS.

scottf:
Replaced by recent work on reconnect

PR #528, opened 2021-08-21
Parity Issue 44 add domain to PubAck

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/44

backward compatible with server v2.3.4


src/main/java/io/nats/client/api/PublishAck.java
```diff
@@ -36,6 +37,7 @@ public PublishAck(Message msg) throws IOException, JetStreamApiException {
         if (stream == null) {
             throw new IOException("Invalid JetStream ack.");
         }
+        domain = JsonUtils.readString(json, DOMAIN_RE, null);
```
kozlovic, line 40: I assume that your utils will not throw an exception if the field is not found, right?

kozlovic (approved):
LGTM assuming domain not found is not "fatal".

src/main/java/io/nats/client/api/PublishAck.java
scottf: yeah, readString handles that

PR #529, opened 2021-08-21
Parity Issue 49 ack new tokens

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/49

backward compatible with server v2.3.4

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsJetStreamMetaData.java
+++ b/src/main/java/io/nats/client/impl/NatsJetStreamMetaData.java
@@ -17,6 +17,8 @@ import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Jetstream meta data about a message, when applicable.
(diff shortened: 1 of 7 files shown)


src/test/java/io/nats/client/impl/JetStreamManagementTests.java
```diff
@@ -400,6 +400,11 @@ private void assertInvalidConsumerUpdate(JetStreamManagement jsm, ConsumerConfig
         // 10013 consumer name already in use
         // 10105 consumer already exists and is still active
         assertTrue(e.getApiErrorCode() == 10013 || e.getApiErrorCode() == 10105);
+        assertEquals(400, e.getErrorCode());
+        assertEquals("consumer name already in use", e.getErrorDescription());
+        // \/ This is just a reminder that the server changed \/
+        // assertEquals(500, e.getErrorCode());
+        // assertEquals("consumer already exists", e.getErrorDescription());
```
scottf, line 407: Testing server breaking change Error Message Code and Text Change

src/main/java/io/nats/client/impl/NatsJetStreamMetaData.java
```diff
@@ -25,53 +25,108 @@
...
+    private final long pending;
+    private final String token;
 
     @Override
     public String toString() {
         return "NatsJetStreamMetaData{" +
-                "stream='" + stream + '\'' +
+                "prefix='" + prefix + '\'' +
+                "domain='" + domain + '\'' +
+                ", accountHash='" + accountHash + '\'' +
```
kozlovic, line 45: If this is somehow public, then `accountHash` and `token` don't need to be exposed. This is just internals needing to properly route the message back. No need to expose those to the user.

src/main/java/io/nats/client/impl/NatsJetStreamMetaData.java
```diff
@@ -25,53 +25,108 @@
...
+     */
+    public String getDomain() {
+        return domain;
+    }
+
+    /**
+     * Get the hash of the account. Might be null
+     * @return the hash
+     */
+    public String getAccountHash() {
```
kozlovic, line 128: See previous comment.

src/test/java/io/nats/client/impl/JetStreamManagementTests.java
kozlovic: Since it looks like you support error code, you should not have to check error strings. That is the point of the error codes is that we can change error strings (fix typos, think internationalization, etc..).

kozlovic (changes requested):
Don't make account hash and token public. If you need to have field/accessors for your testing, ok, but don't make those public then.

src/test/java/io/nats/client/impl/JetStreamManagementTests.java
scottf: This was just a unit test, but it's possible that a user had code like this before the release of api error codes. Also, since we cannot be sure of the version of the server used by someone, the code should be backward compatible. But this server change in specific is technically a breaking change.

src/main/java/io/nats/client/impl/NatsJetStreamMetaData.java
scottf: @kozlovic We should capture this in some ADR. Is there a security reason these should not be public?

kozlovic: It is: https://github.com/nats-io/nats-architecture-and-design/pull/42/files#diff-f52a7740d240a96a07d8e7f9a40d3eb460cee1176914ea5ae1d76175f0d140bdR172.

There is no need for the user to know those, they are simply for routing. Making them public would just add to the "breaking changes" issues if one day we want to alter those.

Also take note regarding the token count check moving forward: https://github.com/nats-io/nats-architecture-and-design/pull/42/files#diff-f52a7740d240a96a07d8e7f9a40d3eb460cee1176914ea5ae1d76175f0d140bdR174

src/test/java/io/nats/client/impl/JetStreamManagementTests.java
kozlovic: I disagree. If your library had error codes, then if user decides to check error string, they do so at their own risk and peril. As I said, the very point of having error codes is that *these* should not changes, the error text should be able to be changed (we have fixed typos for instance).

src/main/java/io/nats/client/impl/NatsJetStreamMetaData.java
scottf: ok will fix

scottf: Fixed.
1. don't expose account hash and token publicly
2. allow more than 12 tokens for future compatibility

kozlovic (approved):
LGTM

PR #531, opened 2021-08-21
document pull primitives as advanced

scottf:
Documentation valid since v2.3.4


src/main/java/io/nats/client/JetStreamSubscription.java
```diff
@@ -146,4 +170,4 @@
      * @throws JetStreamApiException the request had an error related to the data
      */
     ConsumerInfo getConsumerInfo() throws IOException, JetStreamApiException;
-}
+}
```
kozlovic, line 173: Not sure if it matters, but there is no new line here.

kozlovic (approved):
LGTM

PR #533, opened 2021-08-22
Parity Issue 33 Delivery Policy Last By Subject

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/33

Since server v2.3.4


src/main/java/io/nats/client/api/DeliverPolicy.java
```diff
@@ -24,9 +24,10 @@
     Last("last"),
     New("new"),
     ByStartSequence("by_start_sequence"),
-    ByStartTime("by_start_time");
+    ByStartTime("by_start_time"),
+    LastPerSubject("last_per_subject");
```
kozlovic, line 28: Not enough familiar with the Java lib: is that enough for the marshal/unmarshal of this new policy?

scottf: yeah, when going to json, the to string returns policy string (the quoted value). When reading from json, we compare to the policy string.

ColinSullivan1 (approved):
LGTM.

PR #534, opened 2021-08-22
Parity Issue 32 Stream Configuration description

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/32

Since server v2.3.4


kozlovic (approved):
LGTM

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -366,7 +366,7 @@ public Builder(ConsumerConfiguration cc) {
          * @return the builder
          */
         public Builder description(String description) {
-            this.description = validateMaxLength(description, 4096, false, "Description");
+            this.description = validateDescription(description, false);
```
aricart, line 369: I would not validate the description - server allowed length etc should be left to the server.

src/main/java/io/nats/client/support/Validator.java
```diff
@@ -78,6 +78,10 @@ public static String validateMaxLength(String s, int maxLength, boolean required
         });
     }
 
+    public static String validateDescription(String description, boolean required) {
```
aricart, line 81: I would remove this

aricart (changes requested):
Validating the length for the description is adding some external check that will be validated anyway. Rely on the server to accept the description as provided.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
scottf: @ripienaar @kozlovic @ColinSullivan1 @wallyqs Can we get a consensus on whether the client should pre-validate the description length. I wonder if we should build an adr (or add and re-purpose the naming conventions doc ) on what things should be client validated and what should be left for the server.

kozlovic: Agree with Alberto that client should not validate this. Server will do. One less thing to do in the client that also possibly would lead to "breaking changes" if server wanted to lower (or increase) this limit.

scottf: ok. validations removed

aricart (approved):
LGTM

PR #541, opened 2021-09-12
jwt stamping

scottf:
https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-14.md
https://docs.nats.io/developing-with-nats/tutorials/jwt#stamping-jwt-in-languages-other-than-go


src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,110 @@
...
+            "------END USER NKEY SEED------\n" +
+            "\n" +
+            "*************************************************************";
+
+    public static String issueUserCreds(String accSeed, String accId, String userSeed, String userKeyPub) throws GeneralSecurityException, IOException {
+        String jwt = issueUserJWT(accSeed, accId, userKeyPub);
+        return String.format(NATS_USER_JWT_FORMAT, jwt, userSeed);
+    }
+
+    public static String issueUserJWT(String accSeed, String accId, String userKeyPub) throws GeneralSecurityException, IOException {
```
matthiashanel, line 65: This is missing expiration and tags. see:
https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-14.md

matthiashanel (changes requested):
some of the fields are missing

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+        String name;
+        Nats nats;
+        String sub;
+
+        @Override
+        public String toJson() {
+            StringBuilder sb = beginJson();
+            if (exp != null && !exp.isZero() && !exp.isNegative()) {
+                long unixSeconds = exp.toMillis() / 1000;
+                JsonUtils.addField(sb, "exp", unixSeconds);
```
brimworks, line 59: This needs to be `iat + unixSeconds`... since the duration represents how long the token should be valid relative to the issued at time.

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+            "    NKEY Seed printed below can be used to sign and prove identity.\n" +
+            "    NKEYs are sensitive and should be treated as secrets.\n" +
+            "\n" +
+            "-----BEGIN USER NKEY SEED-----\n" +
+            "%s\n" +
+            "------END USER NKEY SEED------\n" +
+            "\n" +
+            "*************************************************************";
+
+    public static String issueUserCreds(String accountSeed, String issuerAccount, String userSeed, String userKeyPub, String optionalName, Duration optionalExpiration, String... optionalTags) throws GeneralSecurityException, IOException {
```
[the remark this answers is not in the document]
brimworks: Agreed. It would be more clear if this took the `NKey accountSigningKey` and `NKey userKey`, and simply avoid the `String accountSeed` variant. In fact, having sensitive credentials as `String` type is not recommended:

https://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords

If you feel this is really necessary, passing as `char[]` is preferred (although, it is trivial for users to call `NKey.fromSeed(char[])` API themselves, so I don't think it is buying us anything, and only adds to the API surface area to maintain.

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+        String jwt = issueUserJWT(accountSeed, issuerAccount, userKeyPub, optionalName, optionalExpiration, optionalTags);
+        return String.format(NATS_USER_JWT_FORMAT, jwt, userSeed);
```
brimworks, line 91: Since this is including the `userSeed`, we should be returning `char[]` rather than `String`, since this is a security best practice. Note that we should avoid any intermediate `String`s in that process. A simple way to do this is to allocate a string of the same length as the userSeed, do the `String.format()` and then convert this to a `char[]`, and then overwrite the char array offsets with the actual user seed.

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+public class JwtUtils {
+    static class Nats implements JsonSerializable {
+        String issuerAccount;
+        String[] tags;
+
+        @Override
+        public String toJson() {
+            StringBuilder sb = beginJson();
+            JsonUtils.addField(sb, "issuer_account", issuerAccount);
+            JsonUtils.addStrings(sb, "tags", Arrays.asList(tags));
```
brimworks, line 38: Note that all of these `JsonUtils` are not properly handling JSON string encoding for characters that need to be escaped. For example `"`, `\` or any control character such as `\n`. I would recommend asserting that all `String` inputs to APIs validate that these special characters are not found... or properly encode strings. Without this, a tag with a Windows fileshare (for example) would cause an invalid JWT to be minted and cause much confusion since no error would occur until the JWT was actually used.

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+
+        // compute the signature off of header + body (. included on purpose)
+        byte[] sig = (ENCODED_CLAIM_HEADER + "." + encBody).getBytes(StandardCharsets.UTF_8);
+        String encSig = ToBase64Url(accountSigningKey.sign(sig));
+
+        // append signature to header and body and return it
+        return ENCODED_CLAIM_HEADER + "." + encBody + "." + encSig;
+    }
+
+    public static String ToBase64Url(byte[] input) {
```
brimworks, line 138: I'd suggest leaving these as private and use the more standard `toBase64Url` (lower-case first) convention.

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+
+        claim.jti = new String(NKey.base32Encode(encoded));
+        claimJson = claim.toJson();
+
+        // all three components (header/body/signature) are base64url encoded
+        String encBody = ToBase64Url(claimJson);
+
+        // compute the signature off of header + body (. included on purpose)
+        byte[] sig = (ENCODED_CLAIM_HEADER + "." + encBody).getBytes(StandardCharsets.UTF_8);
+        String encSig = ToBase64Url(accountSigningKey.sign(sig));
```
brimworks, line 132: I have similar JWT issuing code, but the signature is computed off the `encBody` only and does NOT include the ENCODED_CLAIM_HEADER or dot. Did something change about the spec recently?

More specifically, I thought this is what the code needed to be:

```java
String encSig = ToBase64Url(accountSigningKey.sign(encBody));
```

...but I also use the `ed25519` "alg" in the JWT header, and NOT `ed25519-nkey`. So, perhaps the spec changed between these two algorithms?

brimworks (changes requested):
I think there are a few things that need to be addressed before this can be merged:
* `exp` field needs to be `iat` time plus the duration when JSON serializing.
* We need to validate that an invalid JSON is not generated for String inputs.
* We should be using `char[]` and NOT `String` when dealing with sensitive credentials.

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: I see that https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-14.md explicitly mentions this new way of computing the signature, so I'm assuming this is just the design of the new `ed25519-nkey` algorithm.

scottf:
> I think there are a few things that need to be addressed before this can be merged:
> 
> * `exp` field needs to be `iat` time plus the duration when JSON serializing.
> * We need to validate that an invalid JSON is not generated for String inputs.
> * We should be using `char[]` and NOT `String` when dealing with sensitive credentials.

- `exp` yes needs to be fixed
- invalid json, the json util is dumb. This should be fixed in the util
- char, I've seen this before and I don't really get why it's more secure. In fact I think I debunked this once already, but I could be convinced

src/main/java/io/nats/client/support/JwtUtils.java
scottf: yeah the json utils assumed data from the server or from the client, so this type of checking was unnecessary. Not sure how to address this, but it needs addressing

src/main/java/io/nats/client/NKey.java
```diff
@@ -225,7 +225,7 @@ static int crc16(byte[] bytes) {
         }
     }
 
-    static char[] base32Encode(final byte[] bytes) {
+    public static char[] base32Encode(final byte[] bytes) {
```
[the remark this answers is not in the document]
scottf: I'm okay with moving this to a support class

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+        String jwt = issueUserJWT(accountSigningKey, issuerAccount, userKeyPub, optionalName, optionalExpiration, optionalTags);
+        return String.format(NATS_USER_JWT_FORMAT, jwt, userSeed);
+    }
+
+    public static String issueUserJWT(String accountSeed, String issuerAccount, String userKeyPub, String optionalName, Duration optionalExpiration, String... optionalTags) throws GeneralSecurityException, IOException {
+        NKey accountSigningKey = NKey.fromSeed(accountSeed.toCharArray());
+        return issueUserJWT(accountSigningKey, issuerAccount, userKeyPub, optionalName, optionalExpiration, optionalTags);
+    }
+
+    public static String issueUserJWT(NKey accountSigningKey, String issuerAccount, String userKeyPub, String optionalName, Duration optionalExpiration, String... optionalTags) throws GeneralSecurityException, IOException {
```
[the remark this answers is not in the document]
scottf: I think b/c of it's never been used in the client so far, that we  should avoid using optional. I think I would have been better served by java doc and not naming them that. If you like, I have an entire monologue on demonstrating why "optional is better than null" is bs.

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+            JsonUtils.addField(sb, "sub", sub);
+
+            return endJson(sb).toString();
+        }
+    }
+
+    static final String ENCODED_CLAIM_HEADER =
+            ToBase64Url("{\"typ\":\"JWT\", \"alg\":\"ed25519-nkey\"}");
+
+    static final String NATS_USER_JWT_FORMAT = "-----BEGIN NATS USER JWT-----\n" +
```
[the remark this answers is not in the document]
scottf: will fix

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+    public static String issueUserCreds(NKey accountSigningKey, String issuerAccount, String userSeed, String userKeyPub, String optionalName, Duration optionalExpiration, String... optionalTags) throws GeneralSecurityException, IOException {
+        String jwt = issueUserJWT(accountSigningKey, issuerAccount, userKeyPub, optionalName, optionalExpiration, optionalTags);
+        return String.format(NATS_USER_JWT_FORMAT, jwt, userSeed);
+    }
+
+    public static String issueUserJWT(String accountSeed, String issuerAccount, String userKeyPub, String optionalName, Duration optionalExpiration, String... optionalTags) throws GeneralSecurityException, IOException {
+        NKey accountSigningKey = NKey.fromSeed(accountSeed.toCharArray());
+        return issueUserJWT(accountSigningKey, issuerAccount, userKeyPub, optionalName, optionalExpiration, optionalTags);
+    }
+
```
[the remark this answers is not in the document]
scottf: We don't need no stinkin' documentation! Uhh.... I mean yes, will do.

src/main/java/io/nats/client/support/JwtUtils.java
scottf: @brimworks can you sort this out with @matthiashanel

src/main/java/io/nats/client/support/JwtUtils.java
scottf: I've been in .NET too much lately, yes please fix. Also maybe belongs in the utility with the base32 function.

src/main/java/io/nats/client/support/JwtUtils.java
scottf: defer to Matthias

src/main/java/io/nats/client/support/JwtUtils.java
matthiashanel: agreed on `iat + unixSeconds`
from the adr:
 // expiration (when specified) time in unix seconds, derived from provided duration.

src/main/java/io/nats/client/support/JwtUtils.java
matthiashanel: My java is rusty, but I'm buying the wiping argument. (can't manually overwrite a string to erase the content)

As far as I see it, either move to []byte or get rid of the `issueUserCreds` calls, making sure to make `NATS_USER_JWT_FORMAT` public 
@brimworks what do you think about dripping `issueUserCreds` and providing the printf pattern instead?

perhaps call out the printf pattern in the doc to `issueUserJWT`

src/main/java/io/nats/client/support/JwtUtils.java
matthiashanel: Does java have no native object to json serializer?

src/main/java/io/nats/client/support/JwtUtils.java
```diff
@@ -0,0 +1,145 @@
...
+    public static String issueUserJWT(NKey accountSigningKey, String issuerAccount, String userKeyPub, String optionalName, Duration optionalExpiration, String... optionalTags) throws GeneralSecurityException, IOException {
+        String accSigningKeyPub = new String(accountSigningKey.getPublicKey());
+
+        Claim claim = new Claim();
```
matthiashanel, line 107: Should these raise an exception of sorts? 
* signingKey, is a mandatory account nkey pair to sign the generated jwt.
 * accountId, is a mandatory public account nkey. Will return error when not set or not account nkey.
 * publicUserKey, is a mandatory public user nkey. Will return error when not set or not user nkey.

matthiashanel (changes requested):

src/main/java/io/nats/client/support/JwtUtils.java
scottf: java does not and we've been asked to avoid 3rd party libraries

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: Using char[] rather thing String is a security best practice. The problem with String is that the garbage collector works differently in regards to Strings, and strings may even be "intern"ed so they get reused. Anyways, if you all feel strongly about using String, that is fine, I'm just bringing this up as a Java best-practice.

src/main/java/io/nats/client/support/JwtUtils.java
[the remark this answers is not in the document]
brimworks: >  I have an entire monologue on demonstrating why "optional is better than null" is bs.

I'd love to hear it. I personally find Optional to be a great tool. Especially if used in combination with static code analysis.

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: Sounds like we do need the `iat + unixSeconds`.

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: Based on conversation off-line, we can easily fix this by modifying `JsonUtils` so these characters are properly encoded.

src/main/java/io/nats/client/support/JwtUtils.java
matthiashanel: @brimworks What I mean is basically avoid the entire problem by not having `issueUserCreds` and in the documentation of `issueUserJWT` point at `NATS_USER_JWT_FORMAT` instead.

brimworks: Yeah. I do like that idea. @scottf do you have an opinion?

scottf: I'm fine with whatever you guys want. If It satisfies Matthias's requirements then go for it. We can always add it in later, it's harder to take it back

src/main/java/io/nats/client/support/JwtUtils.java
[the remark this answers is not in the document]
brimworks: I just removed this method to avoid the char[] vs String issue.

src/main/java/io/nats/client/support/JwtUtils.java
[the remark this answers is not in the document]
brimworks: I've changed this to public.

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: done

src/main/java/io/nats/client/NKey.java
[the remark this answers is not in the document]
brimworks: Moved to RandomUtils support class... since it seems to contain an assortment of utilities such as bytesToLong() and nextLong(Random, long). I suspect that class was originally intended for actual random number generation type of stuff, but bytesToLong() has nothing to do with random generation... so it seems reasonable.

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: `issueUserCreds` was removed and the `NATS_USER_JWT_FORMAT` is public along with documentation on how to use it to create credentials.

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: `JsonUtils` now properly escapes string characters that need escaping per the spec.

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: Update to the more standard camelCase.

src/main/java/io/nats/client/support/JwtUtils.java
brimworks: Yes. All of these are now validated and have unit tests for the validation.

brimworks (approved):
I've fixed on the comments I had in regards to this PR, so marking it as approved.

src/test/java/io/nats/client/support/JwtUtilsTests.java
```diff
@@ -0,0 +1,121 @@
...
+                "nats": {
+                    "issuer_account": "ACXZRALIL22WRETDRXYKOYDB7XC3E7MBSVUSUMFACO6OM5VPRNFMOOO6",
+                    "type": "user",
+                    "version": 2
+                },
+                "sub": "UA6KOMQ67XOE3FHE37W4OXADVXVYISBNLTBUT2LSY5VFKAIJ7CRDR2RZ"
+            }
+         */
+        assertEquals("-----BEGIN NATS USER JWT-----\n" +
+            "eyJ0eXAiOiJKV1QiLCAiYWxnIjoiZWQyNTUxOS1ua2V5In0.eyJpYXQiOjE2MzMwNDMzNzgsImp0aSI6Ilg0VFdENEkzWUdDQVo0UFJLVUxPT1VBQ1I1VzVBSUtTTTZNWFZQSlpCTFhIWlpNTkZaQVEiLCJpc3MiOiJBRFE0QllNNUtJQ1I1T1hEU1AzUzNXVko1Q1lFT1JHUUtUNzJTVlJGMlpEVkE3TFRGS01DSVBHWSIsIm5hbWUiOiJVQTZLT01RNjdYT0UzRkhFMzdXNE9YQURWWFZZSVNCTkxUQlVUMkxTWTVWRktBSUo3Q1JEUjJSWiIsIm5hdHMiOnsiaXNzdWVyX2FjY291bnQiOiJBQ1haUkFMSUwyMldSRVREUlhZS09ZREI3WEMzRTdNQlNWVVNVTUZBQ082T001VlBSTkZNT09PNiIsInR5cGUiOiJ1c2VyIiwidmVyc2lvbiI6Mn0sInN1YiI6IlVBNktPTVE2N1hPRTNGSEUzN1c0T1hBRFZYVllJU0JOTFRCVVQyTFNZNVZGS0FJSjdDUkRSMlJaIn0.CUo-0aCxS1DvPpvZ3cZMXmLlQRMKYRfQjjXSFiCkmRrIJKX9crdkcIBuoYIgfczt5oTEDyR1ldS_ucipxCnoAQ\n" +
```
matthiashanel, line 38: just to be sure, the `\n` at the end is an artefact of the test, not the jwt.
I stored this string without the `\n` and nsc is happy. 
```
nsc describe jwt -f x
+---------------------------------------------------------------------------+
|                                   User                                    |
+----------------+----------------------------------------------------------+
| Name           | UA6KOMQ67XOE3FHE37W4OXADVXVYISBNLTBUT2LSY5VFKAIJ7CRDR2RZ |
| User ID        | UA6KOMQ67XOE3FHE37W4OXADVXVYISBNLTBUT2LSY5VFKAIJ7CRDR2RZ |
| Issuer ID      | ADQ4BYM5KICR5OXDSP3S3WVJ5CYEORGQKT72SVRF2ZDVA7LTFKMCIPGY |
| Issuer Account | ACXZRALIL22WRETDRXYKOYDB7XC3E7MBSVUSUMFACO6OM5VPRNFMOOO6 |
| Issued         | 2021-09-30 23:09:38 UTC                                  |
| Expires        |                                                          |
| Issuer Scoped  | Yes                                                      |
+----------------+----------------------------------------------------------+
```

matthiashanel (commented):
I think this looks good. I'll leave the final say on this with @scottf

matthiashanel (approved):
LGTM

PR #549, opened 2021-09-21
handle domain and simplify js message determination

scottf:
1. now able to take a domain instead of a prefix. domains generate prefix of `$JS.<domain>.API`
2. remove the JsPrefixManager - it's unnecessary. The prefix is only for outgoing publishes and can be confined to the JeStreamOptions. Determination of incoming messages being JetStream messages is simple, the reply-to starts with `$JS.ACK`


jnmoyne (approved):
Looks good

PR #563, opened 2021-11-12
Ordered consumer

scottf:


src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -298,17 +298,102 @@ public long getMaxPullWaiting() {
...
-    public boolean getHeadersOnly() {
+    public boolean isHeadersOnly() {
         return headersOnly != null && headersOnly;
     }
 
+    /**
+     * Gets whether deliver policy of this consumer configuration was set or left unset
+     * @return true if the policy was set, false if the policy was not set
+     */
+    public boolean deliverPolicyWasSet() {
```
ColinSullivan1, line 309: Not objecting per se, but do all of these need to be public?  Anything public must be supported until at least the next major version.

scottf: It's a packaging problem. Not sure how to get around it all.

PR #564, opened 2021-11-24
Ordered consumer

scottf:


src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -320,74 +342,141 @@ else if (so.isBind()) {
...
         return sub;
     }
 
+    static class ConsumerConfigurationComparer extends ConsumerConfiguration {
+        public ConsumerConfigurationComparer(ConsumerConfiguration cc) {
+            super(cc);
+        }
+
+        public boolean wouldBeChangeTo(ConsumerConfiguration serverCc) {
+            ConsumerConfigurationComparer serverCcc = new ConsumerConfigurationComparer(serverCc);
```
matthiashanel, line 414: implement equals and tohash instead?

src/main/java/io/nats/client/impl/NatsJetStreamImplBase.java
```diff
@@ -47,7 +47,7 @@ protected ConsumerInfo _getConsumerInfo(String streamName, String consumer) thro
         return new ConsumerInfo(resp).throwOnHasError();
     }
 
-    protected ConsumerInfo _createConsumer(String streamName, ConsumerConfiguration config) throws IOException, JetStreamApiException {
+    ConsumerInfo _createConsumer(String streamName, ConsumerConfiguration config) throws IOException, JetStreamApiException {
```
matthiashanel, line 50: private?

src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
```diff
@@ -26,10 +26,22 @@
...
-        super(sid, subject, null, connection, null, asm, js, stream, consumer, null);
+    NatsJetStreamPullSubscription(String sid, String subject,
+                                  NatsConnection connection,
+                                  NatsJetStream js,
+                                  String stream, String consumer,
+                                  MessageManager statusManager) {
+        super(sid, subject, null, connection, null, js, stream, consumer, statusManager);
     }
 
+    /*
```
matthiashanel, line 37: delete?

matthiashanel (approved):
First time I've looked at this code and have found nothing obvious.
Just minor comments.

LGTM

src/main/java/io/nats/client/impl/NatsJetStream.java
scottf: It's not the same as equals, because equals would understand "not" supplied to be a default value. I don't want this to be public at all, but it's the only way I can test it.

src/main/java/io/nats/client/impl/NatsDispatcher.java
```diff
@@ -88,20 +88,21 @@ public void run() {
...
+                    if (handler == null) {
+                        handler = defaultHandler;
                     }
 
-                    try {
-                        currentHandler.onMessage(msg);
-                    } catch (Exception exp) {
-                        this.connection.processException(exp);
-                    }
+                    if (handler != null) {
```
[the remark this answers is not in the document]
scottf: nope. Will fix with KV stuff

src/main/java/io/nats/client/impl/NatsJetStream.java
scottf: This is not the same as equals, which would understand values not set to be equivalent to a default value. This wants to check if the user specifically set values intended as changes to an existing consumer

src/main/java/io/nats/client/impl/NatsJetStreamImplBase.java
scottf: I need access within the package to this.

src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
scottf: will fix in the kv pr

tbeets:
After the PR merge, but for the record reviewed 1-1 with Scott and LGTM

PR #568, opened 2021-12-07
ack tweaks adr 15 update

scottf:
The changes address handling of user calling a version of ack more than once. I was originally handling this partially in the message (the tracking part) but then it was up to the auto ack code to check that tracking. But the adr now documents what all cases of multiple acks on a message should do, so I moved all the code into the message so it will apply to user calls in addition to internal code.


src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -448,11 +448,7 @@ public AsyncMessageHandler(MessageHandler userHandler, boolean isAutoAck, Consum
             handlers = new ArrayList<>();
             handlers.add(userHandler);
             if (isAutoAck && fnlServerCC.getAckPolicy() != AckPolicy.None) {
-                handlers.add(m -> {
-                    if (m.lastAck() == null || m.lastAck() == AckType.AckProgress) {
-                        m.ack();
-                    }
-                });
+                handlers.add(Message::ack);
```
scottf, line 451: the intelligence is in the ack call now

aricart (approved):
LGTM

src/test/java/io/nats/client/impl/JetStreamPushAsyncTests.java
```diff
@@ -227,24 +227,39 @@ public void testDontAutoAckSituations() throws Exception {
             // Create our JetStream context.
             JetStream js = nc.jetStream();
 
+            int pubCount = 5;
```
scottf, line 230: updated this unit test for better ack condition testing

PR #579, opened 2021-12-23
kv subject prefix changes to properly support accounts

scottf:


src/main/java/io/nats/client/Options.java
```diff
@@ -1161,13 +1161,13 @@ public Builder reconnectBufferSize(long size) {
          * 
          * If the user and password are set in the server URL, they will override these values. However, in a clustering situation,
          * these values can be used as a fallback.
+         *
+         * use the char[] version instead for better security
```
scottf, line 1165: Changed from deprecated to just a better comment.

src/main/java/io/nats/client/support/NatsKeyValueUtil.java
```diff
@@ -55,8 +55,12 @@ public static String toStreamSubject(String bucketName) {
...
     }
 
-    public static String toKeySubject(JetStreamOptions jso, String bucketName, String key) {
-        return (jso.isDefaultPrefix() ? "" : jso.getPrefix()) + KV_SUBJECT_PREFIX + bucketName + DOT + key;
+    public static String toKeySubject(String bucketName, String key) {
+        return  KV_SUBJECT_PREFIX + bucketName + DOT + key;
+    }
+
+    public static String toKeySubjectConsiderPrefix(JetStreamOptions jso, String bucketName, String key) {
+        return (jso.isDefaultPrefix() ? "" : jso.getPrefix()) + toKeySubject(bucketName, key);
```
ripienaar, line 63: Not sure this is correct, though my Java is super rusty so I might be miss reading.  I was trying to add the KV prefix to go and looked at how you did it here.

Here we are using the JetStream Context prefix to also prefix the KV bucket, while this work in very specifically constructed cases this is not correct in general as what we're supposed to do is take a arbitrary prefix, not assume its always going to match the JS prefix?

The example in ADR-19 shows how the server might support things in the future, but today I think we need to take arbitrary prefixes. Wdyt @matthiashanel ?

scottf: The JetStreamOptions (jso) in Java (and .NET) are used on a per context basis. You can many of those contexts per connection. So you could have one JS/KV context with a jso with prefix A and another JS/KV context with a jso with prefix B.

But are you are also saying that one context could have a js-api prefix different from the kv prefix? If that is possible, then yes then I would have to add something to the jso to support this. I suppose it would have to be extended for every thing, kv, object store, etc. 

So is that possible and/or do we want to allow that? @matthiashanel

ripienaar: That’s what I understood - same as nats.go

of course the subjects can differ at this point as it’s entirely provided by users and not some future macro in the server. So I think we need to consider a arbitrary prefix as a option and later refine to helpers that support the server macros (while supporting arbitrary)

scottf: It seems to me that allowing an independent kv prfix from the js prefix is just asking for complication. But, if that is what we want to support... 
1. Can someone make a version of the config found [here](https://github.com/nats-io/nats.deno/pull/239) that would have a different js prefix different than the kv prefix
2. How are differing kv prefixes set by the user? If the user provides a js prefix, should we assume it as the kv prefix (that's what we do now) or should they be completely independent, meaning the user always has to specifically set the kv prefix separately from the js prefix.

scottf: This has all been addressed.

aricart:
LGTM

aricart (approved):
LGTM

PR #593, opened 2022-02-08
addressing flapper

scottf:


scottf:
Going to use admin privileges to merge since this is only a unit test change

PR #595, opened 2022-02-09
Consumer config support for max_batch, max_expire and inactive_threshold 

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/90

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsJetStream.java
+++ b/src/main/java/io/nats/client/impl/NatsJetStream.java
@@ -247,7 +247,7 @@ public class NatsJetStream extends NatsJetStreamImplBase implements JetStream {
 
             userCC = so.getConsumerConfiguration();
 
-            validateNotSupplied(userCC.getMaxPullWaiting(), ConsumerConfiguration.CcNumeric.MAX_PULL_WAITING.initial(), JsSubPushCantHaveMaxPullWaiting);
+            validateNotSupplied(userCC.getMaxPullWaiting(), ConsumerConfiguration.CcChangeHelper.MAX_PULL_WAITING.initial(), JsSubPushCantHaveMaxPullWaiting);
 
             // figure out the queue name
             qgroup = validateMustMatchIfBothSupplied(userCC.getDeliverGroup(), queueName, JsSubQueueDeliverGroupMismatch);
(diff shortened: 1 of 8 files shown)


src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -247,7 +247,7 @@ JetStreamSubscription createSubscription(String subject,
 
             userCC = so.getConsumerConfiguration();
 
-            validateNotSupplied(userCC.getMaxPullWaiting(), ConsumerConfiguration.CcNumeric.MAX_PULL_WAITING.initial(), JsSubPushCantHaveMaxPullWaiting);
+            validateNotSupplied(userCC.getMaxPullWaiting(), ConsumerConfiguration.CcChangeHelper.MAX_PULL_WAITING.initial(), JsSubPushCantHaveMaxPullWaiting);
```
scottf, line 250: CcChangeHelper is commented as 
> INTERNAL CLASS ONLY, SUBJECT TO CHANGE

The name was changed because I needed to support AckWait b/c the server populates ack wait in the consumer info response if the user does not supply it.

src/test/java/io/nats/client/impl/JetStreamGeneralTests.java
```diff
@@ -755,6 +768,45 @@ public void testConsumerCannotBeModified() throws Exception {
...
+
+    // default json from server for reference
+    //    "durable_name": "durable",
+    //    "deliver_policy": "all",
+    //    "ack_policy": "explicit",
+    //    "ack_wait": 30000000000,
+    //    "max_deliver": -1,
+    //    "replay_policy": "instant",
+    //    "max_waiting": 512,
+    //    "max_ack_pending": 20000
```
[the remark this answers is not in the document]
scottf: For now I'm just also putting this test in the integration test suite that runs nightly, so if it changes, we'll know.

src/main/java/io/nats/client/impl/NatsJetStream.java
aricart: not quite sure what you mean - server will set a default if not set.

src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -419,15 +419,19 @@ public boolean wouldBeChangeTo(ConsumerConfiguration serverCc) {
                 || (flowControl != null && flowControl != serverCcc.isFlowControl())
                 || (headersOnly != null && headersOnly != serverCcc.isHeadersOnly())
 
-                || CcNumeric.START_SEQ.wouldBeChange(startSeq, serverCcc.startSeq)
-                || CcNumeric.MAX_DELIVER.wouldBeChange(maxDeliver, serverCcc.maxDeliver)
-                || CcNumeric.RATE_LIMIT.wouldBeChange(rateLimit, serverCcc.rateLimit)
-                || CcNumeric.MAX_ACK_PENDING.wouldBeChange(maxAckPending, serverCcc.maxAckPending)
-                || CcNumeric.MAX_PULL_WAITING.wouldBeChange(maxPullWaiting, serverCcc.maxPullWaiting)
+                || CcChangeHelper.START_SEQ.wouldBeChange(startSeq, serverCcc.startSeq)
```
aricart, line 422: Why aren't the values simple pass-throughs? Is this because when you serialize you may be resetting values the user set?

src/test/java/io/nats/client/impl/JetStreamGeneralTests.java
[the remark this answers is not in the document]
aricart: Yes, the server can tweak whatever the defaults are - if you want to be completely in sync with what the server is doing your code will be more complex and fragile.

aricart (approved):
LGTM

src/main/java/io/nats/client/impl/NatsJetStream.java
scottf: This is where I track the defaults that the server sets and returns in the json.

src/main/java/io/nats/client/impl/NatsJetStream.java
scottf: because usually a user setting any value would be a "change" but in the case of the user sets the same value as the server default, it wouldn't be considered a change. So wouldBeChange checks for unset user value, set user value to something like -1 which means server default and an actual value. It's because my builder is too flexible a straight compare is no good.

src/test/java/io/nats/client/impl/JetStreamGeneralTests.java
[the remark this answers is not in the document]
scottf: I've got an integration test that runs nightly, so if any of these changes, that test will fail and we'll know.

PR #599, opened 2022-02-16
Kv parity issues 94, 95, 96

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/94 Items 1 and 3
https://github.com/nats-io/nats-architecture-and-design/issues/95
https://github.com/nats-io/nats-architecture-and-design/issues/96

* 94: v2.7.2+, the KV stream need to be created with the DiscardNew
* 94: use Pending+Delivered.Consumer from the subscription to determine if end of data marker should be sent
* 95: use of prefix as defined in https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-19.md
* 96 PurgeDeletes should keep "recent" markers


aricart (changes requested):

src/main/java/io/nats/client/impl/NatsKeyValueManagement.java
```diff
@@ -39,7 +39,11 @@
      */
     @Override
     public KeyValueStatus create(KeyValueConfiguration config) throws IOException, JetStreamApiException {
-        return new KeyValueStatus(jsm.addStream(config.getBackingConfig()));
+        StreamConfiguration sc = config.getBackingConfig();
+        if ( jsm.conn.getServerInfo().isOlderThanVersion("2.7.2") ) {
+            sc = StreamConfiguration.builder(sc).discardPolicy(null).build(); // null discard policy will use default
```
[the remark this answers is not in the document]
scottf: The default KeyValueConfiguration sets the backing stream config policy to Discard.New. This code is where the stream create is done and it's at this point that I check the server version. If the server version is old, the policy setting is cleared which means it gets the default of Old.

src/main/java/io/nats/client/KeyValue.java
```diff
@@ -196,7 +193,8 @@
     List<KeyValueEntry> history(String key) throws IOException, JetStreamApiException, InterruptedException;
 
     /**
-     * Remove history from all keys that currently are deleted or purged.
+     * Remove history from all keys that currently are deleted or purged
+     * with using a default KeyValuePurgeOptions
      * THIS IS A BETA FEATURE AND SUBJECT TO CHANGE
```
ColinSullivan1, line 198: Can remove this if this makes the KV API GA.  Alternatively, that could be another PR - up to you.

src/main/java/io/nats/client/api/KeyValuePurgeOptions.java
```diff
@@ -0,0 +1,98 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package io.nats.client.api;
+
+import java.time.Duration;
+
+public class KeyValuePurgeOptions {
```
ColinSullivan1, line 18: Does this need doc?

src/main/java/io/nats/client/impl/NatsKeyValue.java
```diff
@@ -35,32 +37,37 @@
...
-    private final String publishKeyPrefix;
+    private final String rawKeyPrefix;
+    private final String pubSubKeyPrefix;
 
     NatsKeyValue(NatsConnection connection, String bucketName, KeyValueOptions kvo) throws IOException {
         this.bucketName = Validator.validateKvBucketNameRequired(bucketName);
         streamName = toStreamName(bucketName);
         streamSubject = toStreamSubject(bucketName);
-        defaultKeyPrefix = toKeyPrefix(bucketName);
+        rawKeyPrefix = toKeyPrefix(bucketName);
```
ColinSullivan1, line 47: Nitpick...  Could simplify a bit and remove rawKeyPrefix, just assign to pubSubKeyPrefix and reassign if necessary.

src/main/java/io/nats/client/impl/NatsKeyValue.java
```diff
@@ -241,16 +244,53 @@ public NatsKeyValueWatchSubscription watchAll(KeyValueWatcher watcher, KeyValueW
...
+     */
+    @Override
+    public void purgeDeletes(KeyValuePurgeOptions options) throws IOException, JetStreamApiException, InterruptedException {
+        long dmThresh = options == null
+            ? KeyValuePurgeOptions.DEFAULT_THRESHOLD_MILLIS
+            : options.getDeleteMarkersThresholdMillis();
+
+        ZonedDateTime limit;
+        if (dmThresh < 0) {
+            limit = DateTimeUtils.fromNow(600000); // long enough in the future to clear all
```
ColinSullivan1, line 261: May want to assign to a private constant...

ColinSullivan1: Or configurable?

ColinSullivan1 (approved):
LGTM, have a few suggestions...

src/main/java/io/nats/client/api/KeyValuePurgeOptions.java
scottf: yes, will do

src/main/java/io/nats/client/impl/NatsKeyValue.java
scottf: rawKeyPrefix is the base. pubSubKeyPrefix varies depending on whether or not there is a JetStream prefix

PR #600, opened 2022-02-21
Issues 10, 91 StreamInfo request changes to retrieve subjects, deleted details

scottf:


src/main/java/io/nats/client/JetStreamManagement.java
```diff
@@ -74,6 +74,17 @@
...
+    /**
+     * Gets the info for an existing stream, and include subject info.
+     * @param streamName the stream name to use.
+     * @param subjectFilter the subject filter to use. Can be wildcard. Null is equivalent to all subjects, the same as '>'
+     * @throws IOException covers various communication issues with the NATS
+     *         server such as timeout or interruption
+     * @throws JetStreamApiException the request had an error related to the data
+     * @return stream information
+     */
+    StreamInfo getStreamInfoWithSubjectInfo(String streamName, String subjectFilter) throws IOException, JetStreamApiException;
```
scottf, line 86: looking for suggestions for a better name for this. I didn't want to overload getStreamInfo b/c the behavior is different, namely the original does not get any subject info

src/main/java/io/nats/client/api/StreamInfoOptions.java
```diff
@@ -0,0 +1,125 @@
+// Copyright 2020 The NATS Authors
```
kozlovic, line 1: Date: 2022

src/main/java/io/nats/client/api/Subject.java
```diff
@@ -0,0 +1,74 @@
+// Copyright 2020 The NATS Authors
```
kozlovic, line 1: Date: 2022

src/main/java/io/nats/client/api/StreamInfoOptions.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @Override
+    public String toJson() {
+        StringBuilder sb = beginJson();
+        addField(sb, SUBJECTS_FILTER, subjectsFilter);
```
kozlovic, line 66: This JSON field should be added ONLY when the subject filter is actually specified. Is addField() doing that under the cover?

src/test/java/io/nats/client/impl/JetStreamManagementTests.java
```diff
@@ -236,9 +239,55 @@ public void testGetStreamInfo() throws Exception {
...
+                PublishAck pa = jsPublish(js, subject(x), data(x + 2));
+                packs.add(pa);
+                jsm.deleteMessage(STREAM, pa.getSeqno());
+            }
+            jsPublish(js, "foo.bar", 6);
+
             StreamInfo si = jsm.getStreamInfo(STREAM);
             assertEquals(STREAM, si.getConfiguration().getName());
+            assertEquals(6, si.getStreamState().getSubjectCount());
+            assertNull(si.getStreamState().getSubjects());
```
kozlovic, line 263: Ok, that seem to indicate that getStreamInfo() without the option is getting no list at all.

kozlovic (changes requested):
Little change to the date for the new file's header. Also, asking for a confirmation about addField() to make sure that nothing gets added if param is "empty/null" string.

src/main/java/io/nats/client/api/StreamInfoOptions.java
scottf: yes, addField checks for null/empty and doesn't add it.

src/test/java/io/nats/client/impl/JetStreamManagementTests.java
scottf: I could have given an empty list though, but null is more consistent with the rest of the api.

src/main/java/io/nats/client/api/Subject.java
scottf: fixed

kozlovic (approved):

PR #604, opened 2022-02-25
don't increment delivered count if no message was retrieved

scottf:
The next message implementation tries to retrieve a message from the internal buffer and increments the delivered count whether or not there was a message. This seems like a bug, and the PR fixes it. Need confirmation that it should only increment the delivered count when a message is actually found.


ColinSullivan1 (approved):
LGTM.

Issue #616, opened 2022-03-18
Support for timeout propagation in async requests


dxxyy:
## Feature Request
Currently the methods handling the request/reply pattern exposed in the `Connection` class don't have support for propagating a timeout for async clients.

#### Use Case:
We are using the async request/reply pattern in our Java application. While testing at 10TPS, we have noticed high memory usage with reachable objects not being properly garbage collected in a timely manner. 

The async clients using the `request` implementation in `NatsConnection` don't have the possibility to pass a timeout to the `NatsRequestCompletableFuture` being created, which is using a default hardcoded timeout of 10 minutes: [NatsRequestCompletableFuture](https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/support/NatsRequestCompletableFuture.java#L11).

#### Proposed Change:
New API exposed, proposition in PR: [ #617 ](https://github.com/nats-io/nats.java/pull/617)

#### Who Benefits From The Change(s)?
Any client using the request/reply pattern in asynchronous way.

#### Alternative Approaches
Leave as is, I don't see an alternative as the current timeout is private and hardcoded in `NatsRequestCompletableFuture`

dxxyy:
Resolved via  [#617](https://github.com/nats-io/nats.java/pull/617)

PR #617, opened 2022-03-21
Support for timeout propagation in async requests

dxxyy:
Give the possibility to clients to specify a timeout while performing async requests.
More details in [#616](https://github.com/nats-io/nats.java/issues/616)

### DIFF
--- a/src/main/java/io/nats/client/Connection.java
+++ b/src/main/java/io/nats/client/Connection.java
@@ -184,6 +184,18 @@ public interface Connection extends AutoCloseable {
      * Send a request. The returned future will be completed when the
      * response comes back.
      *
+     * @param subject the subject for the service that will handle the request
+     * @param body the content of the message
+     * @param timeout the time to wait for a response
+     * @return a Future for the response, which may be cancelled on error or timed out
+     */
+    CompletableFuture<Message> requestWithTimeout(String subject, byte[] body, Duration timeout);
+
+
+    /**
+     * Send a request. The returned future will be completed when the
+     * response comes back.
+     *
      * <p>The Message object allows you to set a replyTo, but in requests,
      * the replyTo is reserved for internal use as the address for the
      * server to respond to the client with the consumer's reply.</p>
@@ -193,6 +205,20 @@ public interface Connection extends AutoCloseable {
      */
     CompletableFuture<Message> request(Message message);
 
+    /**
+     * Send a request. The returned future will be completed when the
+     * response comes back.
+     *
+     * <p>The Message object allows you to set a replyTo, but in requests,
+     * the replyTo is reserved for internal use as the address for the
+     * server to respond to the client with the consumer's reply.</p>
+     *
+     * @param message the message
+     * @param timeout the time to wait for a response
+     * @return a Future for the response, which may be cancelled on error or timed out
+     */
+    CompletableFuture<Message> requestWithTimeout(Message message, Duration timeout);
+    
     /**
      * Send a request and returns the reply or null. This version of request is equivalent
      * to calling get on the future returned from {@link #request(String, byte[]) request()} with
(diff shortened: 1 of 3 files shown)


scottf (approved):
Looks good to me.

src/main/java/io/nats/client/Connection.java
```diff
@@ -193,6 +193,20 @@
...
+     *
+     * <p>The Message object allows you to set a replyTo, but in requests,
+     * the replyTo is reserved for internal use as the address for the
+     * server to respond to the client with the consumer's reply.</p>
+     *
+     * @param message the message
+     * @param timeout the time to wait for a response
+     * @return a Future for the response, which may be cancelled on error or timed out
+     */
+    CompletableFuture<Message> requestWithTimeout(Message message, Duration timeout);
```
ColinSullivan1, line 208: For parity, I think we'd also want:
```java
CompletableFuture<Message> requestWithTimeout(String subject, byte[] body, Duration timeout);
```

wdyt @scottf ?

src/test/java/io/nats/client/impl/RequestTests.java
```diff
@@ -239,6 +239,39 @@ public void testRequireCleanupOnTimeoutNoNoResponders() throws IOException, Inte
         }
     }
 
+    @Test
+    public void testRequireCleanupOnTimeoutCleanCompletable() throws IOException, InterruptedException {
```
ColinSullivan1, line 243: I'd like to see some tests that also succeed, and then fail with a timeout.  This one does test failure w/ cleanup - but would like to see that the right exception is thrown.

ColinSullivan1:
Thanks for contributing!  I offered a few suggestions but will defer to @scottf.

src/main/java/io/nats/client/Connection.java
scottf: @dxxyy please add.

dxxyy: will do!

src/test/java/io/nats/client/impl/RequestTests.java
dxxyy: Hi! For a "succeed and then fail" what do you mean exactly? I see that the timeout of the `NatsRequestCompletableFuture` is considered only upon cleanup.

I can add an additional test that triggers the timeout exception, coming from the `CompletableFuture`.
Something like:
`assertThrows(TimeoutException.class, () -> nc.requestWithTimeout(nm, Duration.ofMillis(cleanupInterval)).get(100, TimeUnit.MILLISECONDS));`

Thanks for your reply

scottf: so this this is request/reply, the timeout is suggesting how long to wait for a response. Test actually getting a response back and also not getting a response back. Although No responders may come into play. What do other tests for the default timeout do?

dxxyy: added more tests and api update

scottf (approved):
LGTM

ColinSullivan1 (approved):
LGTM.  Thanks for the contribution!

dxxyy:
I see 5 failures in the build, after the pushed merge commit:

1. JetStreamPushTests > testPushDurableNullDeliver
2. JetStreamPushTests > testPushEphemeralWithDeliver
3. JetStreamGeneralTests > testBindPush
4. KeyValueTests > testCreateAndUpdate
5. KeyValueTests > testWatch

-> Theses are not linked to the change (and the new exposed API), let me know if I need to take some actions in order to have this merged.

Thanks!

scottf:
> I see 5 failures in the build, after the pushed merge commit:
Yes they are not related to this PR. They are "flappers", exacerbated by extremely slow GHA runners, but they are functional and build fine in other PR's and the main. It's a crapshoot sometimes. 
This PR will get merged I've just been in the middle of other stuff.

scottf:
Using admin merge because GHA/coveralls isn't reporting after hours of waiting.

PR #619, opened 2022-03-22
Issue 89 backoff

scottf:
Just the backoff portion https://github.com/nats-io/nats-architecture-and-design/issues/89

Also change to Purge Options to use -1 instead of 0 as the default long so they don't get put in the outgoing json


scottf:
replaced by https://github.com/nats-io/nats.java/pull/620

PR #620, opened 2022-03-23
Issue 89 nak delay and backoff

scottf:
https://github.com/nats-io/nats-architecture-and-design/issues/89

Also change to Purge Options to use -1 instead of 0 as the default long so they don't get put in the outgoing json


kozlovic (approved):
LGTM

PR #630, opened 2022-03-30
configuration change check improvement and flappers

scottf:


scottf:
Awaiting code changes require this to be merged with admin privileges.

PR #635, opened 2022-04-22
Value or default bug

scottf:
addressing server defaults and unsigned ints better

### DIFF
--- a/src/main/java/io/nats/client/api/ConsumerConfiguration.java
+++ b/src/main/java/io/nats/client/api/ConsumerConfiguration.java
@@ -67,10 +67,6 @@ public class ConsumerConfiguration implements JsonSerializable {
     protected final Boolean headersOnly;
     protected final List<Duration> backoff;
 
-    private static DeliverPolicy GetOrDefault(DeliverPolicy p) { return p == null ? DEFAULT_DELIVER_POLICY : p; }
-    private static AckPolicy GetOrDefault(AckPolicy p) { return p == null ? DEFAULT_ACK_POLICY : p; }
-    private static ReplayPolicy GetOrDefault(ReplayPolicy p) { return p == null ? DEFAULT_REPLAY_POLICY : p; }
-
     protected ConsumerConfiguration(ConsumerConfiguration cc) {
         this.deliverPolicy = cc.deliverPolicy;
         this.ackPolicy = cc.ackPolicy;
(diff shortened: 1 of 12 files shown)


src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -175,7 +175,7 @@ public String toJson() {
         JsonUtils.addField(sb, DELIVER_SUBJECT, deliverSubject);
         JsonUtils.addField(sb, DELIVER_GROUP, deliverGroup);
         JsonUtils.addField(sb, DELIVER_POLICY, GetOrDefault(deliverPolicy).toString());
-        JsonUtils.addField(sb, OPT_START_SEQ, startSeq);
+        JsonUtils.addUnsigned(sb, OPT_START_SEQ, startSeq);
```
scottf, line 178: add as unsigned writes if the value is greater than 0 versus addField writes it if it's greater or equal

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -981,11 +981,11 @@ public String toString() {
...
     public enum CcChangeHelper {
-        START_SEQ(1, -1),
+        START_SEQ(1, 0),
         MAX_DELIVER(1, -1),
-        RATE_LIMIT(1, -1),
-        MAX_ACK_PENDING(0, 0),
-        MAX_PULL_WAITING(0, 0),
+        RATE_LIMIT(1, 0),
+        MAX_ACK_PENDING(0, -1),
+        MAX_PULL_WAITING(0, -1),
```
scottf, line 988: start seq and rate limit are unsigned, so 1 didn't make sense. max ack pending and max pull waiting are reverted back to they way they were in 2.14.0 (this is 2.14.1)

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -998,7 +998,7 @@ public enum CcChangeHelper {
         }
 
         public long valueOrDefault(Long val) {
-            return val == null ? START_SEQ.Default : val;
+            return val == null ? Default : val;
```
scottf, line 1001: the other code was just wrong

src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -426,9 +426,9 @@ public boolean wouldBeChangeTo(ConsumerConfiguration serverCc) {
                 || (flowControl != null && flowControl != serverCcc.isFlowControl())
                 || (headersOnly != null && headersOnly != serverCcc.isHeadersOnly())
 
-                || CcChangeHelper.START_SEQ.wouldBeChange(startSeq, serverCcc.startSeq)
+                || CcChangeHelper.START_SEQ.wouldBeChange(startSeq, serverCcc.getStartSequence())
```
scottf, line 429: the server sends nothing when an unsigned int is 0. By using the getter here, null is understood as 0. Not strictly necessary, but will allow users to set the value 0 and it not be considered a change if the server sends null for 0

scottf: this has been reverted

src/main/java/io/nats/client/api/ConsumerConfiguration.java
tbeets: Naming of `addUnsigned` method seems confusing.  Is not 0 (zero) an unsigned integer?   Why would I expect a side-effect that 0 (zero) is ignored?

tbeets (approved):
LGTM

PR #645, opened 2022-05-21
Pull config changes, ephemeral pull, unit tests

scottf:
* architecture issue 112 Pull config changes  in ConsumerConfig and JSApiConsumerGetNextRequest
* remove enforcement of durable on pull subscriptions allowing ephemeral pull consumers
* unit tests for above and additional unit tests to shore up coverage

### DIFF
--- a/README.md
+++ b/README.md
@@ -574,7 +574,7 @@ Subscription creation has many checks to make sure that a valid, operable subscr
 | JsSubPushCantHaveMaxPullWaiting | SUB | 90003 | Push subscriptions cannot supply max pull waiting. |
 | JsSubQueueDeliverGroupMismatch | SUB | 90004 | Queue / deliver group mismatch. |
 | JsSubFcHbNotValidPull | SUB | 90005 | Flow Control and/or heartbeat is not valid with a pull subscription. |
-| JsSubFcHbHbNotValidQueue | SUB | 90006 | Flow Control and/or heartbeat is not valid in queue mode. |
+| JsSubFcHbNotValidQueue | SUB | 90006 | Flow Control and/or heartbeat is not valid in queue mode. |
 | JsSubNoMatchingStreamForSubject | SUB | 90007 | No matching streams for subject. |
 | JsSubConsumerAlreadyConfiguredAsPush | SUB | 90008 | Consumer is already configured as a push consumer. |
 | JsSubConsumerAlreadyConfiguredAsPull | SUB | 90009 | Consumer is already configured as a pull consumer. |
--- a/README.md
+++ b/README.md
@@ -588,7 +588,7 @@ Subscription creation has many checks to make sure that a valid, operable subscr
 | JsSubConsumerNotFoundRequiredInBind | SUB | 90017 | Consumer not found, required in bind mode. |
 | JsSubOrderedNotAllowOnQueues | SUB | 90018 | Ordered consumer not allowed on queues. |
 | JsSubPushCantHaveMaxBatch | SUB | 90019 | Push subscriptions cannot supply max batch. |
-| JsSubPushCantHaveMaxBatch | SUB | 90020 | Push subscriptions cannot supply max bytes. |
+| JsSubPushCantHaveMaxBytes | SUB | 90020 | Push subscriptions cannot supply max bytes. |
 
 ### Message Acknowledgements
 
--- a/README.md
+++ b/README.md
@@ -588,6 +588,7 @@ Subscription creation has many checks to make sure that a valid, operable subscr
 | JsSubConsumerNotFoundRequiredInBind | SUB | 90017 | Consumer not found, required in bind mode. |
 | JsSubOrderedNotAllowOnQueues | SUB | 90018 | Ordered consumer not allowed on queues. |
 | JsSubPushCantHaveMaxBatch | SUB | 90019 | Push subscriptions cannot supply max batch. |
+| JsSubPushCantHaveMaxBatch | SUB | 90020 | Push subscriptions cannot supply max bytes. |
 
 ### Message Acknowledgements
 
--- a/src/main/java/io/nats/client/PurgeOptions.java
+++ b/src/main/java/io/nats/client/PurgeOptions.java
@@ -45,26 +45,6 @@ public class PurgeOptions implements JsonSerializable {
         return endJson(sb).toString();
     }
 
-    @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-
-        PurgeOptions options = (PurgeOptions) o;
-
-        if (seq != options.seq) return false;
-        if (keep != options.keep) return false;
-        return subject != null ? subject.equals(options.subject) : options.subject == null;
-    }
-
-    @Override
-    public int hashCode() {
-        int result = subject != null ? subject.hashCode() : 0;
-        result = 31 * result + (int) (seq ^ (seq >>> 32));
-        result = 31 * result + (int) (keep ^ (keep >>> 32));
-        return result;
-    }
-
     /**
      * Get the subject for the Purge Options
      * @return the subject
@@ -141,6 +121,10 @@ public class PurgeOptions implements JsonSerializable {
             return this;
         }
 
+        /**
+         * Build the PurgeOptions
+         * @return the built PurgeOptions
+         */
         public PurgeOptions build() {
             validateSubject(subject, false);
 
--- a/src/main/java/io/nats/client/api/ConsumerConfiguration.java
+++ b/src/main/java/io/nats/client/api/ConsumerConfiguration.java
@@ -1075,7 +1075,7 @@ public class ConsumerConfiguration implements JsonSerializable {
         return val == null ? DURATION_UNSET : val;
     }
 
-    private static Integer normalizeToInt(Long l, int min) {
+    protected static Integer normalizeToInt(Long l, int min) {
         if (l == null) {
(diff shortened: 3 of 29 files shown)


src/main/java/io/nats/client/PurgeOptions.java
```diff
@@ -45,26 +45,6 @@ public String toJson() {
         return endJson(sb).toString();
     }
 
-    @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
```
scottf, line 50: Not needed, never used the object is not compared with anything. Not sure why this was added in the first place.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -68,11 +69,12 @@ public class ConsumerConfiguration implements JsonSerializable {
...
-    protected final Long maxDeliver;
+    protected final Integer maxDeliver;
     protected final Long rateLimit; // server side this is unsigned
-    protected final Long maxAckPending;
-    protected final Long maxPullWaiting;
-    protected final Long maxBatch;
+    protected final Integer maxAckPending;
+    protected final Integer maxPullWaiting;
+    protected final Integer maxBatch;
+    protected final Integer maxBytes;
```
scottf, line 77: After reviewing the server code I realized that the fields were go int, which are 32 bit signed values aka java int. I have only changed the internal variable type, getters and builders were left the same.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -1027,9 +1072,25 @@ protected static Duration getOrUnset(Duration val)
...
+        }
+
+        long l = proposed.longValue();
+        if (l < min) {
+            return INTEGER_UNSET;
+        }
+
+        if (l > Integer.MAX_VALUE) {
+            return Integer.MAX_VALUE;
+        }
```
scottf, line 1091: A value for these integer fields anywhere near Integer MAX_VALUE would not make sense, so instead of an error if the proposed long value was beyond that, I just assumed the user was trying to set some max value and handled it instead of erroring.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -1055,27 +1116,25 @@ private static Duration normalizeDuration(long millis)
      * INTERNAL CLASS ONLY, SUBJECT TO CHANGE
      * Helper class to manage min / default / unset / server values.
      */
-    public enum LongChangeHelper {
-        MAX_DELIVER(1, -1);  // 0 is treated the same as -1 on the server, which is why the server doesn't omit this
+    public enum IntegerChangeHelper {
```
scottf, line 1119: reflecting that it's addressing integers, not longs

src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -255,6 +255,10 @@ JetStreamSubscription createSubscription(String subject,
                 throw JsSubPushCantHaveMaxBatch.instance();
             }
 
+            if (userCC.maxBytesWasSet()) {
+                throw JsSubPushCantHaveMaxBytes.instance();
+            }
+
```
scottf, line 261: new field added so new validation added.

README.md
```diff
@@ -588,6 +588,7 @@ Subscription creation has many checks to make sure that a valid, operable subscr
 | JsSubConsumerNotFoundRequiredInBind | SUB | 90017 | Consumer not found, required in bind mode. |
 | JsSubOrderedNotAllowOnQueues | SUB | 90018 | Ordered consumer not allowed on queues. |
 | JsSubPushCantHaveMaxBatch | SUB | 90019 | Push subscriptions cannot supply max batch. |
+| JsSubPushCantHaveMaxBatch | SUB | 90020 | Push subscriptions cannot supply max bytes. |
```
scottf, line 591: new field added so new validation added.

src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -454,11 +458,12 @@ public List<String> getChanges(ConsumerConfiguration serverCc) {
...
-            record(LongChangeHelper.MAX_DELIVER.wouldBeChange(maxDeliver, serverCcc.maxDeliver), "maxDeliver", changes);
+            record(IntegerChangeHelper.MAX_DELIVER.wouldBeChange(maxDeliver, serverCcc.maxDeliver), "maxDeliver", changes);
 
-            record(maxAckPending != null && !maxAckPending.equals(serverCcc.getMaxAckPending()), "maxAckPending", changes);
-            record(maxPullWaiting != null && !maxPullWaiting.equals(serverCcc.getMaxPullWaiting()), "maxPullWaiting", changes);
-            record(maxBatch != null && !maxBatch.equals(serverCcc.getMaxBatch()), "maxBatch", changes);
+            record(maxAckPending != null && !maxAckPending.equals((int)serverCcc.getMaxAckPending()), "maxAckPending", changes);
+            record(maxPullWaiting != null && !maxPullWaiting.equals((int)serverCcc.getMaxPullWaiting()), "maxPullWaiting", changes);
+            record(maxBatch != null && !maxBatch.equals(serverCcc.getMaxBatch().intValue()), "maxBatch", changes);
+            record(maxBytes != null && !maxBytes.equals(serverCcc.getMaxBytes()), "maxBytes", changes);
```
scottf, line 466: casting to int is required because an Integer can never be equal to a Long object, and the getters represent integers anyway even though their return type was not changed, avoiding backward compatibility problems.

src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
```diff
@@ -96,22 +95,7 @@ public void pullExpiresIn(int batchSize, Duration expiresIn) {
...
-        connection.publish(publishSubject, getSubject(), getPullJson(batch, noWait, expiresIn));
-        connection.lenientFlushBuffer();
-    }
-
-    byte[] getPullJson(int batch, boolean noWait, Duration expiresIn) {
-        StringBuilder sb = JsonUtils.beginJson();
-        JsonUtils.addField(sb, "batch", batch);
-        JsonUtils.addFldWhenTrue(sb, "no_wait", noWait);
-        JsonUtils.addFieldAsNanos(sb, "expires", expiresIn);
-        return JsonUtils.endJson(sb).toString().getBytes(StandardCharsets.US_ASCII);
```
scottf, line 114: _pull and getPullJson are not needed because the json is done in the PullRequestOptions object and the _pull code is moved to the new api that directly takes the PullRequestOptions since the other pulls can delegate to it now

src/main/java/io/nats/client/impl/NatsJetStreamSubscription.java
```diff
@@ -150,6 +151,14 @@ public void pull(int batchSize) {
         throw new IllegalStateException(SUBSCRIPTION_TYPE_DOES_NOT_SUPPORT_PULL);
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void pull(PullRequestOptions pullRequestOptions) {
+        throw new IllegalStateException(SUBSCRIPTION_TYPE_DOES_NOT_SUPPORT_PULL);
+    }
```
scottf, line 160: Only pull subs allow this call and this is the base class, not the pull version so is the default behavior. For simplification, the interface design should be more like .NET where the pull interface is not part of the JS Sub interface

src/main/java/io/nats/client/impl/PullStatusMessageManager.java
```diff
@@ -23,22 +23,13 @@ class PullStatusMessageManager extends MessageManager {
 
     private static final List<Integer> PULL_KNOWN_STATUS_CODES = Arrays.asList(404, 408, 409);
 
-    private int lastStatusCode = -1;
```
scottf, line 26: this field an associated getter were not used as noted by the IDE. So removed them.

src/test/java/io/nats/client/impl/ConsumerConfigurationComparerTests.java
```diff
@@ -101,7 +101,7 @@ public void testChangeFieldsIdentified() {
         assertNotChange(builder(orig).startSequence(null).build(), orig);
         assertChange(builder(orig).startSequence(1).build(), orig, "startSequence");
 
-        assertNotChange(builder(orig).maxDeliver(MAX_DELIVER.Unset).build(), orig);
+        assertNotChange(builder(orig).maxDeliver(INTEGER_UNSET).build(), orig);
```
scottf, line 104: the Unset field has been removed from the object in place of using a constant.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -281,8 +290,8 @@ public Duration getAckWait() {
      * Gets the max delivery amount of this consumer configuration.
      * @return the max delivery amount.
      */
-    public long getMaxDeliver() {
-        return LongChangeHelper.MAX_DELIVER.getOrUnset(maxDeliver);
+    public int getMaxDeliver() {
```
scottf, line 293: changed getters to return int since the underlying value is an int. This is not a breaking change since it is a primitive and used to be primitive long, so there would be no break if the user had read this value into an long variable.
applies to getMaxDeliver, getMaxAckPending, getMaxPullWaiting, getMaxBatch, getMaxBytes

tbeets (approved):
LGTM

PR #648, opened 2022-05-25
Server urls connection management

scottf:
1. architecture issue 113 Add option to ignore discovered urls
2. ServersToTryProvider provide a way that a user can provide the server urls to try on connect / reconnect. This supports eventual tiered server bootstrap.

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -39,16 +39,20 @@ import static io.nats.client.support.NatsConstants.*;
  * starting with the {@link Options.Builder Builder}, since it has a simple list of methods that configure the connection.
  */
 public class Options {
-    // NOTE TO DEVS
-    // To add an option, you have to:
-    // * add property
-    // * add field in builder
-    // * add field in options
-    // * add a chainable method in builder
-    // * add update build
-    // * update constructor that takes properties
-    // * optional default in statics
+    // NOTE TO DEVS!!! To add an option, you have to address:
+    // CONSTANTS * optionally add a default value constant
+    // ENVIRONMENT * most of the time add an environment property
+    // CLASS VARIABLES * add a class variable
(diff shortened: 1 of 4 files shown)


src/main/java/io/nats/client/Options.java
```diff
@@ -1391,8 +1464,16 @@ else if (OPENTLS_PROTOCOL.equals(serverURI.getScheme())) {
         }
     }
 
+    // ----------------------------------------------------------------------------------------------------
+    // CONSTRUCTOR
+    // ----------------------------------------------------------------------------------------------------
     private Options(Builder b) {
-        this.servers = b.servers;
+        this.servers = new ArrayList<>(b.servers); // builder servers is a set so no dupes
+        refinedServers = new ArrayList<>(this.servers.size()); // refined just turns uris into strings since that's what's really used
```
scottf, line 1472: is refined de-duped?

src/main/java/io/nats/client/ServerListProvider.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Get the ordered server list to try for connecting/reconnecting
+     * @param currentServer the server that connection is currently connected to. May be null.
+     * @param optionsServersRefined the list of normalized server URIs given to the options
+     * @param optionsServersUnprocessed the list of server urls exactly how they were given to the options
+     * @param discoveredServersRefined the list of normalized servers from the server info
+     * @param discoveredServersUnprocessed the entire list of servers exactly as returned in the server info
+     * @return the ordered server list
+     */
+    List<String> getServerList(String currentServer,
```
scottf, line 31: simplify

src/main/java/io/nats/client/Options.java
scottf: it was not de-duped, all it was was a String version of the URI list. IO removed it anyway as part of simplification

PR #655, opened 2022-06-03
ability for user to create a Placement object

scottf:
There was never a public way for a user to create the Placement object, so you could not configure a stream with a placement.


tbeets (approved):
LGTM

PR #657, opened 2022-06-06
bump version to 2.15.4

scottf:
version should be bumped as soon as a release is done so merges to main build the correct snapshot.


piotrpio (approved):
LGTM

PR #661, opened 2022-06-08
Ability to create an External object, some api docs

scottf:
In order to create sources and mirrors in stream configuration you must be able to create an External object.


src/main/java/io/nats/client/api/External.java
```diff
@@ -91,4 +101,48 @@ public int hashCode() {
...
+            return this;
+        }
+
+        /**
+         * Set the deliver string.
+         * @param deliver the deliver
+         * @return the builder
+         */
+        public Builder deliver(String deliver) {
+            this.deliver = api;
```
piotrpio, line 136: Shouldn't this be `this.deliver = deliver;`?
Btw, I think should be covered by unit tests to avoid such mistakes.

src/main/java/io/nats/client/api/External.java
```diff
@@ -91,4 +101,48 @@ public int hashCode() {
...
+         * @return the builder
+         */
+        public Builder api(String api) {
+            this.api = api;
+            return this;
+        }
+
+        /**
+         * Set the deliver string.
+         * @param deliver the deliver
```
piotrpio, line 132:
```suggestion
         * @param deliver the delivery subject
```

src/test/java/io/nats/client/api/StreamConfigurationTests.java
```diff
@@ -107,19 +107,23 @@ public void testConstruction() {
...
-                Mirror m2 = new Mirror(l2);
-                if (l1.equals(l2)) {
-                    assertEquals(m1, m2);
-                }
-                else {
-                    assertNotEquals(m1, m2);
+            if (l1.startsWith("{")) {
+                Mirror m1 = new Mirror(l1);
+                assertEquals(m1, m1);
+                assertNotEquals(m1, null);
```
piotrpio, line 115:
```suggestion
                assertNotNull(m1);
```

src/test/java/io/nats/client/api/StreamConfigurationTests.java
```diff
@@ -107,19 +107,23 @@ public void testConstruction() {
...
-            for (String l2 : lines) {
-                Mirror m2 = new Mirror(l2);
-                if (l1.equals(l2)) {
-                    assertEquals(m1, m2);
-                }
-                else {
-                    assertNotEquals(m1, m2);
+            if (l1.startsWith("{")) {
+                Mirror m1 = new Mirror(l1);
+                assertEquals(m1, m1);
```
piotrpio, line 114: What is the point of this comparison?

piotrpio (changes requested):

src/test/java/io/nats/client/api/StreamConfigurationTests.java
scottf: it's reading the list data in the file line by line, doing individual comparisons. I don't want to look at the line unless it's a full object.

src/main/java/io/nats/client/api/External.java
scottf: fixed and added test

src/test/java/io/nats/client/api/StreamConfigurationTests.java
scottf: this is specifically exercising the equals testing m1 versus null. I have changed it to this to be clear
```
//this provides testing coverage 
//noinspection ConstantConditions,SimplifiableAssertion
assertTrue(!m1.equals(null));
```

PR #670, opened 2022-06-24
better push queue example

scottf:
old version did not use the deliver group configuration, relying on the fact that the client will fix it on the initial consumer if the subscription is a queue subscription


tbeets (approved):
LGTM

PR #674, opened 2022-06-29
Accept encoded connection urls

scottf:
Wrote unit test to verify that encoded urls are accepted. This uncovered an issue with the an encoded colon `:` in the url, so this was fixed.

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -19,9 +19,11 @@ import io.nats.client.impl.SocketDataPort;
 import io.nats.client.support.SSLUtils;
 
 import javax.net.ssl.SSLContext;
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Constructor;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLDecoder;
 import java.nio.CharBuffer;
 import java.security.NoSuchAlgorithmException;
 import java.time.Duration;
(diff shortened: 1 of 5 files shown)


src/main/java/io/nats/client/Options.java
```diff
@@ -1950,16 +1950,15 @@ public CharBuffer buildProtocolConnectOptionsString(String serverURI, boolean in
...
                 String userInfo = uri.getUserInfo();
-
                 if (userInfo != null) {
-                    String[] info = userInfo.split(":");
-
-                    if (info.length == 2) {
-                        uriUser = info[0];
-                        uriPass = info[1];
-                    } else {
+                    int at = userInfo.indexOf(":");
```
caleblloyd, line 1954: I think that the split needs to happen _before_ the decode, because a Username or Token could have an encoded colon character in it `:`

consider the user `test:user` with the password `test:password`

that encoded string would be `test%3Auser:test%3Apassword`

Also, consider the token authentication with a token `test:token`

that encoded string would be `test%3Atoken`

scottf: Yeah, the problem was the java uri `getUserInfo` call already decoded the entire part, so I would not know where to split. It's now fixed to properly handle encoded colon by using `getRawInfo` and decoding the parts separately,. Also made sure the plus `+` was not touched during the decoding.

RichardHightower (approved):
LGTM

src/main/java/io/nats/client/support/Encoding.java
```diff
@@ -200,4 +200,32 @@ public static void jsonEncode(StringBuilder sb, String s) {
             }
         }
     }
+
+    public static String uriDecode(String source) {
```
caleblloyd, line 204: This method doesn't handle UTF-8 properly, could it be simplified to just call `URLDecoder.decode(source.replace("+", "%2B"))` ?

```
String test = "+" + URLEncoder.encode("😀") + "+";
System.out.println(test);
// +%F0%9F%98%80+
System.out.println(uriDecode("+%F0%9F%98%80+"));
// +ð+
System.out.println(URLDecoder.decode("+%F0%9F%98%80+".replace("+", "%2B")));
// +😀+
```

caleblloyd (changes requested):

src/main/java/io/nats/client/support/Encoding.java
scottf: changed to support unicode

PR #675, opened 2022-06-29
remove comments that say durable is required during pull

scottf:


arondi (approved):
This looks like a doc change only. Recent server change for pull to allow ephemeral. LGTM

arondi (approved):
LGTM

PR #681, opened 2022-06-30
expose management getStreamNamesBySubjectFilter

scottf:
The ability to get a list of streams that match a subject filter


arondi (approved):
Experimental. LGTM

PR #685, opened 2022-07-04
revert ConsumerConfiguration changes where some fields were downgrade…

scottf:
…d to int from long


arondi (changes requested):
There is an extra space by the version change

arondi (approved):
LGTM

PR #686, opened 2022-07-10
Add numReplicas field in ConsumerConfiguration

goku321:
This adds "numReplicas" field in the `ConsumerConfiguration` class to allow configuring the replicas count for a consumer.


scottf:
I'll start a review, but I found some things to change; we use objects for the field variables, but primitives for the getters. This allows us to track whether the use supplied a value to the build or left it unset. A field change in `ConsumerConfiguration` needs to be reflected in `NatsJetStream getChanges` . There are some more tests to do, the test json comes to mind. I'd also like to split the checking of `numReplicas` to its own test instead of adding it to `testAddDeleteConsumer` and `testGetConsumerInfo`.

scottf:
Replaced by https://github.com/nats-io/nats.java/pull/689

Issue #687, opened 2022-07-15
Unexpected changes of default timeout value in NatsRequestCompletableFuture


osadchuk-roman:
## Defect

In recent PR https://github.com/nats-io/nats.java/pull/659, which was dedicated to fix issue https://github.com/nats-io/nats.java/issues/658, the default waiting for response timeout was decreased from **10 MINUTES** to **2 SECONDS** in `NatsRequestCompletableFuture`. 

I know that there is possibility to use `Connection#request` with explicitly specified timeout as method parameter, but a lot of developers which were using this library previously may relay on default 10m timeout. So, to start using new version `io.nats:jnats` they must to review all places in their code when any request is sending to NATS and explicitly specify needed timeout. Moreover, such implicit changes of timeout may lead to unexpected bugs. Fortunately, I've been lucky to notice this problem by failing integration test before merging version update to master branch.

As for me, 2s delay is not enough to consider it as error scenario and cancel NatsRequestCompletableFuture. But if this is not a bug from your point of view and there is no possibility to revert timeout changes in `NatsRequestCompletableFuture`, it would be nice to have some configurable connection property dedicated for default response timeout rather than just use `Options.DEFAULT_CONNECTION_TIMEOUT` constant value. Such approach will make migration to newer version of `io.nats:jnats` less painful.

#### Versions of `io.nats:jnats` and `nats-server`:
2.15.5

#### Steps or code to reproduce the issue:
Invoke `Connection#request` method without specifying timeout

#### Expected result:
Response timeout more then 2s

#### Actual result:
Response timeout 2s

scottf:
I will change this to be the request timeout plus a small amount. I see how this is a problem for devs using longer than 2 sec timeout for their request.

scottf:
@osadchuk-roman For now, you can use the `Connection#requestWithTimeout`

osadchuk-roman:
Thanks @scottf, yes, I know that `Connection#requestWithTimeout` could be used for now. 
As for me, it will be great to have separate option instead of `Options.DEFAULT_CONNECTION_TIMEOUT`, since "connection timeout" is not the same as "response timeout".  Or at least use `options.getConnectionTimeout()`, which returns value defined by developer, instead of default constant.

scottf:
@osadchuk-roman can you look at this PR and comment?

osadchuk-roman:
@scottf In general I'm OK with such solution. 

[...]

PR #693, opened 2022-07-27
better request timeout management

scottf:
https://github.com/nats-io/nats.java/issues/687

Changed the code to use Options RequestCleanupInterval, which defaults to 5 seconds.


arondi (approved):
LGTM

PR #700, opened 2022-08-09
BUG FIX: kv get against a deleted or purged key

scottf:
BUG FIX: kv get or get with revision against a deleted or purged key should not return an entry.


RichardHightower (approved):
LGTM

PR #720, opened 2022-09-10
Digester Verification

scottf:


piotrpio (approved):
LGTM

aricart:
Wondering if we can make this data available the network so that all clients could test hashing it without having to copy the file - ie make a JSON with the expected hash, followed by the body - clients could HTTP it from github.

PR #818, opened 2022-12-28
service type field

scottf:


src/main/java/io/nats/service/Service.java
```diff
@@ -171,8 +171,8 @@ public Info getInfo() {
         return info;
     }
 
-    public SchemaInfo getSchemaInfo() {
-        return schemaInfo;
+    public SchemaResponse getSchemaInfo() {
```
arondi, line 174: getSchemaResponse

arondi (changes requested):

arondi (approved):
LGTM

PR #826, opened 2023-01-05
Websocket Beta

scottf:
From @brimworks :

This is the simplest possible implementation of websockets with a net increase in code coverage.
* No alterations to DataPort.
* Two new options, one to support httpInterceptors and another to support proxy requests.
* Websockets is supported via a new schema ("ws" or "wss").
* Copy writer headers are at the top of every new file added.
* All websocket implementation details are encapsulated by extending Socket and implementing WebSocket with a WebSocketInputStream and WebSocketOutputStream.

Being critical of myself, these are the points that are lacking:
* Proper websocket closure handshake is not implemented, close() simply forwards to the underlying socket.
* The http interceptors don't allow filtering HTTP responses and are very primitive. They work for my use case, but not sure if it works for all use-cases.
* The handling of upgradeToSecure is a little awkward since websockets is typically performed over a TLS channel and thus there is no need for double encryption. A better solution would involve delegating this to the DataPort... or more specifically, when the secure() option is set, we should NOT throw an error if "wss" transport is used.
* Create URI for server now takes in an "is websocket" parameter, again this is a little awkward. A better solution would involve delegating this to the DataPort.
* No support for `Sec-WebSocket-Extensions: permessage-deflate` (aka websocket level compression... which should really be combined with `Nats-No-Masking: TRUE` to really get the full benefits).
* The implementation of websocket framing strikes a balance between TCP fragmentation and extra buffer copying by using a 1440 sized buffer used for the initial frame header + the first ~1438 bytes. The channel based implementation mitigates this fragmentation by using the GatheringByteChannel and passing in an array of ByteBuffers.

For more information on websockets, I'd really recommend reading the RFC: https://datatracker.ietf.org/doc/html/rfc6455


ColinSullivan1 (approved):
LGTM.  Thanks for the contribution!

src/main/java/io/nats/client/Options.java
```diff
@@ -544,6 +546,10 @@ public class Options {
     private final ExecutorService executor;
     private final ServerListProvider serverListProvider;
 
+    // TODO SFF What is this
```
brimworks, line 549: We needed to add special headers when connecting through a WAF (Web Application Firewall). By supporting a list of interceptors like this, one can inject these sort of custom headers.

src/main/java/io/nats/client/Options.java
```diff
@@ -544,6 +546,10 @@ public class Options {
     private final ExecutorService executor;
     private final ServerListProvider serverListProvider;
 
+    // TODO SFF What is this
+    private final List<java.util.function.Consumer<HttpRequest>> httpRequestInterceptors;
+    private final Proxy proxy;
```
brimworks, line 551: Proxy support was needed so that we could run code that connects through http proxy servers.

src/main/java/io/nats/client/support/WebSocket.java
```diff
@@ -0,0 +1,365 @@
...
+
+    private static String readLine(byte[] buffer, InputStream in) throws IOException {
+        int offset = 0;
+        int lastCh = -1;
+        while (true) {
+            int ch = in.read();
+            switch (ch) {
+            case -1:
+                // Premature EOF (everything should be terminated with \n)
+                return new String(buffer, 0, offset);
```
brimworks, line 148: I forgot to explicitly specify the charset here and 3 lines down. This should be:

```
return new String(buffer, 0, offset, StandardCharsets.ISO_8859_1);
```

...the choice of ISO_8859_1 is purely because every byte maps to a character and because this is a super-set of ascii.

The only issue with the existing code is if the default JVM character set is not compatible with ASCII (such as an EBCDIC variant).

I realize this is asymmetric with the fact that the HTTP request uses UTF-8, but the only code using the response from the `readLine()` is only processing very specific headers for which ASCII is all we care about.

src/test/java/io/nats/client/OptionsTests.java
```diff
@@ -688,6 +710,24 @@ public void testInboxPrefixCoverage() {
         assertEquals("foo.", o.getInboxPrefix());
     }
 
+    @Test
+    public void testSslContextIsProvided() {
+        Options o = new Options.Builder().server("nats://localhost").build();
+        assertNull(o.getSslContext());
+        o = new Options.Builder().server("ws://localhost").build();
```
brimworks, line 717: Open question here: Do we want to support non-SSL websockets? In other words, if the url begins with "ws://" do we want to disable TLS? As it is now "ws://" is synonymous with "wss://"... and do we want to support non-secure "ws://" which is "upgrade"d to TLS using the NATS upgrade to secure mechanism?

brimworks (approved):
Thanks for moving this along! I responded to one of the questions in the Java comments, made a recommendation to be explicit in the character set, and I have an open question regarding supporting non-secure websockets (and maybe supporting post-connect upgrade to TLS)? I'm not sure if the post-connect upgrade to TLS is supported in the NATS server, and for our use-case we just relied on the HTTP TLS (aka "wss://" protocol vs "ws://").

src/main/java/io/nats/client/support/WebSocket.java
scottf: Fixed

src/test/java/io/nats/client/OptionsTests.java
scottf: @ColinSullivan1 @RichardHightower Any opinions?

PR #832, opened 2023-01-22
Add Options.builder() method to get the builder in a more "fluent" manner

nwseOOO:
No much to explain to be honest.

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -566,23 +566,24 @@ public class Options {
     }
 
     /**
-     * Creates a builder for the options.
-     * @return the builder.
-     */
-    public static Builder newBuilder() {
-        return new Builder();
-    }
-
-    /**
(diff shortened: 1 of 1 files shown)


src/main/java/io/nats/client/Options.java
```diff
@@ -565,6 +565,14 @@ public Thread newThread(Runnable r) {
         }
     }
 
+    /**
+     * Creates a builder for the options.
+     * @return the builder.
+     */
+    public static Builder newBuilder() {
```
scottf, line 572: 1. To be consistent with all the other builders, please name this method just `builder`. 
2. Can you move it below the comments for the builder, just before the class declaration.
3. While you are there,  can you remove the comment (line 582 in this code) `is equivalent to calling {@link Nats#connect() Nats.connect()}.` This isn't right
4. Last, can you please sign your commits. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

scottf (changes requested):

PR #834, opened 2023-01-24
service subject in responses must contain group path

scottf:


arondi (approved):
LGTM, nitpick change EndpointResponse variable name in tests `es` to `er`

PR #847, opened 2023-02-11
ServerPool including Hostname resolution

scottf:
Created ServerPool / NatsServerPool instead of ServerListProvider  (still experimental)


src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -162,39 +167,71 @@ void connect(boolean reconnectOnConnect) throws InterruptedException, IOExceptio
...
+        List<NatsUri> serversToTry = getServersToTry();
+        List<NatsUri> resovledToTry = new ArrayList<>();
+        boolean notDoneTrying = true;
+        int ixTry = -1;
+        int szTry = serversToTry.size();
+        while (notDoneTrying && ++ixTry < szTry) {
+
+            // let server list provider resolve hostnames
+            NatsUri nuri = serversToTry.get(ixTry);
+            resovledToTry.clear();
```
wallyqs, line 179:
```suggestion
            resolvedToTry.clear();

```

Issue #860, opened 2023-03-18
Add/remove ConnectionListeners after a Connection is constructed.


davidmcote:
## Feature Request
I would like to add connection-state listeners to Connections that have already been constructed and are being passed into my business logic from elsewhere in the application.

#### Use Case:
Late and multiple bindings of ConnectionListeners to Connections for business logic to listen for events without needing to construct its own Connection object.

#### Proposed Change:
```
Connection.add/remove(ConnectionListener)
```

#### Who Benefits From The Change(s)?
Applications which separate their NATS Connection factory from their business logic, but whose business logic's resource's lifecycle are coupled to the NATS Connection's.

#### Alternative Approaches
If business logic is interested in tearing down resources when its Connection is closed, it must either:
* Construct the Connection itself with an appropriate ConnectionListener
* Poll the Connection's status
* Accept a custom observable in addition to a Connection that fires when the Connection is closed. (Eg: A ConnectionListener that resolves a CompleteableFuture on CLOSE.)

scottf:
Again good idea here, but low priority, contributions are welcome.

davidmcote:
Thanks Scott!  I've proposed this change in #861

scottf:
Merged, closing

PR #861, opened 2023-03-20
Enable add/remove of multiple ConnectionListeners per Connection

davidmcote:
This patch addresses #860.

Late and multiple bindings of ConnectionListeners to Connections enable business logic to listen for events without needing to construct their own Connection object.

This benefits applications which separate their NATS Connection factory from their business logic, but whose business logic's resource's lifecycle are coupled to the state of NATS Connection's.  Eg: They wish to clean up on connection CLOSED events without polling or passing around a separate observable mechanism.

### DIFF
--- a/src/main/java/io/nats/client/Connection.java
+++ b/src/main/java/io/nats/client/Connection.java
@@ -405,6 +405,24 @@ public interface Connection extends AutoCloseable {
     void closeDispatcher(Dispatcher dispatcher);
 
     /**
+     * Attach another ConnectionListener. 
+     * 
+     * <p>The ConnectionListener will only receive Connection events arriving after it has been attached.  When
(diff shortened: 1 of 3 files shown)


davidmcote:
Is it typical to see a subset of the unit test suite failing intermittently (possible unsuccessful connections to a real nats-server process)?  I couldn't seem to get a clean test run on my system before setting out to develop this patch.

My testing here focused on ensuring ConnectionListenerTest remained 100% passing.

scottf:
> Is it typical to see a subset of the unit test suite failing intermittently with issues launching nats-server processes?
Yes. We could verify the server is ready I suppose...

scottf (approved):
LGTM

davidmcote:
Thanks for the review Scott :)

PR #870, opened 2023-03-25
pull handling batch complete

scottf:


arondi (approved):
Looks Good. Like the use of constants, the improved tests, and the reduced code.

Issue #859, opened 2023-03-18
Option.Builder constructor than can pre-populate from an existing Options object


davidmcote:
## Feature Request

Some "*Option.Builders" classes are lacking a constructor that accepts the immutable "*Options" object for pre-populating fields. 

For example, such a constructor can be found in https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/FeatureOptions.java#L52, but no such constructor exists in the main Options.java object used to configure a NATS Connection.

All configuration Builder classes should be copy-constructable from the object they build.

#### Use Case:
1. Constructing Options in multiple stages without passing around a mutable builder.
2. Constructing multiple variations of the same shared constant base Options object.

Eg:
```
public static final Options COMMON_OPTIONS = Options
    .builder()
    ...
    .build();

public Options getApp1Options(String serverUrl) {
    return Options
        .builder(COMMON_OPTIONS)
        .server(serverUrl)
        .build()
}

public Options getApp2Options() {
    return Options
        .builder(COMMON_OPTIONS)
        .connectionListener(aDifferentConnectionListener)
        .build()
}

public Options getApp3Options() {
    return Options
        .builder(COMMON_OPTIONS)
        .alternateValue(123)
        .build()
}

public Options getApp4Options() {
    return Options
        .builder(COMMON_OPTIONS)
        .alternateValue(456)
        .build()
}
```

#### Proposed Change:

In any *Options class missing this pattern, add:
```
class Options {
    public static Options.Builder builder(final Options options) {
        return new Options.Builder(options);
    }

    . . .

    public static class Builder {
        public Builder(final Options options) {
            ... copy from options
        }
    }
}
```

Eg: https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/Options.java and https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/PublishOptions.java

#### Who Benefits From The Change(s)?
Application developers with multiple NATS client components which share only *most* of the information in Options classes.

#### Alternative Approaches
Alternative approaches for application developers who wish to configure Options in separate stages:
1. Wrapper OptionsFactories that return new instances of mutable Builder objects for modification.
2. Or a mirror implementation of Options & Options.Builder classes where staged modification is possible.
3. Implement their own copy-to-Builder utility method and keep it up to date with every field in Options.

Alternative library approaches:
1. Options.Builder(Options.Builder) copy constructor instead of accepting the immutable built class.
[...]

PR #875, opened 2023-03-30
Options Builder constructor that takes existing Options

scottf:
Addresses Issue #859

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -55,6 +55,7 @@ public class Options {
     // BUILD CONSTRUCTOR PROPS * update build props constructor to read new props
     // BUILDER METHODS * add a chainable method in builder for new variable
     // BUILD IMPL * update build() implementation if needed
+    // BUILDER COPY CONSTRUCTOR * update builder constructor to ensure new variables are set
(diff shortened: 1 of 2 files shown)


src/main/java/io/nats/client/Options.java
```diff
@@ -55,6 +55,7 @@ public class Options {
     // BUILD CONSTRUCTOR PROPS * update build props constructor to read new props
     // BUILDER METHODS * add a chainable method in builder for new variable
     // BUILD IMPL * update build() implementation if needed
+    // BUILDER COPY CONSTRUCTOR * update builder constructor to ensure new variables are set
```
arondi, line 58: Thanks for not forgetting this.

src/main/java/io/nats/client/Options.java
```diff
@@ -601,7 +596,7 @@ public static class Builder {
         // ----------------------------------------------------------------------------------------------------
         // BUILDER VARIABLES
         // ----------------------------------------------------------------------------------------------------
-        private final List<NatsUri> natsUris = new ArrayList<>();
+        private final List<NatsUri> natsServerUris = new ArrayList<>();
```
arondi, line 599: can you change the builder instead of the options?

arondi (changes requested):

PR #878, opened 2023-04-04
Time used fetching should never be negative

liu-jianyang:
If device system time changes into the past while fetching from nats, we should not allow the difference in time to be negative. This would cause the call to nextMessageInternal to hang for a long time.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
+++ b/src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
@@ -149,7 +149,7 @@ public class NatsJetStreamPullSubscription extends NatsJetStreamSubscription {
                     batchLeft--;
                 }
                 // try again while we have time
-                timeLeft = maxWaitMillis - (System.currentTimeMillis() - start);
+                timeLeft = maxWaitMillis - Math.max(0, System.currentTimeMillis() - start);
             }
         }
         catch (InterruptedException e) {


src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
```diff
@@ -149,7 +149,7 @@ private List<Message> _fetch(int batchSize, long maxWaitMillis) {
                     batchLeft--;
                 }
                 // try again while we have time
-                timeLeft = maxWaitMillis - (System.currentTimeMillis() - start);
+                timeLeft = maxWaitMillis - Math.max(0, System.currentTimeMillis() - start);
```
scottf, line 152: The right side will always be zero, which also doesn't work, timeLeft will never decrease below maxWaitMillis. I'm working on a solution.

scottf (changes requested):

src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
scottf: This is my proposal. 
```
timeLeft = maxWaitMillis - Math.abs(System.currentTimeMillis() - start);
```

scottf: The `Math.abs` solution will make `timeLeft` < 0 if the clock changes backwards, stopping the loop. Messages that are missed will get put in a buffer and dealt with the next time fetch is called again, see lines 123-127. The new underlying pull will be adjusted.
Just a note regarding what we are calling simplification. I'm making a new "fetch" (really an iterate, it does not block and also making a pull based consumer that is endless, which means it looks like a push subscription but uses pull under the covers. These may or may not be of interest to you, either way I should have a beta by end of week.

liu-jianyang: Yeah, definitely not ideal. Alternatively we could reset and let the user handle it? Since having the current time be less than the original start time is invalid anyway.

scottf: The Math.abs version will do the trick. If the current time is less than start, the loop will stop. If the time changes forward, the right side is still valid and makes timeLeft less than 0, also stopping.

If you would like to make the changes, please go ahead, otherwise I will make a PR. If you can, can you please verify your commit. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

scottf (approved):
Approved

src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
pramodsri: Should we not move away from using System.currentTimeMillis() and use something like System.nanoTime()?

PR #879, opened 2023-04-05
Pull-fetch-time-left

liu-jianyang:
Update to use System.nanoTime() instead of System.currentTimeMillis() to move away from wall clock.


src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
```diff
@@ -149,7 +150,7 @@ private List<Message> _fetch(int batchSize, long maxWaitMillis) {
                     batchLeft--;
                 }
                 // try again while we have time
-                timeLeft = maxWaitMillis - (System.currentTimeMillis() - start);
+                timeLeft = maxWaitMillis - TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start);
```
scottf, line 153: What do you think about making everything nanos?
```
long maxWaitNanos = maxWaitMillis * 1_000_000; // is million right?
...
long timeLeft = maxWaitNanos;
...
Message msg = nextMessageInternal( Duration.ofNanos(timeLeft) );
...
timeLeft = maxWaitNanos - (System.nanoTime() - start);
```

Also, I really need you to verify your commits. It's an CNCF rule, I'm allowed to skip it on occasion, but I'd prefer not to.

scottf (changes requested):

src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
liu-jianyang: Ah gotcha. I was just using the github workspace to make these changes. Let me look into how to do it. I'll ask for the review once again after I figure it out.

liu-jianyang: Yeah, nanos would be better here too, since the `poll` method we use in MessageQueue uses nanos.

scottf (approved):
LGTM

src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
```diff
@@ -138,9 +138,10 @@ private List<Message> _fetch(int batchSize, long maxWaitMillis) {
             // timeout > 0 process as many messages we can in that time period
             // If we get a message that either manager handles, we try again, but
             // with a shorter timeout based on what we already used up
-            long timeLeft = maxWaitMillis;
-            while (batchLeft > 0 && timeLeft > 0) {
-                Message msg = nextMessageInternal( Duration.ofMillis(timeLeft) );
+            long maxWaitNanos = maxWaitMillis * 1_000_000;
+            long timeLeftNanos = maxWaitNanos;
```
PramodSSImmaneni, line 142: Shouldn't timeLeftNanos be initialized to maxWaitNanos - (System.nanoTime() - start) as well?

liu-jianyang: I think the reasoning here is that we want to at least allow processing a single message to send back, which could no longer be the case if `timeLeftNanos` is set to `maxWaitNanos - (System.nanoTime() - start)` since we could set `maxWaitNanos` really small.

PR #887, opened 2023-04-11
New connection option reportNoResponders

scottf:
Adding a `reportNoResponders` connection option flag. Will make core no responders exception differently when core-publishing to a subject.

This should really be the default behavior, but it would be a breaking change.

```
java.util.concurrent.CancellationException
```
to this
```
java.util.concurrent.ExecutionException: io.nats.client.JetStreamStatusException: 503 No Responders Available For Request
```

### DIFF
--- a/src/examples/java/io/nats/examples/DontCancelOnNoResponders.java
+++ b/src/examples/java/io/nats/examples/DontCancelOnNoResponders.java
@@ -41,7 +41,7 @@ public class DontCancelOnNoResponders {
 
         // new option
         options = new Options.Builder().server(Options.DEFAULT_URL)
-            .dontCancelNoResponders()
+            .reportNoResponders()
             .build();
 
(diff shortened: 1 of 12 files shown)


tbeets (changes requested):
Suggest phrasing connection option in the positive, i.e. `reportNoResponders` rather than the negative.

tbeets (approved):
LGTM

src/examples/java/io/nats/examples/DontCancelOnNoResponders.java
```diff
@@ -0,0 +1,60 @@
+// Copyright 2020 The NATS Authors
```
ColinSullivan1, line 1: Might want to update the copyright date while you're editing this file.

ColinSullivan1 (approved):
LGTM.

PR #891, opened 2023-04-17
Http Request Version Default To 1.1

aditya-agarwal-groww:


scottf:
@brimworks Does this seem like a reasonable change?

brimworks:
I don't think this client supports chunked encoding and thus it really doesn't support http 1.1. Maybe the request filtering mechanism could be modified to support changes to the http version so individual usage of this library could change the version?

brimworks:
Here is a nice summary of http 1.1 clients: https://www2.cs.uh.edu/~gnawali/courses/cosc6377-f12/p1/http.html#http1.1clients Technical these requirements are not met, which is why I don't think we should change the http version (by default).

aditya-agarwal-groww:
@brimworks For Websocket connections we need to have Http 1.1 , should we then use http 1.1 for that particular use case ?

scottf:
@brimworks Thanks for stepping in here. @aditya-agarwal-groww I'll merge it when the build completes and that will build the next 2.16.11-SNAPSHOT. I'll try to do a full release within the next 2 weeks.

aditya-agarwal-groww:
@scottf Sure, thanks a lot !!

aditya-agarwal-groww:
hey @scottf for some reason coverage/coveralls job  is still pending

scottf:
> hey @scottf for some reason coverage/coveralls job is still pending

Coverage doesn't work for forks. I'll be merging this, I've just been out of office.

PR #898, opened 2023-04-27
Allow configuring a message handler to be executed before a message is published

cpiotr:
Timestamping messages in client code yields spurious results, because there is internal NATS queuing involved. Consequently, any GC or safepoint pause which happens between the time a message is enqueued and then dequeued gets incorporated into the delay attributed to NATS.

This commit introduces a custom message handler to be executed before a message is published. This allows reducing the time gap currently assigned to the message broker.


scottf:
Can you tell us more about the use case? 

If it's possible that the message changes, the protocol and message size will have to be recalculated, which will take time. At a minimum, this will need to be done in `accumulate` function, which happens before the sendMessageBatch, which adds some time also. Maybe nanos, but if I'm understanding correctly, you want to mark the message at the last possible moment.

And this recalculation would have to check whether there was a change or always recalculate, if we move the size calculation to the last thing, it's still problematic because I've already tracked that message size.

Is this for JetStream, b/c if it is, then you should rely on the server timestamp as the source of truth. The timestamp is available in the meta data that comes to the subscriber with the message.

scottf:
I don't think there is much point in putting this in. There are going to be variations in timing that have nothing to do with the client library. For instance the network differences between message publishers to the server, differing cpu loads affecting speed on machines including the servers, especially in containerized configurations, GC like you said but also on the consumer side, differences in actual machine clocks (yes, even though they should all be the same, we're talking nanos or milliseconds).

Issue #900, opened 2023-05-04
WebSocketInputStream incorrectly handles network buffers with multiple WebSocket Frames


GrafBlutwurst:
## Defect
 - [ ] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve) (Sadly not applicable since it's non-deterministic, depending on network delivery by server)

The problem is in WebsocketInputStream, in particular here 

```
public int read(byte[] buffer, int offset, int length) throws IOException {
        // Just in case we get headers with empty payloads:
        while (0 == header.getPayloadLength()) {
            if (!readHeader()) {
                return -1;
            }
        }
        if (header.getOpCode() == OpCode.CLOSE) {
            // Ignore the websocket close message body:
            in.skip(header.getPayloadLength());
            return -1;
        }
        length = in.read(buffer, offset, length);
        if (-1 == length) {
            return length;
        }
        return header.filterPayload(buffer, offset, length);
    }
```

This code gets called by NatsConnectionReader with some length `n` and offset 0.

1. we preread the websocket header in the first while loop 
2. assume it's not CLOSE
3. we then read from the network stream at maximum `n` bytes. Assume at this point a second websocket frame has arrived so `buffer will look something like 
```
PROTOCOL LINE
PAYLOAD
<WebSocketFrameHeader Bytes> PROTOCOL LINE
PAYLOAD
```
Note that the first websocket frame header has already been stripped by the initial while loop
4. in `filterPayload` we'll return the bytes read based on the payload length of the first websocket frame header
5. NatsConnectionReader  processes the first message but never the second because the number of bytes read returned by WebSocketInputStream.read will be the number of bytes of the first websocket frames payload
6. NatsConnectionReader calls `read` on WebSocketInputStream again. At this point the we have skipped over the second Frame that was put into the buffer in the initieal `read`

Pertinent Code references: 

https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/support/WebsocketInputStream.java#L57
https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/impl/NatsConnectionReader.java#L122

### Fix

In WebsocketInputStream we'll have to deal with the possibilities that our network read may contain multiple websocket frames, correctly deal with all involved headers and return the sum of the payload lengths as the number of bytes read. I'll get started on a fix but it may take a while because I'm not very familiar with the internals of this codebase. Any help would be appriciated.  

[...]

PR #902, opened 2023-05-04
only read payload from network stream

GrafBlutwurst:
Addresses: https://github.com/nats-io/nats.java/issues/900

Might need a regression test?


brimworks (approved):
Good catch, I am not sure how I missed this. Sorry!

GrafBlutwurst:
I see some coverage check that hasn't completed yet. Do I need to trigger that manually? Also I'm not currently set up for commit signing, or is that done by the maintainer team when merging?

scottf:
@GrafBlutwurst Coverage doesn't work for forks which explains the missing check. We prefer signed commits, so for the future please set that up, but it's fine for this PR. Thanks for the contribution!

PR #905, opened 2023-05-08
Bump GH Actions versions (Go; node12 deprecation)

philpennock:
Bump various GitHub Actions versions to move away from those using the deprecated node.js 12 runtime, and so remove the warnings.

Bump Golang to the latest security fixed patch-level within the same minor, and quote the string for defensiveness against version number parsing and YAML string vs version numbers.  While we're unlikely to switch to patch-level-unspecified in future, protect against mistakes by quoting, as we're doing for all repos.


philpennock:
Closing in favor of the maintainer's PR #907

scottf (approved):
LGTM

PR #913, opened 2023-05-15
Fix hostname resolution for proxied connections

bartosz822:
Proxied connections resolve the hostname to an IP address while tunneling even when noResolveHostname option is set. 
This behavior is unwanted in some scenarios, especially when a proxy does whitelisting based on hostnames.
This pr should fix this.


scottf:
@bartosz822 I'm inclined to approve this, but first...
1. Should we differentiate between proxied and isNoResolveHostnames. 
2. Can you provide a small code snippet / output that shows the difference from different input?
3. How does this work with discovered servers, which usually come in the form of an ip address? Probably no effect, but I just want to think about it.

scottf:
@bartosz822 Just waiting for another pair of eyes on this.

scottf:
@bartosz822 I'm closing this, it's redundant. If you trace back, you can see it's already done before it gets here.

```
io.nats.client.impl.SocketDataPort.connect(SocketDataPort.java:62)
io.nats.client.impl.SocketDataPort.connect(SocketDataPort.java:52)
io.nats.client.impl.NatsConnection.tryToConnect(NatsConnection.java:425)
io.nats.client.impl.NatsConnection.connect(NatsConnection.java:207)
io.nats.client.impl.NatsImpl.createConnection(NatsImpl.java:29)
io.nats.client.Nats.createConnection(Nats.java:303)
io.nats.client.Nats.connect(Nats.java:210)
```

Look at the code around NatsConnection, line 207. 
NatsConnection line 194 calls `resolveHost` (NatsConnection, line 1759)
NatsConnection, line 1762 checks for an ip address 
NatsConnection, line 1763 is `serverPool.resolveHostToIps` (NatsServerPool, line 188)
NatsServerPool resolveHostToIps checks the flag `isNoResolveHostnames`

bartosz822:
@scottf I've seen the code you've mentioned and this pull request is not redundant, could you please reopen it? I've observed in proxy logs using this library that the  connection was made using the IP address even though `isNoResolveHostnames` flag was set. The `NatsConnection` that is passed to the `connect` method is indeed unresolved, but the constructor of `InetSocketAddress` does the resolution under the hood, look at the code from `java.net`: 

```    
public InetSocketAddress(String hostname, int port) {
        checkHost(hostname);
        InetAddress addr = null;
        String host = null;
        try {
            addr = InetAddress.getByName(hostname);
        } catch(UnknownHostException e) {
            host = hostname;
        }
        holder = new InetSocketAddressHolder(host, addr, checkPort(port));
    }```

Take also a look at the implementation of `java.net.Socket#connect(java.net.SocketAddress, int)`, it checks if the `SocketAddress` is resolved or not.

**Comment by scottf** [comment]
What was the input, meaning the servers specified in the options? You saw the stack trace. Also, if there is any resolving or checking to be done, it needs to be done before SocketDataPort.connect. I suppose it's possible that the SocketDataPort may need to be changed to never resolve as it should take exactly what it was given

**Comment by bartosz822** [comment]
@scottf I haven't received any new feedback on this PR since quite a while. Is it ok? Is there something else that needs to be done?

**Comment by GrafBlutwurst** [comment]
Hi, I was curious how we're doing on this MR? The reason that I'm asking is that we currently have to maintain some firewall rules using IP addresses over domain names which the colleagues from security don't like much.

**Comment by scottf** [comment]
> Hi, I was curious how we're doing on this MR? The reason that I'm asking is that we currently have to maintain some firewall rules using IP addresses over domain names which the colleagues from security don't like much.

It's been quite a while since I reviewed this but I remember it was incomplete. Resolve hostnames is on by default, but you can turn it off in options via the options builder by calling noResolveHostnames()

**Comment by GrafBlutwurst** [comment]
Looking at the description it seems that the `noResolveHostnames` doesn't work for proxied connections. We observe the same still or was this fixed in another MR and I just missed it?

**Comment by scottf** [comment]
@GrafBlutwurst I think the problem is legitimate, just not the fix in this PR. Also, why are you proxying connections?

**Comment by scottf** [comment]
@GrafBlutwurst I really wanted to close this but the user insisted I was not understanding the problem. Which may be true. Is it possible that you can open a fresh defect issue in the repo and we can try to start this process over? I'm certain, I was not happy with the proposed code change, but frankly it's been a while and I can't remember why.

**Comment by GrafBlutwurst** [comment]
@scottf Here's the defect issue. https://github.com/nats-io/nats.java/issues/967 

Let me know if I can help in any way. Looking at the code it's also how I would have solved the issue I think :/ , but maybe you have a better idea.

**Comment by scottf** [comment]
> @scottf Here's the defect issue. #967
> 
> Let me know if I can help in any way. Looking at the code it's also how I would have solved the issue I think :/ , but maybe you have a better idea.

If I remember correctly, by the time it got to that point, it should have already been resolved. So it needs to be traced back and determined where it should happen.

**Comment by GrafBlutwurst** [comment]
I think the point here is that in a proxied case with firewalls we never want the hostname to be resolved to an IP on the client. Connections should always be established by hostname so that we don't need 2 firewall rules. 

In particular the constructor call `new InetSocketAddress(host, port)` does resolution underneath thus on the firewall this connection attempt is by IP. Thus necessitating the use of `InetSocketAddress.createUnresolved(host, port)`.

At least that was my understanding when looking at the code and would address our issue with hostname resolution.

**Comment by scottf** [comment]
@GrafBlutwurst Is this still an issue? Recent versions of the client support proxies much better, addressing the fact that the client to proxy is secure, but commonly proxy to server is insecure. Also there is the Options.noResolveHostnames option.
To setup the client for proxy, see this: https://github.com/nats-io/nats.java?tab=readme-ov-file#reverse-proxy

File: src/main/java/io/nats/client/impl/SocketDataPort.java [hunk 1]
```diff
@@ -74,7 +74,9 @@ public void connect(@NonNull NatsConnection conn, @NonNull NatsUri nuri, long ti
                 socket = connectToFastestIp(options, host, port, (int) timeout);
             } else {
                 socket = createSocket(options);
-                socket.connect(new InetSocketAddress(host, port), (int) timeout);
+                InetSocketAddress inetSocketAddress = (options.isNoResolveHostnames() && !nuri.hostIsIpAddress()) ?
+                        InetSocketAddress.createUnresolved(host, port) : new InetSocketAddress(host, port);
+                socket.connect(inetSocketAddress, (int) timeout);
```
scottf, line 84: Also, resolution is supposed to be done before it gets here. Look in the NatsConnection line 540 and work backward. Also look in NatsServerPool, where it does resolveHostToIps. 
Resolution does not belong here, it belongs further back up the call chain.

bartosz822: I agree that the resolution probably belongs further back up the call chain. However the problem is that even if it didn't happen earlier, the default constructor of InetSocketAddress will perform it.

scottf: Would it be correct to say that this is only an issue in some cases AND only when a proxy is provided?

Why can't the proxy be expected to do this work?

bartosz822: Yes it is an issue only in some cases, it only really is an issue if a proxy is doing whitelisting based on domain names, which is a common use-case in secure/enterprise environments.

scottf: So the question remains, why can't this be part of the Option.proxy's responsibility? If it can be, we can simply add documentation to that effect. 

I'm suspicious how common this really is in secure/enterprise environments. Enterprises are the most common users of this client, it's been around for 7 years, and this is the first time I'm hearing about the issue. 

At a minimum we need an opt in solution to ensure backward compatibility, like `isEnableInetAddressCreateUnresolved` or something like that.

```
if (options.isEnableFastFallback()) {
    socket = connectToFastestIp(options, host, port, (int) timeout);
} 
else {
    InetSocketAddress inetSocketAddress;
    if (options.isEnableInetAddressUnresolved() && !nuri.hostIsIpAddress()) {
        inetSocketAddress = InetSocketAddress.createUnresolved(host, port);
    }
    else {
        inetSocketAddress = new InetSocketAddress(host, port);
    }
    socket.connect(inetSocketAddress, (int) timeout);
} 
```

scottf: @bartosz822 bump

bartosz822: Added the suggested option.

I don't understand how it could be part of the Option.proxy responsibility? What do you mean by that?

scottf: The proxy seems to have some flexibility. You test has a proxy implementation. It seems like code could just go in there. But at least there is a backward compatible way now, so I think that's fine.

bartosz822: Ah, in the test I have flexibility, but in the real world the proxy is an external server managed by different people. 

Yes the workaround for that issue currently for us is to ask to whitelist an ip address in the proxy configuration, however that's not a great solution since IPs can change.

bartosz822:
> @bartosz822 The unit test is failing

Something is off because that test works for me locally, I'll need to take a closer look at what's happening

bartosz822:
Hey @scottf, sorry I totally forgot about this PR.

To address your last comment:

  1. enableInetAddressCreateUnresolved now auto-sets noResolveHostnames — so the two flags stay consistent. Kept the name as-is since it describes exactly what it does at the SocketDataPort level.
  2. Made another attempt to fix the unit test on different OSes, if it still doesn't work I will remove it.
  3. Added Options getter/setter test in OptionsTests
  4. Regarding NatsInetAddressProvider — createUnresolved is specifically about not resolving, so it doesn't fit into the address provider whose purpose is to provide resolved addresses. The fix is intentionally scoped to SocketDataPort.connect() where the InetSocketAddress constructor was doing unwanted resolution.

PR #930, opened 2023-06-21
[BUG] Consumer Config fields are 32 bits (ints) not 64 bits (longs)

scottf:
Building a consumer config allowed the user to supply a long for fields that where only 32 bit signed values. Even though it would be unusual for a developer to provide a value larger than 2,147,483,647 it's still possible from the builder interface.
This fix checks the value and if it's larger than 2,147,483,647 it just makes it 2,147,483,647. 
This seems reasonable as the user probably meant unlimited which this still provides.


aricart (approved):
LGTM

PR #931, opened 2023-06-21
Address secure Websocket changes in server 2.9.19

scottf:
This change is backward compatible when the server is older than 2.9.19


wallyqs (approved):
LGTM

PR #932, opened 2023-06-22
Better testing checks against server versions

scottf:
* Improved version check utils for use in regression testing to ensure features are tested properly against server versions.
* Fixed test running against 2.10 dev


piotrpio (approved):
LGTM

PR #936, opened 2023-06-30
Simplified Ordered Consumer

scottf:


src/main/java/io/nats/client/api/OrderedConsumerConfig.java
```diff
@@ -0,0 +1,169 @@
...
+            return this;
+        }
+
+        /**
+         * sets the max amount of expire time for the server to allow on pull requests.
+         *
+         * @param maxExpires the max expire duration
+         * @return Builder
+         */
+        public Builder maxExpires(Duration maxExpires) {
```
piotrpio, line 111: I am wondering whether it makes sense to allow setting `maxExpires`, `maxBytes` and `maxBatch` for ordered consumers. After all, those values will be set in consume configuration and ordered consumer is not meant to be used outside of the application. So does it make sense to set limits on itself when it's always the same process that will use it?

src/main/java/io/nats/client/api/OrderedConsumerConfig.java
```diff
@@ -0,0 +1,169 @@
...
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * OrderedConsumerConfig can be created using a Builder.
+     * The builder supports chaining and will create a default set of options if
+     * no methods are calls, including setting the filter subject to "&gt;"
+     */
+    public static class Builder {
```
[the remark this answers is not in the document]
scottf: It's set somewhere else.
```
if (ordered) {
...
    ConsumerConfiguration.Builder b = ConsumerConfiguration.builder(builder.cc)
                .ackPolicy(AckPolicy.None)
                .maxDeliver(1)
                .ackWait(Duration.ofHours(22))
                .name(name)
                .memStorage(true)
                .numReplicas(1);
```

src/main/java/io/nats/client/api/OrderedConsumerConfig.java
scottf: @piotrpio Not sure. All the other simplification allows them, they are limits. The user still can set batch size messages or bytes. I suppose I could interrogate that before I make the final configuration.

src/main/java/io/nats/client/api/OrderedConsumerConfig.java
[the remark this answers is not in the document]
piotrpio: All right, and then you just reconfigure the config when recreating consumer, I missed where it's being set 👍

src/main/java/io/nats/client/api/OrderedConsumerConfig.java
piotrpio: Yes, user can set batch sizes etc on `Consume`, I was just wondering what's the point of setting those limits if only your application will be setting the values in `Consume`. It certainly does not hurt to give those options, I just don't see the reason to ever use them.

scottf: I guess this needs discussion / adr?

aricart: why not just let it be what the server defaults?

aricart: Ordered consumer should have the min set of options possible.

src/main/java/io/nats/client/support/MiscUtils.java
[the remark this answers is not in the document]
scottf: It's either that or constantly having 1 static util method in a class. Like RandomUtils.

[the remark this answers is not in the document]
scottf: Changed

piotrpio (approved):
LGTM!

PR #946, opened 2023-07-23
Improving Options with better construction via properties

scottf:
* Allow creation of connections requiring an AuthHandler for JWT to specify the credentials file in a properties files, instead of needing to provide an instance of AuthHandler in code.
* Allow creation of connections requiring an SSL context to specify key and trust store information in a properties files so an SSLContext can be created automatically instead of needing to provide an instance of an SSLContext in code.
* Support properties keys with or without the prefix 'io.nats.client.'

| Property| Description|
| --- | --- |
| credential.path | Used to set the path to a credentials file to be used in a FileAuthHandler |
| keyStore | Property for the keystore path used to create an SSLContext |
| trustStore | Property for the truststore path used to create an SSLContext |
| keyStorePassword | Property for the keystore password used to create an SSLContext |
| trustStorePassword | Property for the truststore password used to create an SSLContext |
| tls.algorithm | Property for the algorithm used to create an SSLContext |

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ The Services Framework further streamlines their development by providing observ
 
 Check out the [ServiceExample](src/examples/java/io/nats/examples/service/ServiceExample.java)
 
-## Versions Specific Notes
+### Versions Specific Notes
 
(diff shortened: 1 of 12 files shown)


ColinSullivan1:
Looks good, really like the per-connection properties here. A minor suggestion would be the ability to build the with options and override.  E.g. to set the props file in the builder via method and also be able to set a name, reconnect tries, etc.  Best of both worlds.

PR #964, opened 2023-09-01
Add StatisticsCollector to Options for custom statistics tracking

### LINKED ISSUE #960: Add a hook into the Statistics interface to observe counter changes
### What motivated this proposal?

I want to track connection statistics in Prometheus. Since the `Statistics` class only provides counter values, I can't directly increment a prometheus counter from the statistics object.

### What is the proposed change?

Provide a way on the `Statistics` interface to register an observer for each of the counters that can observe the counts being incremented directly. This way a Prometheus counter could be incremented in addition to the NATS in memory counters.

### Who benefits from this change?

Anyone that wants to track connection metrics in an external monitoring system.

### What alternatives have you evaluated?

I'm considering using a prometheus gauge and just documenting very clearly that it's actually a counter.

johnlcox:
Add an interface for statistics collection that can be set on `Options` to allow for custom metric collection (e.g. to prometheus).

This fixes #960


scottf (approved):
LGTM

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1231,11 +1232,12 @@ void deliverReply(Message msg) {
             statistics.incrementRepliesReceived();
         }
         else if (!oldStyle && !subject.startsWith(mainInbox)) {
-            if (advancedTracking && responsesRespondedTo.get(key) != null) {
-                statistics.incrementDuplicateRepliesReceived();
-            }
-            else {
-                statistics.incrementOrphanRepliesReceived();
+            if (advancedTracking) {
```
johnlcox, line 1235: Previously, all orphan counter was being incremented for every duplicate or orphan when `advancedTracking` was not enabled. Does this change to first check the `advancedTracking` flag and only when enabled increment duplicate or orphan replies.

scottf (approved):
LGTM

PR #971, opened 2023-09-13
Construct Options builder from properties file for user.

scottf:
Developer used to have to load the properties file themselves. Now they can just supply the file path and the Options builder will do it for them.

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -23,11 +23,14 @@ import io.nats.client.support.SSLUtils;
 
 import javax.net.ssl.SSLContext;
 import java.io.File;
+import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.net.Proxy;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


src/main/java/io/nats/client/Options.java
```diff
@@ -694,19 +697,24 @@ public Builder() {}
         // ----------------------------------------------------------------------------------------------------
         /**
          * Constructs a new {@code Builder} from a {@link Properties} object.
-         *
-         * <p>If {@link Options#PROP_SECURE PROP_SECURE} is set, the builder will
-         * try to get the default context{@link SSLContext#getDefault() getDefault()}.
-         * If a context can't be found, no context is set and an IllegalArgumentException is thrown.</p>
-         *
```
scottf, line 701: This text was moved to the build method in a previous PR

jnmoyne (approved):
LGTM

jnmoyne (approved):
LGTM

PR #982, opened 2023-09-22
Subject Transforms in Mirror/Info and Source/Info

scottf:


src/main/java/io/nats/client/api/SourceInfoBase.java
```diff
@@ -67,22 +72,19 @@ public External getExternal() {
...
      * @return the error
      */
     public Error getError() {
         return error;
     }
-
-    @Override
-    public String toString() {
-        return "Mirror{" +
-            "name='" + getName() + '\'' +
```
scottf, line 81: not needed in the base class and Mirror was wrong anyway

jnmoyne (approved):
LGTM

PR #984, opened 2023-09-26
Multiple Filter Subjects Review

scottf:


src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -116,6 +115,7 @@ protected ConsumerConfiguration(ConsumerConfiguration cc) {
         this.memStorage = cc.memStorage;
         this.backoff = cc.backoff == null ? null : new ArrayList<>(cc.backoff);
         this.metadata = cc.metadata == null ? null : new HashMap<>(cc.metadata);
+        this.filterSubjects = cc.filterSubjects == null ? null : new ArrayList<>(cc.filterSubjects);
```
scottf, line 118: This was not released. I released that I need this to be null and not an empty list because it's the only way to determine if the user sets data in the list, which they could set empty, and this is used for comparison against a server version of the config during subscribe to ensure there are no mismatches.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -158,6 +150,14 @@ protected ConsumerConfiguration(ConsumerConfiguration cc) {
...
         backoff = readNanosList(v, BACKOFF, true);
         metadata = readStringStringMap(v, METADATA);
+
+        String tempFs = emptyAsNull(readString(v, FILTER_SUBJECT));
+        if (tempFs == null) {
+            filterSubjects = readOptionalStringList(v, FILTER_SUBJECTS);
+        }
+        else {
+            filterSubjects = Collections.singletonList(tempFs);
+        }
```
scottf, line 160: moved this down a bit in the code since it is more than a one liner

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -237,6 +231,14 @@ else if (filterSubjects.size() == 1) {
...
         JsonUtils.addField(sb, MEM_STORAGE, memStorage);
         JsonUtils.addField(sb, METADATA, metadata);
+        if (filterSubjects != null) {
+            if (filterSubjects.size() > 1) {
+                JsonUtils.addStrings(sb, FILTER_SUBJECTS, filterSubjects);
+            }
+            else if (filterSubjects.size() == 1) {
+                JsonUtils.addField(sb, FILTER_SUBJECT, filterSubjects.get(0));
+            }
+        }
```
scottf, line 241: moved down, no change

src/test/java/io/nats/client/impl/JetStreamGeneralTests.java
```diff
@@ -801,107 +806,112 @@ public void testSubscribeDurableConsumerMustMatch() throws Exception {
             JetStream js = nc.jetStream();
             JetStreamManagement jsm = nc.jetStreamManagement();
 
-            createDefaultTestStream(jsm);
+            String stream = stream();
+            String subject = subject();
+            createMemoryStream(nc, stream, subject);
```
scottf, line 811: Whenever I'm in tests now, I make them less generic, meaning I give unique values to things like stream and subject instead of using a fixed constant. There is no difference in test behavior, but this will allow me to re-use server instances in the future if necessary, as sometimes resources on build machines are low, and not having to start up a server every time would failures based on inability to start servers. The change will allow allow more parallelization of tests.

src/test/java/io/nats/client/impl/JetStreamPullTests.java
```diff
@@ -293,7 +303,7 @@ public void testBasic() throws Exception {
 
     @Test
     public void testNoWait() throws Exception {
-        runInJsServer(nc -> {
+        runInJsServer(noPullWarnings(), nc -> {
```
scottf, line 306: this reduces unnecessary output on tests.

src/test/java/io/nats/client/support/ValidatorTests.java
```diff
@@ -545,55 +545,55 @@ public void testSemver() {
     }
 
     @Test
-    public void testListsAreEqual() {
+    public void testConsumerFilterSubjectsAreEquivalent() {
```
scottf, line 548: changed the name to match the method being tested, which changed it's name to clearer.

src/test/java/io/nats/client/utils/TestBase.java
```diff
@@ -120,10 +120,18 @@ public static void runInJsServer(ErrorListener el, InServerTest inServerTest) th
         runInServer(false, true, new Options.Builder().errorListener(el), null, inServerTest);
     }
 
+    public static void runInJsServer(VersionCheck vc,  ErrorListener el, InServerTest inServerTest) throws Exception {
+        runInServer(false, true, new Options.Builder().errorListener(el), vc, inServerTest);
+    }
+
```
scottf, line 126: macro to make it easier to test only if a server version is appropriate. Useful during regression tests against older servers.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
scottf: I think it's just the nature of the `? :`
Since this.filterSubjects is already null by default I could have just
```
if (cc.filterSubjects != null) {
    this.filterSubjects = new ArrayList<>(cc.filterSubjects);
}
```
If you think that's clearer, I can change it. I would do the same with the other list.
---
The reason I copy it is because I always copy list input to constructors because people have reported bugs because _they_ re-used a list or structure. Probably not necessary, just paranoid.

src/main/java/io/nats/client/support/Validator.java
```diff
@@ -44,6 +44,9 @@ public static String validateSubjectTerm(String subject, String label, boolean r
             throw new IllegalArgumentException(label + " cannot end with '.'");
         }
 
+        if (subject.contains(".>.")) {
+            int x = 0;
```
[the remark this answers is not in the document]
scottf: that was debug. I need to remove it.

piotrpio (approved):
LGTM!

PR #999, opened 2023-10-04
Allow only truststore

scottf:


jnmoyne (approved):
LGTM
May want to also update the README that says you can just pass the keystore for it to create the SSL context for you to say that it will create the SSL context for you if you specify either keystore or trust store (though it seems if you just pass the keystore then it will create the SSL context but anyways still won't be able to successfully connect to the server).

PR #1013, opened 2023-10-16
Client Compatibility Tests engine

scottf:


src/test/java/io/nats/compatibility/ObjectStoreCommand.java
```diff
@@ -0,0 +1,280 @@
...
+            respond();
+        }
+        catch (Exception e) {
+            handleException(e);
+        }
+    }
+
+    private void doGetObject() {
+        try {
+            _createBucket();
```
piotrpio, line 85: You don't need to create state for the tests - compatibility tester does it for you. So you can remove `_createBucket()` and `_putObject()` here and in other tests and just do `_getObject()`. So this test can be reduced to:
```java
        try {
            ObjectStore os = nc.objectStore(bucket);
            ByteArrayOutputStream baos = new ByteArrayOutputStream();
            _respondDigest(os.get(object, baos));
        }
```

piotrpio: Same goes for all other tests preparing initial state: `doUpdateMetadata()`, `doPutObject()` etc.

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
```diff
@@ -0,0 +1,280 @@
...
+            osm = nc.objectStoreManagement();
+        }
+        catch (IOException e) {
+            handleException(e);
+            return;
+        }
+
+        ObjectStoreConfiguration osc = extractObjectStoreConfiguration();
+        try {
+            osm.delete(osc.getBucketName());
```
piotrpio, line 235: You don't need that here - compat tester handles the cleanup for you.

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
```diff
@@ -0,0 +1,280 @@
...
+                result = oi.getDigest();
+                latch.countDown();
+            }
+        }
+
+        @Override
+        public void endOfData() {}
+    }
+
+    protected void _respondDigest(ObjectInfo oi) {
```
piotrpio, line 204: I don't think this validates what it's supposed to validate. You're taking digest from the object you got and just returning it back, so it will match whatever was created by the compatibility tester. What you should do is calculate the SHA for the object you got and return the SHA, not the digest from `ObjectInfo`.

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
```diff
@@ -0,0 +1,280 @@
...
+        Log.info("RESPOND " + subject + " digest " + digest);
+        byte[] payload = Base64.getUrlDecoder().decode(digest.substring(8));
+        nc.publish(replyTo, payload);
+    }
+
+    private void _putObject(String objectName, String description) {
+        try {
+            ObjectStore os = nc.objectStore(bucket);
+            ObjectMeta meta = ObjectMeta.builder(objectName).description(description).build();
+            InputStream inputStream = Utility.getFileAsInputStream("nats-server.zip");
```
piotrpio, line 215: I don't think you should keep `nats-server.zip` in the reposiotry - compat tester gives you the URL to the thing you should put in the object store. You can simply download whatever's in the received config and use this instead. That way you are not risking having the test fails if at some point the url in config changes (for whatever reason).

src/test/resources/data/object_store_file.dat
piotrpio, line 1: This should be removed.

src/test/resources/data/nats-server.zip
piotrpio, line 1: As I said in the other comment, I think it's safer to download from the provided url when needed.

src/test/java/io/nats/compatibility/Log.java
```diff
@@ -0,0 +1,121 @@
...
+        if (PRINT_THREAD_ID) {
+            String tn = Thread.currentThread().getName().replace("pool-", "p").replace("-thread-", "t");
+            start = "[" + tn + "@" + time() + "] " + label;
+        }
+        else {
+            start = "[" + time() + "] " + label;
+        }
+
+        if (extraStr == null || extraStr.isEmpty()) {
+//            if (error) {
```
piotrpio, line 84: To remove.

src/test/java/io/nats/compatibility/Log.java
```diff
@@ -0,0 +1,121 @@
...
+//                System.err.println(start);
+//            }
+//            else {
+                System.out.println(start);
+//            }
+            return;
+        }
+
+        extraStr = NEWLINE_INDENT + extraStr.replaceAll("\n", NEWLINE_INDENT);
+//        if (error) {
```
piotrpio, line 94: To remove.

piotrpio (changes requested):

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
scottf: 1. Will Fix.
2. We should just have a test that calculates a digest.

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
scottf: What's the difference between that and hard-coding bucket names in the default test?

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
scottf: As the client gets an object, it calculate the digest and checks that it matches what comes from the server. If the value does not match, I throw an exception.
```
if (totalBytes != oi.getSize()) { throw OsGetSizeMismatch.instance(); }
if (totalChunks != oi.getChunks()) { throw OsGetChunksMismatch.instance(); }
if (!digester.matches(oi.getDigest())) { throw OsGetDigestMismatch.instance(); }
```

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
scottf: Fixing tests for state setup.

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
scottf: Also, why is there a url in the get-link test?

scottf: Changed code to download the file from the url in the json

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
scottf: Except if you kill the test in the middle, the bucket is sometimes left around. This caused much pain, so once I figured it out, I left it in there.

src/test/resources/data/nats-server.zip
scottf: files removed.

src/test/java/io/nats/compatibility/Log.java
scottf: Removed.

src/test/java/io/nats/compatibility/ObjectStoreCommand.java
piotrpio: I think the url in get link is there by accident, it has no value for the client.

piotrpio (approved):
Looks good, just one suggestion I forgot about earlier.

It would be nice to be able to customize the nats URL when running the test - it is possible in compat tester using `--nats-url` flag, so would be good to enable running compatibility tests on custom URL. Go and Rust tests utilize `NATS_URL` env variable for this, which may easily be passed to docker container, which is where those tests are executed.

scottf:
> Looks good, just one suggestion I forgot about earlier.
> 
> It would be nice to be able to customize the nats URL when running the test - it is possible in 

I Added command line and env support for server

PR #1014, opened 2023-10-16
Accept Tls First as an Option

scottf:


piotrpio:
Will the implementation of TLS first come as a follow up to this PR or will you update this PR?

scottf:
Branch name was supposed to be `tls-first` not `ttl-first`

PR #1015, opened 2023-10-17
Accept Tls First as an Option

scottf:


src/main/java/io/nats/client/Options.java
```diff
@@ -1465,6 +1475,15 @@ public Builder ignoreDiscoveredServers() {
             return this;
         }
 
+        /**
+         * Set tls first behavior on. Default is off
+         * @return the Builder for chaining
```
piotrpio, line 1480: It would be good to expand this comment to mention that the respective settings have to be set in the server itself for this to work. E.g. go client has:
```go
	// TLSHandshakeFirst is used to instruct the library perform
	// the TLS handshake right after the connect and before receiving
	// the INFO protocol from the server. If this option is enabled
	// but the server is not configured to perform the TLS handshake
	// first, the connection will fail.
```

piotrpio (commented):
Looks good, just one nit

src/main/java/io/nats/client/Options.java
scottf: updated the api and added notes to the readme.

piotrpio (approved):
LGTM!

PR #1022, opened 2023-10-23
Initialize SocketDataPort directly

scottf:
Initialize SocketDataPort directly when dataPortType is not overridden. Better for GraalVM


Jarema (approved):
LGTM!

PR #1027, opened 2023-10-26
Update jnats-server-runner to 1.2.6

scottf:
jnats-server-runner 1.2.6 has better output when server fails to run during unit testing.


aricart (approved):
LGTM

PR #1031, opened 2023-10-30
TimeoutException instead of CancellationException on timeout

maximbreitman:
It would be nice to take the first step in this direction - #797 
Since this change is still "breaking", we can make it optional until next version

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -2140,7 +2140,7 @@ public class Options {
      * Get whether to throw {@link java.util.concurrent.TimeoutException} on timeout instead of {@link java.util.concurrent.CancellationException}.
      * @return the flag
      */
-    public boolean isUseTimeoutException() {
+    public boolean useTimeoutException() {
         return useTimeoutException;
     }
 
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -444,6 +444,11 @@ public class Options {
      */
     @Deprecated
     public static final String PROP_UTF8_SUBJECTS = "allow.utf8.subjects";
+    /**
+     * Property used to throw {@link java.util.concurrent.TimeoutException} on timeout instead of {@link java.util.concurrent.CancellationException}.
+     * {@link Builder#useTimeoutException()}.
+     */
+    public static final String PROP_USE_TIMEOUT_EXCEPTION = PFX + "use.timeout.exception";
 
     // ----------------------------------------------------------------------------------------------------
     // PROTOCOL CONNECT OPTION CONSTANTS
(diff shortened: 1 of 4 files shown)


src/main/java/io/nats/client/support/NatsRequestCompletableFuture.java
```diff
@@ -6,11 +6,14 @@
 import java.time.Duration;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeoutException;
 
 /**
  * This is an internal class and is only public for access.
  */
 public class NatsRequestCompletableFuture extends CompletableFuture<Message> {
+    public static final boolean USE_TIMEOUT_EXCEPTION = Boolean.getBoolean(System.getProperty("nats.use-timeout-exception")); // TODO: Remove. Use TimeoutException instead of CancellationException.
```
scottf, line 15: Can we add `nats.use-timeout-exception` in the Option class
```
public static final String PROP_USE_TIMEOUT_EXCEPTION = PFX + "use.timeout.exception";
```
This will also mean that you will have to modify the `NatsRequestCompletableFuture` constructor to receive the setting from the connection and then change `NatsConnection` line 1170

scottf (changes requested):

src/main/java/io/nats/client/support/NatsRequestCompletableFuture.java
maximbreitman: done. however, it would be nice if you labeled these changes as temporary to support legacy. in the next version, it's preferable to use TimeoutException without additional questions, IMO.

src/main/java/io/nats/client/Options.java
```diff
@@ -2117,6 +2136,14 @@ public boolean isTlsFirst() {
         return tlsFirst;
     }
 
+    /**
+     * Get whether to throw {@link java.util.concurrent.TimeoutException} on timeout instead of {@link java.util.concurrent.CancellationException}.
+     * @return the flag
+     */
+    public boolean isUseTimeoutException() {
```
scottf, line 2143: To be more consistent with other booleans in this class, please use `useTimeoutException` no `is`

src/main/java/io/nats/client/support/NatsRequestCompletableFuture.java
```diff
@@ -54,4 +54,8 @@ public boolean wasCancelledClosing() {
     public boolean wasCancelledTimedOut() {
         return wasCancelledTimedOut;
     }
+
+    public boolean isUseTimeoutException() {
+        return useTimeoutException;
+    }
```
scottf, line 60: you don't need this method.

scottf (changes requested):
Looks good except for nitpicks.

src/main/java/io/nats/client/support/NatsRequestCompletableFuture.java
scottf: I'm okay with adding a comment to make this the default in future versions.

scottf:
@maximbreitman Please update your branch

scottf (approved):
LGTM

PR #1052, opened 2023-12-07
Make certain classes Serializable as required by Apache Flink

somratdutta:
I propose an update to the NATS Java Client: making StreamConfiguration and Consumer Configuration classes Serializable.  This change is vital for integrating NATS with Apache Flink (Nats-Flink Connector with JetStream support) , as Flink requires all fields in its operators and functions to be Serializable for distributed processing.


scottf (changes requested):
1. Please add serialVersionUID 
    ```
    private static final long serialVersionUID = 1L;
    ```
2. Nitpick, please add a space
    ```
    JsonSerializable, Serializable
    ```

scottf:
besides the review comments, I would like to understand why would you need a stream configuration? The connector should not be adding or modifying streams, that's not it's job. And all streams have names, which are serializable.
Similar to that why are you needing consumer configuration? For the connector, there should be very few settings that are ever used, I think it would be better to have your own serializable structure and simply use the builder each time, or even better just bind by stream name and consumer name.

sourabhaggrawal:
Hello @scottf thank you for taking time to review the changes, I agree the flink connector should be there only to publish/consume messages and not create the Stream/Consumer at all. I also think of a usecase where one might have 100s of flink job to run and may have to create those resource in advance, which is why we want to create these resources via flink connector. Please let me know your thoughts on this, happy to accept your suggestions as well :)

souravagrawal:
Hello @scottf thank you for taking time to review the changes, I agree the flink connector should be there only to publish/consume messages and not create the Stream/Consumer at all. I also think of a usecase where one might have 100s of flink job to run and may have to create those resource in advance, which is why we want to create these resources via flink connector. Please let me know your thoughts on this, happy to accept your suggestions as well :)

somratdutta:
There is no dependency on this PR currently with our work i.e. JetStream Support on Syandia's NATS-Flink connector.
https://github.com/synadia-io/flink-connector-nats/pull/4
I am fine to close this PR.
@scottf

scottf:
Closed. We cannot change serialization without a major release. To be handled by client.

PR #1054, opened 2023-12-13
Message Headers Improvement

scottf:
When a message is made, copy the headers and make them read only. Currently headers could be modified since they are just a reference to what the dev put in the message, but the length calculations are already done and changing headers will have a bad side effect.

### DIFF
--- a/src/main/java/io/nats/client/impl/Headers.java
+++ b/src/main/java/io/nats/client/impl/Headers.java
@@ -41,22 +41,26 @@ public class Headers {
 	private int dataLength;
 
 	public Headers() {
-		this(null, false, (String[])null);
+		this(null, false, null);
 	}
 
 	public Headers(Headers headers) {
-		this(headers, false, (String[])null);
+		this(headers, false, null);
 	}
 
-	public Headers(Headers headers, boolean readOnly, String... keysToRemoveWhenCopying) {
+	public Headers(Headers headers, boolean readOnly) {
+		this(headers, readOnly, null);
+	}
+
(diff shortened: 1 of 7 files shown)


src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -36,21 +36,47 @@ public class Headers {
...
+						if (tempValuesMap.remove(key) != null) {
+							dataLength -= tempLengthMap.remove(key);
+						}
+					}
+				}
+			}
+		}
+		this.readOnly = readOnly;
+		if (readOnly) {
+			valuesMap = Collections.unmodifiableMap(tempValuesMap);
```
sdklib, line 74: unmodifiableMap throws an `UnsupportedOperationException` when `computeIfAbsent` is called which is happening in the `_add` method.

scottf: @sdklib Are you trying to modify the headers after the message is created? This PR was intended to prevent this.

sdklib: Yup, we're using AOP to add a correlation id for proper logging across microservices to the headers.

After updating to the newest version, some unit tests fail, due to readOnly being true and the AOP Pointcut failing to add the correlation ID to the header.

The fix was easy, I was just wondering if the response should really be `UnsupportedOperationException` or maybe better returning a `Headers.class` clone that doesn't have the add method in the first place. :-)

sdklib: Feared so, never the less, thank you so much! 👍

PR #1074, opened 2024-02-07
Add SLF4J Simple Logging Facade 

stefanLeo:
This a pull request to add SLF4J logging facade to NATS Java Library. 
This has a number of benefits: 

1. Consumers of the library can plug in their logging framework of choice and have all the NATS Java logs where they expect them to be
2. A lot of boilerplate code was removed by using SLF4J
3. I added a few logs, but not too many especially around areas where I would have expected some and where we had troubles with resulting in us extending NATS Java to get logs from that code points. E.g., DNS resolving.
4. I added simplelogger for testing so that there are no differences in daily usage for testing

Happy to receive your thought and feedback on this. 
Happy also to change things upon your input/request. 

@scottf THX in advance for taking the time to at least have a quick look at this PR. 
Let me know if there is anything I have to accept in terms of contribution agreements.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,10 +36,12 @@ repositories {
 }
 
 dependencies {
+    implementation 'org.slf4j:slf4j-api:2.0.12'
     implementation 'net.i2p.crypto:eddsa:0.3.0'
     testImplementation 'org.junit.jupiter:junit-jupiter:5.9.0'
     testImplementation 'io.nats:jnats-server-runner:1.2.8'
     testImplementation 'nl.jqno.equalsverifier:equalsverifier:3.12.3'
+    testImplementation 'org.slf4j:slf4j-simple:2.0.12'
 }
 
 sourceSets {
--- a/dependencies.md
+++ b/dependencies.md
@@ -11,27 +11,28 @@ This file lists the dependencies used in this repository.
 
 #### Runtime Dependencies
 
-| Dependency                           | License                                 |
-|--------------------------------------|-----------------------------------------|
-| net.i2p.crypto:eddsa:0.3.0           | CC 1.0 Universal                        |
+| Dependency                 | License          |
+|----------------------------|------------------|
+| net.i2p.crypto:eddsa:0.3.0 | CC 1.0 Universal |
+| org.slf4j:slf4j-api:2.0.12 | MIT License      |
 
 #### Test Dependencies
 
-| Dependency                                      | License                                 |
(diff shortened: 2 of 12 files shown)


scottf:
@stefanLeo First, I/we appreciate contributions. Regarding contributions, commits must be verified and unit tests should be provided where appropriate.

As far as integrating this PR, I'm not going to, for the following reasons.
* We avoid any 3rd party dependencies if at all possible, For instance, I had to write a json parser by hand. This is mostly a liability and maintenance issue (avoiding them all,) imagine if SLF4J has a security issue like the Log4J problem from a year or so back.
* No dependencies makes the the client easier to integrate. 
* Adding a dependency like this could be considered a major version change as someone might have to change/edit/re-code their logging system.
* Some of the places you added logging already call handlers to listeners, logging would be redundant.
* We cannot remove public api like `public boolean isTraceConnection()`, as this would violate semver without a major version change.
* The default error listener can easily be replaced in connection options, for example with exactly what you did.

Something I would take is a way to plug in the implementation of the tracer. So maybe in the Options Builder there is a setter that takes a `ConnectionTraceHandler` or something like that has

```
void timeTrace(String format, Object... args);
void timeTrace(boolean trace, String message);
```

And a have a default implementation that is used if isTraceConnection is set to true. We could then change all the non-public trace implementations in NatsConnection to use the implementation.

In the options, if the dev sets the ConnectionTraceHandler implementation, this automatically turns trace connection on.

stefanLeo:
Understood - Thank your your feedback and review though. 

Given the strategy outlined - would you mind if I create another request following your pattern of basically letting library consumers hook in their custom listener classes for logging so as to easily exchange the current "sytem.out.println" code as an option in a non-breaking way? Similar than for the errorlisteners. The goal would be to integrate the logs currently produced by the library statically to system.out into the logging framework of choice of library consumers so that they can have all relevant logs in a single place, whatever that may be. I.e., building your own poor mans version of a logging facade without the fancy log levels and param injection. 

Regarding Log4j2 and Security - just fyi: https://openssf.org/blog/2022/07/18/results-of-sigstore-and-slf4j-security-audits/
After all it is just a logging facade and no framework of its own. I think it is really sound, small, super well known in the Java community and secure, but I see your point. 
You're right, it is somehow breaking in the sense that people have to choose their logging lib if they care about the logs once you introduce slf4 into nats-java.

scottf:
Yes, 100% would welcome a PR for the plugin for tracing.

PR #1077, opened 2024-02-11
User Pluggable Connection Time Trace Logger

scottf:
Allow users to provide a replacement for the trace logging feature. Currently it's just simple check a flag / write to the console.
This change provides a fully backward compatible implementation that can be overridden per connection.

#### Also Fixed
Also fixed timeCheck reporting


src/main/java/io/nats/client/Options.java
```diff
@@ -1661,6 +1681,22 @@ else if (useDefaultTls) {
...
+                }
+                else {
+                    timeTraceLogger = (f, a) -> {};
+                }
+            }
+            else {
+                // if the dev provided an impl, we assume they meant to time trace the connection
+                traceConnection = true;
+            }
+
```
scottf, line 1699: the old impl just had the `traceConnection` flag. Since we are switching to an interface/impl, this checks the flag and makes the appropriate impl. If the flag is false, the impl is a no-op, which is actually less operations than the old implementation that would call a method, pass a the traceConnection flag and check the flag even when trace connection is off. The no-op is just a method call as a comparison.

src/main/java/io/nats/client/Options.java
```diff
@@ -1731,13 +1771,8 @@ public Builder(Options o) {
...
     private Options(Builder b) {
-        if (b.natsServerUris.isEmpty()) {
-            this.natsServerUris = Collections.singletonList(DEFAULT_NATS_URI);
-        }
-        else {
-            this.natsServerUris = Collections.unmodifiableList(b.natsServerUris);
-        }
-        this.unprocessedServers = b.unprocessedServers;  // exactly how the user gave them
+        this.natsServerUris = Collections.unmodifiableList(b.natsServerUris);
+        this.unprocessedServers = Collections.unmodifiableList(b.unprocessedServers);  // exactly how the user gave them
```
scottf, line 1775: 1.After reviewing the builder, I noticed that `natsServerUris` was already never empty so that was unnecessary. 
2. All lists, since they can be viewed from public api needs to be unmodifiable.

stefanLeo:
Yes. Fully understood. So fhe use cases are from my PoV:
1) Have an easy integration to grab whatever you log deep inside your code and ingest into my custom logging framework incl. the timestamps etc.
I fully agree that excessive logging ist not a good idea here, but adding logs at some critical points can help a lot with troubleshooting especially for new users who do not yet know what is important for nats to add to as listeners etc...
2) If the lib provides its set of logs it is much easier for you as a maintainer to debug/analyse issues on a customer site or if anyone opens a bug...
After all it is then your logs and not just some provided by some custom injected logging that may be compleltely wrong and/or it takes you ages to understand those logs.
3) So far for logging in general. Specifically: With DNS we had issues with the DNS server during failover tests  in Openshift which led to reconnection timeouts of 30+ secs. It took us one week to figure out that the dns part was problematic using a profiling techniques.  I.e., imho whenever something you request from an external system  you should log it and the returned result in case it is something that does not happen all the time e.g.. connection interruptions or abortions, dns resolving, config parsing issues,... things like that.
Simply to make everyones life easier when having to troubleshoot issues or integrate nats.

scottf:
Ok let's think about places where we can safely add logging into. Anything around Connection/Reconnects. Anything around JetStreamManagement, but that's high level, so not sure how that is useful. I know sometimes the errors are IOException and it's kinda useless. Where else?

stefanLeo:
Examples from my Pov:

- DNS resolving 
- Reconnections
- authentication issues/TLS issues
- Connection abortions due to whatever (timeouts, ping outages, ...)
- Unparsabel messages received like the IllegalStateExceptions / NumberFormatExceptions in NatsConnectionReader > you get them via the handlers I suppose anyhow.. but imho logging the error before throwing an exception still makes sense. 
- Some others that may be valuable: 
   -- If the file used in fileauthhandler is null you assume this is a challengae only auth > log this assumption. 
   -- jwt token parsing issues in fileauthhandler (right now it is an Exception inside an IllegalStateException inside an IOException
   -- not a nats expert so there may be more in that direction to be added peu à peu

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -355,37 +358,42 @@ else if (first.equals(cur)) {
...
+        else if (remaining < 1_000_000) {
+            timeTraceLogger.trace(message + String.format(", %d (ns) remaining", remaining));
+        }
+        else if (remaining < 1_000_000_000) {
+            long ms = remaining / 1_000_000;
+            timeTraceLogger.trace(message + String.format(", %d (ms) remaining", ms));
+        }
+        else {
+            double seconds = ((double) remaining) / 1_000_000_000.0;
+            timeTraceLogger.trace(message + String.format(", %.3f (s) remaining", seconds));
```
scottf, line 395: This block is the change to reporting the time check

src/test/java/io/nats/client/ConnectTests.java
```diff
@@ -331,10 +333,40 @@ public void testSlowConnectionNoTimeout() throws Exception {
...
+        assertTrue(traces.get(i++).startsWith("connecting data port"));
+        assertTrue(traces.get(i++).startsWith("reading info"));
+        assertTrue(traces.get(i++).startsWith("starting reader"));
+        assertTrue(traces.get(i++).startsWith("starting writer"));
+        assertTrue(traces.get(i++).startsWith("sending connect message"));
+        assertTrue(traces.get(i++).startsWith("sending initial ping"));
+        assertTrue(traces.get(i++).startsWith("starting ping and cleanup timers"));
+        assertTrue(traces.get(i++).startsWith("updating status to connected"));
+        assertTrue(traces.get(i++).startsWith("status updated"));
+        assertTrue(traces.get(i).startsWith("connect complete"));
```
scottf, line 369: This is assuring backward compatibility on messages in case someone was relying on them in logs.

src/test/java/io/nats/client/impl/JetStreamGeneralTests.java
```diff
@@ -982,6 +982,45 @@ public void testGetJetStreamValidatedConnectionCoverage() {
...
+        nc.traceTimeCheck("lt 0, gt -1_000_000", -1);
+        assertEquals("lt 0, gt -1_000_000, 1 (ns) beyond timeout", l.lastTrace);
+
+        nc.traceTimeCheck("lt 0, gt -1_000_000_000", -1_000_000);
+        assertEquals("lt 0, gt -1_000_000_000, 1 (ms) beyond timeout", l.lastTrace);
+
+        nc.traceTimeCheck("lt 0, lt -1_000_000_000", -1_100_000_000);
+        assertEquals("lt 0, lt -1_000_000_000, 1.100 (s) beyond timeout", l.lastTrace);
+    }
+
```
scottf, line 1023: This tests the new implementation of time check reporting.

jarretlavallee (approved):
LGTM

PR #1078, opened 2024-02-13
chore: upgrade documentation to specify backoff behaviour

imclem:
Hi there 👋

Just stumbled upon an issue setting both ackwait and backoff in the consumer configuration. 

As per the [documentation ](https://docs.nats.io/using-nats/developer/develop_jetstream/consumers#delivery-reliability):
> You can control the timing of re-deliveries using either the single AckWait duration attribute of the consumer, or as a sequence of durations in the BackOff attribute (which overrides AckWait).

Backoff will override AckWait, I thought it would be great to specify this in the javadoc.
I think it's also worth logging an error at runtime when both are set but I'm not sure where to plug this ?

Let me know if you want me to do any change

### DIFF
--- a/src/main/java/io/nats/client/api/ConsumerConfiguration.java
+++ b/src/main/java/io/nats/client/api/ConsumerConfiguration.java
@@ -1166,6 +1166,7 @@ public class ConsumerConfiguration implements JsonSerializable {
 
         /**
          * Set the list of backoff. Will override ackwait setting.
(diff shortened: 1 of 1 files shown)


scottf (changes requested):
1. Can you add a link to the doc? https://docs.nats.io/using-nats/developer/develop_jetstream/consumers#delivery-reliability

2. Commits need to be signed. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

scottf (approved):

scottf (approved):

imclem:
@scottf I've added the link, is this what you wanted ? I've also signed the last commit

PR #1081, opened 2024-02-18
KV Watch From Revision

scottf:


src/main/java/io/nats/client/impl/NatsWatchSubscription.java
```diff
@@ -30,25 +32,29 @@ public NatsWatchSubscription(JetStream js) {
...
-                ConsumerConfiguration.builder()
-                    .ackPolicy(AckPolicy.None)
-                    .deliverPolicy(deliverPolicy)
-                    .headersOnly(headersOnly)
-                    .filterSubject(subscribeSubject)
-                    .build())
+            .configuration(ConsumerConfiguration.builder()
+                .ackPolicy(AckPolicy.None)
+                .deliverPolicy(deliverPolicy)
+                .startSequence(fromRevision)
```
piotrpio, line 54: Is it ok to always be setting this option (even to 0 if not used)?

piotrpio (approved):
LGTM, just one question

src/main/java/io/nats/client/impl/NatsWatchSubscription.java
scottf: Yes. I check the value before setting it and the builder knows how to deal with it.

PR #1083, opened 2024-02-22
Optional dispatching with executor instead of blocking

scottf:
Added optional ability to make the dispatcher deliver messages using the task executor instead of with a blocking call.


aricart (approved):
Lgtm

PR #1095, opened 2024-03-04
ReentrantLock instead of synchronized

scottf:
* Use ReentrantLock in place of synchronized since it's better for virtual threads and equally if not more performant.
* Ensure consumer context implementation is thread safe and switch to reentrant lock instead of synchronized.


MauriceVanVeen (approved):
LGTM

PR #1098, opened 2024-03-13
KV Transform support by ensuring using stream name in KV under the co…

scottf:
…ver subscriptions.

### DIFF
--- a/src/main/java/io/nats/client/api/SourceBase.java
+++ b/src/main/java/io/nats/client/api/SourceBase.java
@@ -48,8 +48,7 @@ public abstract class SourceBase implements JsonSerializable {
         subjectTransforms = SubjectTransform.optionalListOf(readValue(jv, SUBJECT_TRANSFORMS));
     }
 
-    @SuppressWarnings("rawtypes") // Don't need the type of the builder to get its vars
-    SourceBase(SourceBaseBuilder b) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/nats/client/impl/NatsFeatureBase.java
```diff
@@ -71,6 +71,7 @@ protected MessageInfo _getBySeq(long seq) throws IOException, JetStreamApiExcept
 
     protected void visitSubject(String subject, DeliverPolicy deliverPolicy, boolean headersOnly, boolean ordered, MessageHandler handler) throws IOException, JetStreamApiException, InterruptedException {
         PushSubscribeOptions pso = PushSubscribeOptions.builder()
+            .stream(streamName)
```
scottf, line 74: It's always better to supply the stream when you know it, this should have been there originally.

src/main/java/io/nats/client/api/SourceBase.java
```diff
@@ -132,6 +131,7 @@ public SourceBaseBuilder(SourceBase base) {
             this.startTime = base.startTime;
             this.filterSubject = base.filterSubject;
             this.external = base.external;
+            this.subjectTransforms = base.getSubjectTransforms();
```
scottf, line 134: This was originally missed.

PR #1101, opened 2024-03-22
SSLContextFactory

scottf:
Adding an SSLContextFactory allows Options to take a class name property for the factory to which makes an SSLContext, instead of having to provide a concrete implementation. This allows classes using the library more flexibility.


jnmoyne (approved):
LGTM

PR #1104, opened 2024-03-29
Resolve complex conditional

photizzo:
- Introduced a new method `isDispatcherCreationNeeded` to simplify the conditional logic in the Service constructor.
- This refactoring encapsulates the check for any null dispatchers into a single, clear, and reusable method. 
- Now, if additional dispatchers are introduced or the logic for determining a missing dispatcher changes, we only need to update one place, reducing the risk of errors and improving code manageability.

### DIFF
--- a/src/main/java/io/nats/service/Service.java
+++ b/src/main/java/io/nats/service/Service.java
@@ -87,11 +87,10 @@ public class Service {
         pingResponse = new PingResponse(id, b.name, b.version, b.metadata);
         infoResponse = new InfoResponse(id, b.name, b.version, b.metadata, b.description, b.serviceEndpoints.values());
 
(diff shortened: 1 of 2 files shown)


scottf:
This code is not used anywhere else so 
- this adds an unnecessary function call
- makes the code less readable because it's important to know the contents here, not just a function name

photizzo:
> This code is not used anywhere else so
> 
> * this adds an unnecessary function call
> * makes the code less readable because it's important to know the contents here, not just a function name

Thank you @scottf for pointing that out. 

I has also considered the scenerio incase additional dispatchers are added. It would make the update more manageable. Doing some deep dive like you suggested.

Issue #941, opened 2023-07-14
Document that, when using an ordered push consumer, devs should use JetStreamSubscription#getConsumerInfo().getCalculatedPending(), not JetStreamSubscription#getPendingMessageCount()


matthewadams:
For the use case where a stream consumer wants to retrieve all events using a subject filter, it should be documented that the developer should _not_ use `JetStreamSubscription#getPendingMessageCount()`, and _should_ use `JetStreamSubscription#getConsumerInfo().getCalculatedPending()`.  This was tribally known prior to my filing this issue, but needs to be called out explicitly in the documentation.  Further, I think that the javadoc of these properties should be beefed up so that those that are looking at the source understand the differences.

Link to Slack discussion where this is called out: https://natsio.slack.com/archives/C055AB5M1KM/p1689197213989439?thread_ts=1689111301.307819&cid=C055AB5M1KM

Issue #547, opened 2021-09-20
connect trace does not log server uri when reconnecting


hoodaajay1:
When we connect to nats initially the trace shows the server UI

[07:01:00.298] connect trace: creating connection object
[07:01:00.324] connect trace: creating executors
[07:01:00.325] connect trace: creating reader and writer
[07:01:00.33] connect trace: connection object created
[07:01:00.33] connect trace: starting connect loop
[07:01:00.331] connect trace: setting status to connecting
**[07:01:00.331] connect trace: trying to connect to nats://localhost:4222**
[07:01:00.331] connect trace: starting connection attempt, 30.000 (s) remaining
[07:01:00.332] connect trace: waiting for reader, 29.999 (s) remaining

However when it tries to reconnect the URI is not logged which is not helpful as if we have different nats servers to connect to & one of them has a problem its tough to find out.

Changing line 357 in NatsCOnnection.java to will fix it.
timeCheck(trace, end, "starting connection attempt to " + serverURI);

[07:02:26.735] connect trace: starting connection attempt, 30.000 (s) remaining
[07:02:26.736] connect trace: waiting for reader, 29.999 (s) remaining
[07:02:26.736] connect trace: waiting for writer, 29.999 (s) remaining
[07:02:26.736] connect trace: cleaning pong queue, 29.999 (s) remaining
[07:02:26.737] connect trace: connecting data port, 29.998 (s) remaining
[07:02:26.738] connect trace: reading info, version and upgrading to secure if necessary, 29.997 (s) remaining
[07:02:26.739] connect trace: starting reader, 29.996 (s) remaining
[07:02:26.74] connect trace: starting writer, 29.995 (s) remaining
[07:02:26.741] connect trace: sending connect message, 29.995 (s) remaining
[07:02:26.741] connect trace: sending initial ping, 29.994 (s) remaining
[07:02:26.742] connect trace: updating status to connected, 29.993 (s) remaining
[07:02:26.742] connect trace: status updated

scottf:
That would probably be useful. I'll look into it when I get a chance, you are always welcome to make a fork and submit a PR.

hoodaajay1:
Sure I would be happy to do that. 

By the way I tried to check in to a new branch but got 403. 

Thanks

scottf:
https://github.com/nats-io/nats.java/pull/1106

PR #1105, opened 2024-03-29
Introduced `shouldRecordStats` explaining variable in the `onMessage` method to clarify the intent behind conditional checks for recording statistics.

photizzo:
While previous [feedback](https://github.com/nats-io/nats.java/pull/1104#issuecomment-2027225690) has raised valid concerns about adding an additional function call and its impact on code readability, this PR aims to address a specific need for clarity in how we handle statistics recording in the onMessage method and error handling

Earlier comment for [reference](https://github.com/nats-io/nats.java/pull/1104#issuecomment-2027225690)

### DIFF
--- a/src/main/java/io/nats/service/EndpointContext.java
+++ b/src/main/java/io/nats/service/EndpointContext.java
@@ -17,7 +17,7 @@ class EndpointContext {
     private final Connection conn;
     private final ServiceEndpoint se;
     private final ServiceMessageHandler handler;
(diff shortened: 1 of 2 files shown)


scottf:
Refactoring one use code is not necessary. If you sincerely want to make contributions please try to add or fix functionality. Here are some issue you could try to address:
https://github.com/nats-io/nats.java/issues/547
https://github.com/nats-io/nats.java/issues/941

PR #1109, opened 2024-04-03
Ensure NatsConnection inboxDispatcher is started prior to publishing messages

### LINKED ISSUE #1065: Unexpected timeout when requesting subject without handlers
### Observed behavior

When two requests (subject without handlers) started simultaneously on different threads but the same connection - sometimes only one is cancelled immediately by 503 (CancellationException). Second is cancelled by timeout.

### Expected behavior

Both requests fail immediately by CancellationException.

### Server and client version

Server: reproduced on 2.8.4 and 2.10.9
Client: reproduced on 2.15.3 and 2.17.2

### Host environment

aarch64 with nats cluster running in docker

### Steps to reproduce

```java
public class CancellationErrorExample {
    private static void runFor503(Connection connection) {
        try {
            // Handler for `subject1` does not exist, and we should
            // immediately get 503 (CancellationException with null message).
            connection.requestWithTimeout("subject1", new byte[]{}, Duration.ofSeconds(5)).get();
        } catch (Exception e) {
            System.out.println(Thread.currentThread().getName() + ": " + e.getClass().getSimpleName() + " " + e.getMessage());
        }
    }

        Options options = new Options.Builder().server("nats://localhost:4222").useTimeoutException().build();

        while (true) {
            try (final Connection connection = Nats.connect(options)) {
                System.out.println();

                Thread t1 = new Thread(null, () -> runFor503(connection), "Thread 1");
                Thread t2 = new Thread(null, () -> runFor503(connection), "Thread 2");

                t1.start();
                t2.start();

                t1.join();
                t2.join();
            } catch (Exception e) {
                throw new RuntimeException(e);
            }
        }
    }
}
```

Typical output:

```
Thread 1: CancellationException null
Thread 2: CancellationException null

Thread 1: CancellationException null
Thread 2: CancellationException null

Thread 1: CancellationException null
Thread 2: ExecutionException java.util.concurrent.TimeoutException: Future cancelled, response not registered in time, check connection status.
```

Seems like this problem is related to `inboxDispatcher` initialization in `NatsConnection.java`. _INBOX subscription is not started properly (by the first thread) when second request is already sent.

There is a comment about race condition. It says that AtomicReference `is probably enough`, but seems like it's not for this case.

nathanschile:
If multiple treads publish messages at the same time, some messages may not be published because the inboxDispatcher is not started yet.  Change updates NatsConnection to ensure the inboxDispatcher is started prior to publishing messages.

 Fixes #1065

### DIFF
--- a/README.md
+++ b/README.md
@@ -1221,7 +1221,7 @@ The java doc is located in `build/docs` and the example jar is in `build/libs`. 
 
 which will create a folder called `build/reports/jacoco` containing the file `index.html` you can open and use to browse the coverage. Keep in mind we have focused on library test coverage, not coverage for the examples.
 
-Many of the tests run nats-server on a custom port. If nats-server is in your path they should just work, but in cases where it is not, or an IDE running tests has issues with the path you can specify the nats-server location with the environment variable `nats_-_server_path`.
+Many of the tests run nats-server on a custom port. If nats-server is in your path they should just work, but in cases where it is not, or an IDE running tests has issues with the path you can specify the nats-server location with the environment variable `nats_server_path`.
 
 ## TLS Certs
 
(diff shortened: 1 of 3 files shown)


scottf (approved):
LGTM

Issue #1044, opened 2023-11-22
Unable to run on Android 7.0 because java.util.Base64


zpsw:
### Observed behavior

java.util.concurrent.ExecutionException: java.lang.NoClassDefFoundError: Failed resolution of: Ljava/util/Base64;

### Expected behavior

can run on Android 7.0

### Server and client version

jnats 2.17.1

### Steps to reproduce

run on Android 7.0

scottf:
@zpsw How does a 4 year old Android version not have part of Java 8? I don't think this is fixable, but I'm welcome any suggestions you might have to fix this.

zpsw:
Android 7, named Nougat, is a version from 7 years ago. This version indeed lacks java.util.Base64. My current approach is to fork a copy of the code, extract java.util.Base64, and then replace all references. I'm not sure if this is a good solution.
https://stackoverflow.com/questions/55016350/android-nougat-devices-not-supporting-java-util-base64-library
<img width="468" alt="image" src="https://github.com/nats-io/nats.java/assets/50564542/26ddf6e2-877b-403d-9a7a-e6d07fe7145f">
<img width="264" alt="image" src="https://github.com/nats-io/nats.java/assets/50564542/8b13b23c-3999-40fd-b8f9-cdadef127dde">

scottf:
@zpsw Do you have an implementation that we could possibly put in the code? If you wrote the code that would be the best, if it's a 3rd party library, even OSS, I'd have to consider it carefully.

scottf:
@zpsw We are currently in discussion to hard fork the repo and port it to api 24. base64 isn't just the only thing in the code that needs to change. I'll try to remember to keep you up to date.

pwittchen:
I've added possible solution for this issue in the PR #1120.

scottf:
Closing due to not going to be addressed.

PR #1120, opened 2024-04-11
keeping compatibility with Android apps using Android API older than API 26

pwittchen:
…, Oreo (Android 8.0) by adding internal implementation of the `Base64` class named as `Base64Utils` to distinguish it from the native Java API, resolves issue #1044

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -14,6 +14,7 @@
 package io.nats.client;
 
 import io.nats.client.impl.*;
+import io.nats.client.support.Base64Utils;
 import io.nats.client.support.HttpRequest;
 import io.nats.client.support.NatsConstants;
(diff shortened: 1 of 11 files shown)


scottf:
A couple things in general for work on this open source repo.
1. We cannot use the Base64Utils because of it's copyright and lack of an open source friendly license.
2. We cannot change the signature of any public method since this would be a breaking change and require a major version update.

Regarding a client that is compatible with Android 24 or 26...
We are currently in the processes of building a conversion tool that will pull the latest version of the client from git and replace all instances of non-compatible code. This includes having replacements for several things including Base64, Duration, CompletableFuture, etc.
This is currently being done for a paying customer and it has not been determined how this will be publicly available. At a minimum, we need to find replacements that are licensed in a way that it is acceptable to use in the form we release the project.

PR #1127, opened 2024-04-22
tls_available support

scottf:
Add support for server setting of `tls_available`


src/main/java/io/nats/client/Options.java
```diff
@@ -1710,6 +1710,10 @@ else if (useDefaultTls) {
                 }
             }
 
+            if (tlsFirst && sslContext == null) {
+                throw new IllegalStateException("SSL context required for tls handshake first");
+            }
+
```
scottf, line 1716: I did not used to check this before, but if there is no SSL Context, we can't do any tls anyway, so it's a fast fail essentially.

src/main/java/io/nats/client/Options.java
```diff
@@ -2101,10 +2105,10 @@ public int getMaxControlLine() {
 
     /**
      *
-     * @return true if there is an sslContext for this Options, otherwise false, see {@link Builder#secure() secure()} in the builder doc
+     * @return true if there is an sslContext for these Options, otherwise false, see {@link Builder#secure() secure()} in the builder doc
      */
     public boolean isTLSRequired() {
-        return tlsFirst || this.sslContext != null;
+        return sslContext != null;
```
scottf, line 2111: reflects checking tls first and ssl context in the builder

src/main/java/io/nats/client/api/ServerInfo.java
```diff
@@ -67,7 +68,8 @@ public ServerInfo(String json) {
         headersSupported = readBoolean(jv, HEADERS);
         authRequired = readBoolean(jv, AUTH_REQUIRED);
         nonce = readBytes(jv, NONCE);
-        tlsRequired = readBoolean(jv, TLS);
+        tlsRequired = readBoolean(jv, TLS_REQUIRED);
+        tlsAvailable = readBoolean(jv, TLS_AVAILABLE);
```
scottf, line 72: 71: Better variable name
72: read the value from the json

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -584,36 +584,25 @@ void checkVersionRequirements() throws IOException {
...
-            else if (!isTLSRequired && serverInfo.isTLSRequired()) {
+            else if (serverInfo.isTLSRequired()) {
                 throw new IOException("SSL required by server.");
             }
-            if (upgradeRequired) {
-                this.dataPort.upgradeToSecure();
-            }
         }
     }
-
```
scottf, line 616: This whole block is simplified.
1. getOptions returns the options variable, just use the variable directly. There is no lazy loading or anything.
2. No need to check 2.9.19 because websocket never does upgrade, it was just extraneous

wallyqs (approved):
LGTM

PR #1129, opened 2024-04-30
Add quickfix to be able to use auth handler and user info

ryad-eldajani:
When trying to connect to a NATS server with [Auth Callout](https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_callout), the client may be required to use a combination of an auth handler (NKey + Seed) plus user info (username + password). This is working as expected in the go reference implementation.

However, this is not working in the current nats.java implementation, as user info is stripped out when building connection options:

```java
AuthHandler authHandler = Nats.staticCredentials(jwt, nkey);
Options connectOptions = Options.builder()
        .server(config.getUrl())
        .authHandler(authHandler)
        .userInfo(username, password)  
        .build();
```
This PR fixes the behavior in the client options to allow the usage of an auth handler in combination with username and password.

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -2371,7 +2371,9 @@ public class Options {
             appendOption(connectString, Options.OPTION_NKEY, nkey, true, true);
             appendOption(connectString, Options.OPTION_SIG, encodedSig, true, true);
             appendOption(connectString, Options.OPTION_JWT, jwt, true, true);
-        } else if (includeAuth) {
+        }
+
+        if (includeAuth) {
             String uriUser = null;
             String uriPass = null;
             String uriToken = null;
--- a/src/test/java/io/nats/client/AuthHandlerForTesting.java
+++ b/src/test/java/io/nats/client/AuthHandlerForTesting.java
@@ -18,13 +18,21 @@ import java.security.GeneralSecurityException;
 
 public class AuthHandlerForTesting implements AuthHandler {
     private final NKey nkey;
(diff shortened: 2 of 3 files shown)


scottf (changes requested):
I've checked with some other devs. Apparently both should be sent and the server decides
```
if (includeAuth) {
    if (nonce != null && this.getAuthHandler() != null) {
        ...
    }
    // check for user/pass ...
}
```

Also, if possible can you verify your commit? I won't require it for this your first, but for subsequent contributions please verify your commits. [Managing commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification)

scottf:
@ryad-eldajani I messed up your pr, I had made a change to options and it got merged here and overwrote your change and reverted back to before your change. Sorry, can you please update.

scottf:
@ryad-eldajani So my commit contains my original modifying of the includeAuth check, which came before I got the feedback from the developers that it's not one or the other, but you can send both. Please make that change, again sorry for the screwup.

ryad-eldajani:
@scottf Thanks for your valuable feedback. I'll try to redo this change within the next few days, as I'm currently out of office for vacation.

Also, sorry for the missing verification. I have committed this change using my working email address, which is currently not assigned to my personal GPG key. Will keep that in mind on the next PR.

Best regards!

ryad-eldajani:
Hi @scottf thanks for your patience. I have created a new PR based on this discussion. Please see https://github.com/nats-io/nats.java/pull/1138

PR #1138, opened 2024-05-13
Add fix to be able to use NKey + Seed auth handler and username + pas…

ryad-eldajani:
Updated PR based on discussion in https://github.com/nats-io/nats.java/pull/1129

When trying to connect to a NATS server with [Auth Callout](https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_callout), the client may be required to use a combination of an auth handler (NKey + Seed) plus user info (username + password). Currently, only either auth handler or user info is append in the options when building the client options string.

AuthHandler authHandler = Nats.staticCredentials(jwt, nkey);
Options connectOptions = Options.builder()
        .server(serverUri)
        .authHandler(authHandler)
        .userInfo(username, password)  
        .build();

This PR fixes the behavior in the client options to allow the usage of an auth handler in combination with username and password.

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -2374,51 +2374,50 @@ public class Options {
                 appendOption(connectString, Options.OPTION_SIG, encodedSig, true, true);
                 appendOption(connectString, Options.OPTION_JWT, jwt, true, true);
             }
-            else {
-                String uriUser = null;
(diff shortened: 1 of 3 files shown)


scottf (approved):
LGTM

ryad-eldajani:
Dear @scottf thanks for your review. Is there something missing or anything I can do to have this merged? Thanks in advance!

PR #1155, opened 2024-05-31
Improve socket close behavior

scottf:
### Force Reconnect improvement
Execute `forceClose` with an executor task inside `forceReconnect`. This prevents the force reconnect behavior from being blocked if close blocks.

---

### SocketDataPortWithWriteTimeout tuning
In the SocketDataPortWithWriteTimeout implementation of the timeout watcher task:
1. Call the `forceReconnect` directly instead of running it on an executor task thread. It's already running on an executor task thread because that's where the watcher runs.
2. Removed the call to close the output stream of the socket. It's going to get closed anyway when the socket gets closed.

---

### Socket SO LINGER

Added an `Options.socketSoLinger(int socketSoLinger)` for the user to set the underlying `socket.setSoLinger()`. This will be used in normal close operations. If a `forceClose` is requested, any option value supplied will be overridden with `setSoLinger(true, 0)`

---

### Force Close

Added a `forceClose()` to the `DataPort` interface (with a default implementation that calls `close()`, to ensure interface backward compatibility) 
1. the base internal implementation, `SocketDataPort` implements `forceClose()`
2. `forceReconnect` calls `forceClose()`
3. `handleCommunicationIssue` calls `forceClose`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,7 +13,7 @@ plugins {
     id 'signing'
 }
 
-def jarVersion = "2.18.2"
+def jarVersion = "2.19.0"
 
 def isRelease = System.getenv("BUILD_EVENT") == "release"
 def brn = System.getenv("BRANCH_REF_NAME")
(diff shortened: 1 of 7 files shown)


aricart (approved):
LGTM

PR #1164, opened 2024-06-21
Making JetStreamReader more robust

roeschter:
Issue, when the reader receivers fewer messages than expected in a pull, it will eventually stall. Most easily reproduced with a server restart/reconnect. 

Not sure whether we consider this a bug.  Maybe make it a feature request.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
+++ b/src/main/java/io/nats/client/impl/NatsJetStreamPullSubscription.java
@@ -342,6 +342,11 @@ public class NatsJetStreamPullSubscription extends NatsJetStreamSubscription {
                 if (currentBatchRed == batchSize) {
                     currentBatchRed = 0;
                 }
+            } else {
+            	//We may need to recover if for whatever reason a pull does not return 
+            	//as many messages as expected.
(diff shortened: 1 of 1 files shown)


scottf:
Can you please provide some example code or a unit test to demonstrate this?

scottf:
Also, can you look at "simplification", i.e. using the ConsumerContext to consumer messages. It's very robust and probably does what you need. There are examples here: https://github.com/nats-io/nats.java/tree/main/src/examples/java/io/nats/examples/jetstream/simple

roeschter:
Seems ConsumerContext  suffers from the same issue:

To reproduce:
1. Run either of the two below - standard JS enabled NATS server on localhost
2. Stop the server and restart immediately
3. Neither JetStreamReader nor ConsumerContext are recovering 
4. Make the suggested change for JetStreamReader - should recover after the timeout set in the request.

Please check my Playground - https://github.com/roeschter/natsPlayground - you should have seen an invite
See the /reproducer package

- JetstreamSubConsumerContext
- JetstreamSubPullReader

roeschter:
Correction: ConsumerContext  does self recovers after timeout - The default is 30s, I did not wait that long.

PR #1168, opened 2024-06-26
CONNECT username and password need json encoding

scottf:


src/test/java/io/nats/client/AuthTests.java
```diff
@@ -88,6 +90,32 @@ private void assertEncoded(String encoded, int port) throws IOException, Interru
         c.close();
     }
 
+    @Test
+    @EnabledOnOs({ WINDOWS })
```
scottf, line 94: I could not get this run on GH Actions. I'm assuming that it has something to do with how the command line handles control characters on unix systems.

src/main/java/io/nats/client/support/Encoding.java
```diff
@@ -270,45 +270,59 @@ public static String jsonEncode(String s) {
...
-                default:
-                    if (ch < ' ') {
-                        sb.append(String.format("\\u%04x", (int) ch));
-                    }
-                    else {
-                        sb.append(ch);
-                    }
-                    break;
-            }
+            appendChar(sb, s.charAt(x));
```
scottf, line 273: refactored for reuse when the entry is string or char array

src/main/java/io/nats/client/Options.java
```diff
@@ -2500,10 +2499,11 @@ private static void appendOption(CharBuffer builder, String key, String value, b
         _appendOptionEnd(builder, quotes);
     }
 
-    private static void appendOption(CharBuffer builder, String key, char[] value, boolean quotes, boolean comma) {
-        _appendStart(builder, key, quotes, comma);
+    @SuppressWarnings("SameParameterValue")
+    private static void appendOption(CharBuffer builder, String key, char[] value, boolean comma) {
```
scottf, line 2503: When used for a char array, it always represents a string value, so must be in quotes

src/main/java/io/nats/client/Options.java
```diff
@@ -2465,24 +2464,24 @@ public CharBuffer buildProtocolConnectOptionsString(String serverURI, boolean in
             }
 
             if (uriUser != null) {
-                appendOption(connectString, Options.OPTION_USER, uriUser, true, true);
+                appendOption(connectString, Options.OPTION_USER, jsonEncode(uriUser), true, true);
```
scottf, line 2467: json encode since this is json, and it's a string that is provided by the user. I said this to say that I didn't encode some strings in this function since I know that they are already json safe, like the jwt which is base64 encoded.

Issue #1182, opened 2024-07-19
NATS option duration based properties do not accept duration string format 


PeterJakubik:
### Observed behavior

Hi
I would like to change the NATS `cleanupinterval` by setting the property in springboot application yaml like
```yaml
io.nats.client.cleanupinterval: PT10S
```

Reason is, that the publisher replays in ~10 secs.

It looks like the `NatsAutoConfiguration` doesn't reflect this cleanupinterval interval set in properties.

### Expected behavior

I expect the property in springboot application.yaml is reflected for NatsConnection options

### Server and client version

nats-spring 0.5.7

### Host environment

springboot

### Steps to reproduce

sample

scottf:
The property currently expects milliseconds so this will work for now:
```
io.nats.client.cleanupinterval: 10000
```

I have created this PR to handle the duration string: https://github.com/nats-io/nats.java/pull/1183

PeterJakubik:
Thanks for quick reply. I suppose I miss something obvious for NATS autoconfiguration. 

I have a demo that has configures NATS `cleanupinterval` to 10000 https://github.com/PeterJakubik/nats-properties/blob/main/src/main/resources/application.yml#L5 

But once the NATS Connection is injected the `cleanupinterval` https://github.com/PeterJakubik/nats-properties/blob/ac0abbe33896be9945573517da5f88caa8ab32de/src/main/java/com/example/demo/DemoApplication.java#L35 is still 5 sec

Why is the NatsAutoConfiguration not picking the `cleanupinterval` from application.yml ?

scottf:
@PeterJakubik Did you see the unit tests I added? The options object is definitely built correctly. So that leaves 2 things. 
1. The options, although set, are not used properly or where you expect them to be used. 
2. Whatever spring mechanism you are using is not working.

PeterJakubik:
I have seen the unit tests and they are correct. I just can't see how java properties like `io.nats.client.cleanupinterval` get loaded into `NatsConnection`.  I tried adding value `io.nats.client.cleanupinterval=12000` to `nats-spring-samples/autoconfigure-sample`   application.properties. I expect this value to be reflected into `NatsConnection` but this doesn't happen.

Please bear with me, how could I configure the `io.nats.client.cleanupinterval` property value in spring in such a way, that the  `NatsConnection` will use it?

scottf:
[...]

PR #1183, opened 2024-07-19
Handle duration string when accepting duration properties

scottf:
Address Issue https://github.com/nats-io/nats.java/issues/1182


MauriceVanVeen (approved):
LGTM

Issue #1184, opened 2024-07-22
Uneven connections rebalance with `forceReconnect`


ajax-koval-d:
### Observed behavior

After nats cluster upgrade, we end up with nats connections distributed unevenly between different nodes. To fix this, we want to use `forceReconnect`.

Problem here is that node resolution in `forceReconnect` is not fair, which becomes quite evident in cluster with two nats server nodes. Nats cluster upgrade leaves all connection hanging on one node, and then after triggering `forceReconnect` on each connection, all connection just reconnect on another node, instead of distributing evenly.

As I understand, reason for this is that currently connected server is guaranteed to be at the end of list in a `NatsServerPool`, meaning, that client will not try to reconnect to the currently connected server.

To fix this I'd propose to shuffle list in `NatsServerPool` inside of `forceReconnect` implementation.

### Expected behavior

After force reconnect all connections are distributed evenly.

### Server and client version

nats-server: v2.10.14
nats.java client: 2.9.1

### Steps to reproduce

1. Run nats cluster with 2 servers
2. Open at least 2 connections to the servers with `ignoreDiscoveredServers=true`
Next, simulating nats server upgrade
4. Turn off server 1
5. Wait at least 5 sec
6. Turn on server 1
7. Wait until server starts
8. Turn off server 2
9. Wait at least 5 sec
10. Turn on server 2
11. Wait until servers starts
At this point all connections will be on server 1
12. Trigger `forceReconnect` on all opened connections

Here's the [test case](https://github.com/ajax-koval-d/nats.java/pull/1) (`testRebalanceWithForceReconnect`) that demonstrates the issue.

ajax-koval-d:
@scottf, Thanks for quick response!

I'd like to clarify a bit my idea: what I want is to call `Collections.shuffle` in a middle of a `forceReconnect`, right before `reconnectImpl`, but without putting current server on the bottom, [like this](https://github.com/ajax-koval-d/nats.java/blob/1f1b82424f49c8d4036269ef0a31ceccba0b8f04/src/main/java/io/nats/client/impl/NatsConnection.java#L339). I don't want to change `NatsServerPool` or `reconnectImpl` logic in any way, so standard behavior of reconnect, when connection, for example, disconnected due to network issues will be kept intact - currently connected server will be last on the list.

[...]

PR #1190, opened 2024-07-30
Add option to `forceReconnect` API to guarantee fair server distribution

ajax-koval-d:
This addresses https://github.com/nats-io/nats.java/issues/1184

### DIFF
--- a/src/test/java/io/nats/client/impl/ReconnectTests.java
+++ b/src/test/java/io/nats/client/impl/ReconnectTests.java
@@ -722,6 +722,30 @@ public class ReconnectTests {
     }
 
     @Test
+    void testForceReconnectWithFairDistribution() throws Exception {
+        ListenerForTesting listener = new ListenerForTesting();
+        ThreeServerTestOptions tstOpts = makeThreeServerTestOptions(listener, false);
+        runInJsCluster(tstOpts, (nc0, nc1, nc2) -> {
+            for (int i = 0; i < 10; i++) {
+                listener.getConnectionEvents().clear();
+                ServerInfo si = nc0.getServerInfo();
+                String connectedServer = si.getServerId();
+
+                nc0.forceReconnect(ForceReconnectOptions.builder().fairServerDistribution().build());
+                standardConnectionWait(nc0);
+
+                si = nc0.getServerInfo();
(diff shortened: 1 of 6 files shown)


src/test/java/io/nats/client/impl/ReconnectTests.java
```diff
@@ -774,23 +798,24 @@ public void testForceReconnectQueueBehaviorCheck() throws Exception {
 
             String subject = subject();
             ForceReconnectQueueCheckDataPort.WRITE_CHECK = "PUB " + subject;
-            _testForceReconnectQueueCheck(subject, pubCount, subscribeTime, port, false, 0);
+            _testForceReconnectQueueCheck(subject, pubCount, subscribeTime, port, false, true, 0);
```
ajax-koval-d, line 801: Did not cover all possible scenarios since in that case locally test was running 17 secs and almost never passed

scottf:
@ajax-koval-d I'm not going to accept the PR for the following reasons.
1. The NatsServerPool implementation matches the behavior in the other clients. Adding behavior that is not standard is confusing to developers who use multiple language clients.
2. There is already a way to provide your own implementation of ServerPool
3. The idea behind force reconnect is not to reconnect to the same server. It's used in error conditions, lame duck mode handling and when you specifically want to disconnect from the current server. Reconnecting to the same server defeats the purpose.
4. For backward compatibility, all new api on existing interfaces must have a default implementation, otherwise it is a breaking change.

scottf:
@ajax-koval-d I am willing to discuss changes to the ServerPool interface that allows signaling of state, giving more flexibility to behavior of ServerPool implementations. For instance shuffle would not make sense, but something that indicates a connection state about change/to change.

ajax-koval-d:
@scottf Regarding point 3 - my impression was, that according to [this adr](https://github.com/nats-io/nats-architecture-and-design/issues/259) one of the purposes of reconnect api is also rebalance, and current implementation does not always fulfill that.

My problem with implementing ServerPool is that I don't need to modify usual reconnect flow, all I want to change is force reconnect behaviour.

> something that indicates a connection state about change/to change

Not sure that I follow you here. Do you mean to provide new method in ServerPool interface, like goingToForceReconnect so that there'll be ability to react on it in implementation?

PR #1191, opened 2024-07-31
Recommended (new) APIs and more example code

roeschter:
This is a test edit to verify style guides. 

Starting to make good on my promise to improve the documentation. 

Purpose:
1. Point users to the recommended (new) way of using Jetstream
2. Add more example code snipplets
3. Make sure such guidelines and code snipplet appear as early as possible in the developers journey. In this case on the Connection class. 
4. More detailed examples would then be in ConsumerContext and StreamContext

### DIFF
--- a/src/main/java/io/nats/client/api/MessageGetRequest.java
+++ b/src/main/java/io/nats/client/api/MessageGetRequest.java
@@ -44,6 +44,8 @@ public class MessageGetRequest implements JsonSerializable {
 
     /**
      * @deprecated use static method forSequence with .serialize instead
+     * @param sequence start sequence
+     * @return rendered output
      */
     @Deprecated
     public static byte[] seqBytes(long sequence) {
@@ -52,6 +54,8 @@ public class MessageGetRequest implements JsonSerializable {
 
     /**
      * @deprecated use static method lastForSubject with .serialize instead
+     * @param subject filter subject
+     * @return rendered output
      */
     @Deprecated
     public static byte[] lastBySubjectBytes(String subject) {
@@ -60,6 +64,8 @@ public class MessageGetRequest implements JsonSerializable {
 
     /**
      * @deprecated use static method forSequence instead
+     * 
+     * @param sequence start sequence number
      */
     @Deprecated
     public MessageGetRequest(long sequence) {
@@ -68,6 +74,8 @@ public class MessageGetRequest implements JsonSerializable {
 
     /**
      * @deprecated use static method lastForSubject instead
+     * 
+     * @param lastBySubject filter subject
      */
     @Deprecated
     public MessageGetRequest(String lastBySubject) {
--- a/src/main/java/io/nats/client/impl/Headers.java
+++ b/src/main/java/io/nats/client/impl/Headers.java
@@ -263,16 +263,16 @@ public class Headers {
 	/**
 	 * Returns the number of keys (case-sensitive) in the header.
 	 *
-	 * @return the number of keys
+	 * @return the number of header entries
 	 */
 	public int size() {
 		return valuesMap.size();
 	}
 
 	/**
-	 * Returns trtruethis map contains no keys.
+	 * Returns ture if map contains no keys.
 	 *
-	 * @return true if this map contains no keys
+	 * @return true if there are no headers
 	 */
 	public boolean isEmpty() {
 		return valuesMap.isEmpty();
@@ -342,7 +342,7 @@ public class Headers {
 	 * Returns a {@link List} view of the values for the specific (case-sensitive) key.
 	 * Will be {@code null} if the key is not found.
 	 *
-	 * @param key get entry by key
+	 * @param key the key whose associated value is to be returned
 	 * @return a read-only list of the values for the case-sensitive key.
 	 */
 	public List<String> get(String key) {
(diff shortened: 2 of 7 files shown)


scottf (approved):
LGTM

roeschter:
Will fix Javadoc errors

roeschter:
Fixed various Javadoc errors and warning. Added links to overview.html
Javadoc processes without wanrings
Done editing - will continue on another branch.

scottf:
@roeschter There is a test failing, but I fixed and merged it a few days ago. Can you update your branch

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -269,9 +270,9 @@ public int size() {
 	}
 
 	/**
-	 * Returns <tt>true</tt> if this map contains no keys.
+	 * Returns trtruethis map contains no keys.
```
scottf, line 273: Not sure what was meant here.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -341,6 +342,7 @@ public Set<String> keySetIgnoreCase() {
 	 * Returns a {@link List} view of the values for the specific (case-sensitive) key.
 	 * Will be {@code null} if the key is not found.
 	 *
+	 * @param key get entry by key
```
scottf, line 345: How about "The key whose associated value is to be returned"

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -351,7 +353,7 @@ public List<String> get(String key) {
 	/**
 	 * Returns the first value for the specific (case-sensitive) key.
 	 * Will be {@code null} if the key is not found.
-	 *
+	 * @param key get entry by key
```
scottf, line 356: ditto

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -363,6 +365,7 @@ public String getFirst(String key) {
 	 * Returns the last value for the specific (case-sensitive) key.
 	 * Will be {@code null} if the key is not found.
 	 *
+	 * @param key get entry by key
```
scottf, line 368: ditto ditto

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -374,6 +377,7 @@ public String getLast(String key) {
 	 * Returns a {@link List} view of the values for the specific (case-insensitive) key.
 	 * Will be {@code null} if the key is not found.
 	 *
+	 * @param key get entry by key
```
scottf, line 380: you get the idea

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -448,6 +452,8 @@ public byte[] getSerialized() {
 	/**
 	 * @deprecated
 	 * Appends the serialized bytes to the builder.
+	 * 
+	 * @param bab the ByteArrayBuilder to append
```
scottf, line 456: This isn't used and probably was public only for testing, it was unfortunate I released it this way, but I can't take it back. I would just make this super deprecated

src/main/javadoc/overview.html
```diff
@@ -17,6 +17,10 @@
 <head></head>
 <body>
 Java API for the <a href="http://nats.io">NATs messaging system</a>.
+<p>The Java API is hosted on <a href="https://github.com/nats-io/nats.java">Github NATs Java API</a>. Please check the <a href="https://github.com/nats-io/nats.java/blob/main/README.md"><b>README</b></a> for updates, hints and best practices.
+<p>Examples are included <a href="https://github.com/nats-io/nats.java/tree/main/src/examples/java/io/nats/examples">NATs Client API examples</a>.
+<p>More examples, including other programming languages, at <a href="https://natsbyexample.com/">NATs By Example</a>.
+
```
scottf, line 23: Can you please add this also: https://github.com/nats-io/java-nats-examples. By github convention, if README.MD is found in the root of the repo, it is shown on the main page of the repo, so I don't know if you need to reference it's actual page.

src/main/java/io/nats/client/api/MessageGetRequest.java
```diff
@@ -52,6 +54,8 @@ public static byte[] seqBytes(long sequence) {
 
     /**
      * @deprecated use static method lastForSubject with .serialize instead
+     * @param subject filter subject
+     * @return rendered output
```
scottf, line 58: I personally hate having comments on deprecated api because it might encourage their use. It might be good to have proper links to the replacement though. There are a couple items like this.

scottf (changes requested):

src/main/javadoc/overview.html
roeschter: Will add to NATS by example.
I was pointing to README explicitely because that makes it harder to miss the instructions. When following the main project link the code is on the top.

src/main/java/io/nats/client/api/MessageGetRequest.java
roeschter: I know, I just wanted to get rid of the warnings.

PR #1193, opened 2024-08-08
Jetstream Usage

roeschter:
Inline Examples

### DIFF
--- a/src/main/java/io/nats/client/JetStream.java
+++ b/src/main/java/io/nats/client/JetStream.java
@@ -28,25 +28,26 @@ import java.util.concurrent.CompletableFuture;
  * 
  * <h3>Basic usage</h3>
  * 
- * <p>{@link #publish(String, byte[]) JetStream.Publish} will send a message on the specified subject as a Req/Rep waiting for acknowledgement. 
- * An error will be received if no stream is litetning on said subject.
- * 
- * <p>{@link Connection#publish(String, byte[]) Connection.Publish} can be used as well to send to a stream on a subject, but no confirmation or error will come back. 
- * Use non acknowledged publish with caution as it is easy to overwhelm Jetstream this way.
+ * <p>{@link #publish(String, byte[]) JetStream.Publish} will send a message on the specified subject, waiting for acknowledgement. 
+ * A  <b>503 No responders</b> error will be received if no stream is listening on said subject.
  * 
  * <p>{@link #publishAsync(String, byte[]) PublishAsync} will not wait for acknowledgement but return a {@link CompletableFuture CompletableFuture}, 
  * which can be checked for acknowledgement at a later point.
  * 
- * <p> Use {@link #getStreamContext(String ) getStreamContext(String)} for <b>consuming/subscribing</b> messages from Jetstream. 
+ * <p> Use {@link #getStreamContext(String ) getStreamContext(String)} to access a simplified API for <b>consuming/subscribing</b> messages from Jetstream. 
  * It is <b>recommened</b> to manage consumers explicitely through {@link StreamContext StreamContext} or {@link JetStreamManagement JetStreamManagement}
  * 
- * <p>{@link #subscribe(String)} is a convenience method for implicitely creating a consumer on a stream and receiving messages. 
- * This is simpler but eventually more prone to configuration errors than the {@link ConsumerContext ConsumerContext} based subscription. We recommend to manage consumers explcitely.
+ * <p>{@link #subscribe(String)} is a convenience method for implicitly creating a consumer on a stream and receiving messages. This method should be used for ephemeral (not durable) conusmers. 
+ * It can create a named durable consumers though Options, but we prefer to avoid creating durable consumers implictly. 
+ * It is <b>recommened</b> to manage consumers explicitely through {@link StreamContext StreamContext} or {@link JetStreamManagement JetStreamManagement}
+ *  
+ *  {@link ConsumerContext ConsumerContext} based subscription. 
  * 
- * <h3>Creating Streams, consumer and publishing</h3>
+ * <h3>Recommended usage for creating streams, consumers, publish and listen on a stream</h3>
  * <pre>
  *  io.nats.client.Connection nc = Nats.connect();
- *		
+ *	
+ *  //Setting up a stream and a consumer	
  *  JetStreamManagement jsm = nc.jetStreamManagement();
  *  StreamConfiguration sc = StreamConfiguration.builder()
  *    .name("my-stream")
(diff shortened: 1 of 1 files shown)


src/main/java/io/nats/client/JetStream.java
```diff
@@ -13,15 +13,91 @@
...
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 
 /**
- * JetStream context for creation and access to streams and consumers in NATS.
+ * JetStream context for access to streams and consumers in NATS.
+ * 
+ * <h3>Basic usage</h3>
+ * 
+ * <p>{@link #publish(String, byte[]) JetStream.Publish} will send a message on the specified subject as a Req/Rep waiting for acknowledgement.
```
scottf, line 31: publish is different than request, I think it would better to talk about them separately.

src/main/java/io/nats/client/JetStream.java
```diff
@@ -13,15 +13,91 @@
...
 import java.util.concurrent.CompletableFuture;
 
 /**
- * JetStream context for creation and access to streams and consumers in NATS.
+ * JetStream context for access to streams and consumers in NATS.
+ * 
+ * <h3>Basic usage</h3>
+ * 
+ * <p>{@link #publish(String, byte[]) JetStream.Publish} will send a message on the specified subject as a Req/Rep waiting for acknowledgement. 
+ * An error will be received if no stream is litetning on said subject.
```
scottf, line 32: 1. listening
2. 503 No responders

src/main/java/io/nats/client/JetStream.java
```diff
@@ -13,15 +13,91 @@
...
+ * 
+ * <p>{@link Connection#publish(String, byte[]) Connection.Publish} can be used as well to send to a stream on a subject, but no confirmation or error will come back. 
+ * Use non acknowledged publish with caution as it is easy to overwhelm Jetstream this way.
```
scottf, line 35: "Use core publish..."

Not sure if you are aware of JetStreamOptions.publishNoAck which does a core publish and returns null for the PublishAck

Now that I said that, please remove the whole thing. Core publish to a stream is a footgun and we don't want to advertise it.

src/main/java/io/nats/client/JetStream.java
```diff
@@ -13,15 +13,91 @@
...
+ * 
+ * <p>{@link #publishAsync(String, byte[]) PublishAsync} will not wait for acknowledgement but return a {@link CompletableFuture CompletableFuture}, 
+ * which can be checked for acknowledgement at a later point.
+ * 
+ * <p> Use {@link #getStreamContext(String ) getStreamContext(String)} for <b>consuming/subscribing</b> messages from Jetstream.
```
scottf, line 40: Maybe add something before this to mention "Simplified Consuming" and explain what it is.

src/main/java/io/nats/client/JetStream.java
```diff
@@ -13,15 +13,91 @@
...
+ * <p>{@link Connection#publish(String, byte[]) Connection.Publish} can be used as well to send to a stream on a subject, but no confirmation or error will come back. 
+ * Use non acknowledged publish with caution as it is easy to overwhelm Jetstream this way.
+ * 
+ * <p>{@link #publishAsync(String, byte[]) PublishAsync} will not wait for acknowledgement but return a {@link CompletableFuture CompletableFuture}, 
+ * which can be checked for acknowledgement at a later point.
+ * 
+ * <p> Use {@link #getStreamContext(String ) getStreamContext(String)} for <b>consuming/subscribing</b> messages from Jetstream. 
+ * It is <b>recommened</b> to manage consumers explicitely through {@link StreamContext StreamContext} or {@link JetStreamManagement JetStreamManagement}
+ * 
+ * <p>{@link #subscribe(String)} is a convenience method for implicitely creating a consumer on a stream and receiving messages.
```
scottf, line 43: implicitly

src/main/java/io/nats/client/JetStream.java
```diff
@@ -13,15 +13,91 @@
...
+ * Use non acknowledged publish with caution as it is easy to overwhelm Jetstream this way.
+ * 
+ * <p>{@link #publishAsync(String, byte[]) PublishAsync} will not wait for acknowledgement but return a {@link CompletableFuture CompletableFuture}, 
+ * which can be checked for acknowledgement at a later point.
+ * 
+ * <p> Use {@link #getStreamContext(String ) getStreamContext(String)} for <b>consuming/subscribing</b> messages from Jetstream. 
+ * It is <b>recommened</b> to manage consumers explicitely through {@link StreamContext StreamContext} or {@link JetStreamManagement JetStreamManagement}
+ * 
+ * <p>{@link #subscribe(String)} is a convenience method for implicitely creating a consumer on a stream and receiving messages. 
+ * This is simpler but eventually more prone to configuration errors than the {@link ConsumerContext ConsumerContext} based subscription. We recommend to manage consumers explcitely.
```
scottf, line 44: Remove "prone to errors". This is great for ephemeral (temporary) consumers, but discouraged for durable consumers and as you say we recommend to manage consumers "explicitly" (vs "explcitely")

src/main/java/io/nats/client/JetStream.java
```diff
@@ -13,15 +13,91 @@
...
+ *       
+ *  ConsumerConfiguration consumerConfig = ConsumerConfiguration.builder()
+ *    .durable("my-consumer")
+ *    .build();
+ *		
+ *  jsm.createConsumer("my-stream", consumerConfig);
+ *    	
+ *  io.nats.client.JetStream js = nc.jetStream();
+ *  ConsumerContext consumerContext = js.getConsumerContext("my-stream", "my-consumer"); 
+ *	consumerContext.consume(
```
scottf, line 67: I think any example of consume needs to show
```
MessageConsumer mc = consumerContext.consume
```

I'm not entirely sure if se should add an example here, maybe can we reference an example or something instead.

src/main/java/io/nats/client/JetStream.java
```diff
@@ -13,15 +13,91 @@
...
+ *    msg -&gt; {
+ *      System.out.println("   Received " + msg.getSubject());
+ *      msg.ack();
+ *    });
+ *    	
+ *  js.publish("foo.joe", "Hello World".getBytes());
+ *  </pre>
+ *  
+ *	<h3>Asynchronous publishing</h3>
```
scottf, line 76: Ditto on the example here. Are the api docs the right place for examples?

scottf (changes requested):
Comments and let's discuss having examples, I'm -1 for adding them here.

src/main/java/io/nats/client/JetStream.java
roeschter: OK

src/main/java/io/nats/client/JetStream.java
roeschter: Good point, removed, the whole idea is to show people the most common and correct usage patterns.

src/main/java/io/nats/client/JetStream.java
roeschter: Usually I want ready to run examples outside of the API docs. We have an issue with Jetstream though as the recommended usage pattern is NOT obvious. 
I think this will the biggest example we will have. In all other place it should be back to 3-5 line code fragments

roeschter: If we move the examples, I want to keep a big big link back to the Github location. And at least a few lines of pseudo code to hint at the usage pattern and then link saying: "See full code eaxmples here"

roeschter: A short pseudo code version could be:

//For full example see: <Github> link

Connection nc = Nats.connect();
//Setting up a stream and a consumer	

StreamConfiguration sc = StreamConfiguration.builder() ......
jsm.addStream(sc);
       
ConsumerConfiguration consumerConfig = ConsumerConfiguration.builder()  ....
 jsm.createConsumer("my-stream", consumerConfig);
     	
//Listening and publishing  

ConsumerContext consumerContext = js.getConsumerContext("my-stream", "my-consumer"); 
MessageConsumer mc = consumerContext.consume(  new Handler( .... ) );
  	
 js.publish("foo.joe", "Hello World".getBytes());

scottf (approved):
LGTM

PR #1208, opened 2024-08-19
Removing writerLock in favor of setting a flag to call flushBuffer

scottf:
* Instead of using writerLock around flushBuffer (socket flush), set a flag to tell the sendMessageBatch to flushBuffer after a write, instead of locking around send.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnectionWriter.java
+++ b/src/main/java/io/nats/client/impl/NatsConnectionWriter.java
@@ -174,13 +174,6 @@ class NatsConnectionWriter implements Runnable {
             dataPort.write(sendBuffer, sendPosition);
         }
 
-        try {
-            if (flushBuffer.get()) {
-                dataPort.flush();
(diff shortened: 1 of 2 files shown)


src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -122,60 +123,65 @@ boolean isRunning() {
...
-            connection.getNatsStatistics().registerWrite(sendPosition);
+            stats.incrementOutMsgs();
+            stats.incrementOutBytes(size);
+
+            msg = msg.next;
         }
-        finally {
-            writerLock.unlock();
+
+        // no need to write if there are no bytes
```
scottf, line 172: I added this. It makes sense that there could be times when there are no messages pending so why try to write or even call write. There is probably a lower level check in the write call, but since DataPort is an interface, it won't hurt

MauriceVanVeen (approved):
LGTM

PR #1221, opened 2024-09-03
Javadocs03

roeschter:
Misc documentation changes regarding new style API-
Fail fast on createConsumerContext when a push consumer is provided
- Added a proper test for failure - But tested so far with my own test code only

### DIFF
--- a/src/main/java/io/nats/client/BaseConsumerContext.java
+++ b/src/main/java/io/nats/client/BaseConsumerContext.java
@@ -18,7 +18,7 @@ import java.time.Duration;
 
 /**
  * The Consumer Context provides a convenient interface around a defined JetStream Consumer
- * <p> Note: ConsumerContext requires a <b>pull consumer</b>. 
+ * <p> Note: ConsumerContext requires a <b>pull consumer</b>.
  */
 public interface BaseConsumerContext {
     /**
--- a/src/main/java/io/nats/client/BaseConsumerContext.java
+++ b/src/main/java/io/nats/client/BaseConsumerContext.java
@@ -18,6 +18,7 @@ import java.time.Duration;
 
 /**
  * The Consumer Context provides a convenient interface around a defined JetStream Consumer
+ * <p> Note: ConsumerContext requires a <b>pull consumer</b>. 
  */
 public interface BaseConsumerContext {
     /**
--- a/src/main/java/io/nats/client/JetStream.java
+++ b/src/main/java/io/nats/client/JetStream.java
@@ -25,75 +25,75 @@ import java.util.concurrent.CompletableFuture;
 
 /**
  * JetStream context for access to streams and consumers in NATS.
- * 
+ *
  * <h3>Basic usage</h3>
- * 
- * <p>{@link #publish(String, byte[]) JetStream.Publish} will send a message on the specified subject, waiting for acknowledgement. 
+ *
+ * <p>{@link #publish(String, byte[]) JetStream.Publish} will send a message on the specified subject, waiting for acknowledgement.
  * A  <b>503 No responders</b> error will be received if no stream is listening on said subject.
- * 
- * <p>{@link #publishAsync(String, byte[]) PublishAsync} will not wait for acknowledgement but return a {@link CompletableFuture CompletableFuture}, 
+ *
+ * <p>{@link #publishAsync(String, byte[]) PublishAsync} will not wait for acknowledgement but return a {@link CompletableFuture CompletableFuture},
  * which can be checked for acknowledgement at a later point.
- * 
- * <p> Use {@link #getStreamContext(String ) getStreamContext(String)} to access a simplified API for <b>consuming/subscribing</b> messages from Jetstream. 
+ *
+ * <p> Use {@link #getStreamContext(String ) getStreamContext(String)} to access a simplified API for <b>consuming/subscribing</b> messages from Jetstream.
  * It is <b>recommened</b> to manage consumers explicitely through {@link StreamContext StreamContext} or {@link JetStreamManagement JetStreamManagement}
- * 
- * <p>{@link #subscribe(String)} is a convenience method for implicitly creating a consumer on a stream and receiving messages. This method should be used for ephemeral (not durable) conusmers. 
- * It can create a named durable consumers though Options, but we prefer to avoid creating durable consumers implictly. 
+ *
+ * <p>{@link #subscribe(String)} is a convenience method for implicitly creating a consumer on a stream and receiving messages. This method should be used for ephemeral (not durable) conusmers.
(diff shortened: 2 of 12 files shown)


src/main/java/io/nats/client/JetStream.java
```diff
@@ -39,9 +39,8 @@
  * 
  * <p>{@link #subscribe(String)} is a convenience method for implicitly creating a consumer on a stream and receiving messages. This method should be used for ephemeral (not durable) conusmers. 
  * It can create a named durable consumers though Options, but we prefer to avoid creating durable consumers implictly. 
- * It is <b>recommened</b> to manage consumers explicitely through {@link StreamContext StreamContext} or {@link JetStreamManagement JetStreamManagement}
+ * It is <b>recommened</b> to manage consumers explicitely through {@link StreamContext StreamContext} and {@link ConsumerContext ConsumerContext} or {@link JetStreamManagement JetStreamManagement}
```
scottf, line 42: "explicitly" https://www.google.com/search?q=explicitely

src/main/java/io/nats/client/BaseConsumerContext.java
```diff
@@ -18,6 +18,7 @@
 
 /**
  * The Consumer Context provides a convenient interface around a defined JetStream Consumer
+ * <p> Note: ConsumerContext requires a <b>pull consumer</b>.
```
scottf, line 21: So this may eventually not be true. For instance I know Alberto and javascript are working on supporting push consumers. And also, it takes effort to make a push consumer because you have to specifically add the delivery subject. I'm not sure if this is necessary, so maybe just let me know what led you to add this? I guess we can go back if we do support push

src/main/java/io/nats/client/JetStreamReader.java
```diff
@@ -16,7 +16,8 @@
 import java.time.Duration;
 
 /**
- * This interface provides push like ability for a pull consumer. Was the pre-cursor to simplified consume.
+ * This interface provides a simple iterative access to a pull consumer. 
+ * <p>Note: This interface is superseded by  {@link ConsumerContext ConsumerContext}. For examples for <b>recommended usage</b> see {@link JetStream JetStream}.
```
scottf, line 20: This is basically deprecated because of simplifications. Not sure I want anything to call it out.

src/main/java/io/nats/client/api/ConsumerConfiguration.java
```diff
@@ -30,8 +32,10 @@
 
 /**
  * The ConsumerConfiguration class specifies the configuration for creating a JetStream consumer on the client and
- * if necessary the server.
+ * if necessary the server.
```
scottf, line 35: Can you set your editor to trim trailing spaces?

src/main/java/io/nats/client/api/Error.java
```diff
@@ -32,7 +32,12 @@ public class Error implements JsonSerializable {
     static Error optionalInstance(JsonValue vError) {
         return vError == null ? null : new Error(vError);
     }
-
+    
+    public static Error instance(int code, String description) {
+        return new Error(code, NOT_SET, description);
+    }
+    
+
```
scottf, line 40: Remove. See comment below about NatsJetStreamClientError vs JetStreamApiException

src/main/java/io/nats/client/impl/NatsConsumerContext.java
```diff
@@ -56,6 +60,13 @@ public class NatsConsumerContext implements ConsumerContext, SimplifiedSubscript
         defaultDispatcher = new AtomicReference<>();
         lastConsumer = new AtomicReference<>();
         if (unorderedConsumerInfo != null) {
+        	if (unorderedConsumerInfo != null) {
+                //Fail fast on push consumers as all operation on ConsumerContext expect a pull consumer.
+                //If we don't fail fast, calls to consume() or iterat() will hang with the server returning: code=409, message='Consumer is push based'
+                ConsumerConfiguration consumerConfig = unorderedConsumerInfo.getConsumerConfiguration();
+                if ( consumerConfig.getDeliverSubject() != null )
+                    throw new JetStreamApiException(io.nats.client.api.Error.instance(409, "Consumer is push based. ConsumerContext only supports pull consumers."));
+        	}
```
scottf, line 69: 1. I'm not 100% no but... I'm pretty sure it was discussed in depth during simplification design to not fast fail. Maybe it should change. Also maybe it should be done in the caller. But for now just leave it.
2. JetStreamApiException are reserved for JS errors that come from the server. For errors like this that come from the client, make a NatsJetStreamClientError

src/main/java/io/nats/client/impl/NatsOrderedConsumerContext.java
```diff
@@ -25,7 +25,7 @@
 public class NatsOrderedConsumerContext implements OrderedConsumerContext {
     private final NatsConsumerContext impl;
 
-    NatsOrderedConsumerContext(NatsStreamContext streamContext, OrderedConsumerConfiguration config) {
+    NatsOrderedConsumerContext(NatsStreamContext streamContext, OrderedConsumerConfiguration config) throws JetStreamApiException {
```
scottf, line 28: Remove.

scottf (changes requested):
Minor spelling. Please remove all functional changes to a different PR so we can move ahead with the doc changes.

src/main/java/io/nats/client/BaseConsumerContext.java
roeschter: It is not obvious that that push consumers don't work and can result in late errors, late during application deployment. This is driven by a lot of mix-ups between push and pull consumer I see from user in slack.
Reading just the documentation I initially assumed that ConsumerContext would support push and pull and was quite a bit surprised.

src/main/java/io/nats/client/JetStreamReader.java
roeschter: Yes, I want to call it out. Either we dare to say deprecated or euphemistically "superseded".

roeschter:
OK, split the functional changes. Cleaned up trailing white space.

src/main/java/io/nats/client/JetStream.java
roeschter: Fixed spelling

src/main/java/io/nats/client/api/ConsumerConfiguration.java
roeschter: Done

src/main/java/io/nats/client/impl/NatsConsumerContext.java
roeschter: OK, moved to another branch. You can in that branch.

scottf (approved):
LGTM

PR #1236, opened 2024-10-08
Easier access to timeout in JetStream implementations

scottf:
JetStream implementation code relies on the the timeout, first from any supplied JetStreamOptions, second from connection options. This just simplifies, so code is not duplicated and can be retrieved via one method that understands the priority.


MauriceVanVeen (approved):
LGTM

PR #1238, opened 2024-10-10
Direct Batch EOD and loop changes

scottf:


src/main/java/io/nats/client/impl/NatsJetStreamManagement.java
```diff
@@ -351,12 +351,8 @@ private MessageInfo _getMessage(String streamName, MessageGetRequest messageGetR
...
     public List<MessageInfo> fetchMessageBatch(String streamName, MessageBatchGetRequest messageBatchGetRequest) throws IOException, JetStreamApiException {
         validateMessageBatchGetRequest(streamName, messageBatchGetRequest);
-        List<MessageInfo> results = new ArrayList<>();
-        _requestMessageBatch(streamName, messageBatchGetRequest, msg -> {
-            if (msg != MessageInfo.EOD) {
-                results.add(msg);
-            }
-        });
+        final List<MessageInfo> results = new ArrayList<>();
+        _requestMessageBatch(streamName, messageBatchGetRequest, false, results::add);
```
scottf, line 355: false here will prevent sending the EOD since we don't need it.

src/main/java/io/nats/client/impl/NatsJetStreamManagement.java
```diff
@@ -367,7 +363,7 @@ public List<MessageInfo> fetchMessageBatch(String streamName, MessageBatchGetReq
     public LinkedBlockingQueue<MessageInfo> queueMessageBatch(String streamName, MessageBatchGetRequest messageBatchGetRequest) throws IOException, JetStreamApiException {
         validateMessageBatchGetRequest(streamName, messageBatchGetRequest);
         final LinkedBlockingQueue<MessageInfo> q = new LinkedBlockingQueue<>();
-        conn.getOptions().getExecutor().submit(() -> _requestMessageBatch(streamName, messageBatchGetRequest, q::add));
+        conn.getOptions().getExecutor().submit(() -> _requestMessageBatch(streamName, messageBatchGetRequest, true, q::add));
```
scottf, line 366: queue needs the EOD, so used true

src/main/java/io/nats/client/impl/NatsJetStreamManagement.java
```diff
@@ -377,32 +373,31 @@ public LinkedBlockingQueue<MessageInfo> queueMessageBatch(String streamName, Mes
     @Override
     public void requestMessageBatch(String streamName, MessageBatchGetRequest messageBatchGetRequest, MessageInfoHandler handler) throws IOException, JetStreamApiException {
         validateMessageBatchGetRequest(streamName, messageBatchGetRequest);
-        _requestMessageBatch(streamName, messageBatchGetRequest, handler);
+        _requestMessageBatch(streamName, messageBatchGetRequest, true, handler);
```
scottf, line 376: request always sends EOD to the user

src/main/java/io/nats/client/impl/NatsJetStreamManagement.java
```diff
@@ -377,32 +373,30 @@ public LinkedBlockingQueue<MessageInfo> queueMessageBatch(String streamName, Mes
...
                 }
                 if (msg.isStatusMessage()) {
                     Status status = msg.getStatus();
                     // Report error, otherwise successful status.
-                    if (status.getCode() < 200 || status.getCode() > 299) {
-                        MessageInfo messageInfo = new MessageInfo(Error.convert(status), true);
-                        handler.onMessageInfo(messageInfo);
+                    if (status.getCode() != Status.EOB) {
+                        handler.onMessageInfo(new MessageInfo(Error.convert(status), true));
+                        sendEod = false; // the error is the EOD since we always end on error
```
scottf, line 399: 1. The only time status is not an error is if it is specifically the EOB from the server. We know nothing about any other status functionality, so for now send anything else, even 20x to the user. We need to raise this issue in the go client, because I want to know what other 20x they are expecting

scottf: 2. There is no need to send both an EOD and an error. The error is enough to signify the request has ended.

src/main/java/io/nats/client/impl/NatsJetStreamManagement.java
```diff
@@ -377,32 +373,30 @@ public LinkedBlockingQueue<MessageInfo> queueMessageBatch(String streamName, Mes
...
 
             String requestSubject = prependPrefix(String.format(JSAPI_DIRECT_GET, streamName));
             conn.publish(requestSubject, replyTo, messageBatchGetRequest.serialize());
 
-            long maxTimeMillis = getTimeout().toMillis();
-            long timeLeft = maxTimeMillis;
-            long start = System.currentTimeMillis();
             while (true) {
-                Message msg = sub.nextMessage(timeLeft);
+                Message msg = sub.nextMessage(getTimeout());
```
scottf, line 390: Since there is a sentinel, one timeout is sufficient. We don't need all the timeout calculations.

src/main/java/io/nats/client/impl/NatsJetStreamManagement.java
```diff
@@ -414,18 +408,21 @@ public void _requestMessageBatch(String streamName, MessageBatchGetRequest messa
...
         } finally {
-            try {
-                handler.onMessageInfo(MessageInfo.EOD);
-            } catch (Exception ignore) {
+            if (sendEod) {
+                try {
+                    handler.onMessageInfo(MessageInfo.EOD);
+                }
+                catch (Exception ignore) {
+                }
```
scottf, line 425: don't send the EOD if the request didn't ask (fetch) or in the case of an abnormal status

MauriceVanVeen (approved):
LGTM

scottf:
closed to change branch name

PR #1251, opened 2024-11-14
Fix simplified ordered consuming when a delivery policy was set.

scottf:


roeschter:
All fine one project side as well

roeschter (approved):
Looks good - tests flawlessly in chaos test with many  for the sample code where we first saw the issue. Survives reconnects.

PR #1253, opened 2024-11-15
Fix: exception on simple consume when disconnected immediately

roeschter:
Fixed exception handling. Making sure the original connection exception is visible.


scottf (approved):
LGTM

PR #1254, opened 2024-11-20
feat(nats-connection): implement named executor thread factories

krzysiekigry:
goodbye, "pool-2-thread-1"

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -38,6 +38,7 @@ import java.time.format.DateTimeParseException;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
 
 import static io.nats.client.support.Encoding.*;
 import static io.nats.client.support.NatsConstants.*;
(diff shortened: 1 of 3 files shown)


scottf:
@krzysiekigry So the idea is fine, but it needs to be done in an optional way, since someone might currently rely on that naming scheme in their existing application / logs. 

In the Options Builder, you can call `public Builder executor(ExecutorService executor)`, so maybe add these:
```
public Builder callbackExecutor(ExecutorService executor) 
public Builder connectExecutor(ExecutorService executor) 
```
With the defaults matching exactly the existing i.e. `newSingleThreadExecutor()`

And so starting at line 159 of NatsConnection, that block of code would look like this:
```
this.executor = options.getExecutor();
this.callbackRunner = options.getCallbackExecutor();
this.connectExecutor = options.getConnectExecutor();
```

---

Also, please sign your commits. Here is a good link to help with that: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

scottf:
> Thanks for the suggestion, but I believe this approach works better. When NatsConnection shuts down, it terminates the executors. This can cause issues when attempting to create a new NatsConnection, as the executors would already be shut down. Please let me know if I’m mistaken.

Factory is a good idea.

PR #1256, opened 2024-12-03
NoWait documentation

roeschter:
Clarified NoWait - saw this causing confusion in customer. In prarticular the difference between NoWait with and without expires


scottf (approved):
LGTM

PR #1257, opened 2024-12-04
Executor and Executor Factories in Options can be created via properties

scottf:
Added the ability for the Options builder to accept properties for the Executor Service implementation, the Connect thread executor factory and the Callback thread executor factory.


aricart (approved):
LGTM

PR #1261, opened 2024-12-13
Add fix to abort unnecessary connection attempts

ryad-eldajani-usv:
Currently, the client library is attempting to connect to all available servers in the client configuration. This is correct behavior e.g. in case of a technical issue.

In case of an authentication error however, the client library should not be required to retry a connection to another server as the authentication is going to fail again.


scottf:
The current behavior is intentional.

PR #1263, opened 2024-12-17
Remove Schema related leftovers

### LINKED ISSUE #1262: Remove Schema related leftovers
### Proposed change

According to https://github.com/nats-io/nats-architecture-and-design/issues/220 everything related to the Schema in the service should've been removed, but I found some leftovers.

I'll send a PR with everything removed.

### Use case

Consistency with the spec.

### Contribution

_No response_

d0x7:
Closes #1262

### DIFF
--- a/src/main/java/io/nats/client/support/ApiConstants.java
+++ b/src/main/java/io/nats/client/support/ApiConstants.java
@@ -172,6 +172,7 @@ public interface ApiConstants {
     String RESPONSE          = "response";
     String RETENTION         = "retention";
     String SAMPLE_FREQ       = "sample_freq";
+    String SCHEMA            = "schema";
     String SEALED            = "sealed";
     String SEQ               = "seq";
     String SERVER_ID         = "server_id";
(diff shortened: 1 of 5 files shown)


src/main/java/io/nats/client/support/ApiConstants.java
```diff
@@ -172,7 +172,6 @@ public interface ApiConstants {
     String RESPONSE          = "response";
     String RETENTION         = "retention";
     String SAMPLE_FREQ       = "sample_freq";
-    String SCHEMA            = "schema";
```
scottf, line 175: This is a public variable and cannot be removed due to semver. It could be marked deprecated, but it's not hurting anything, so just remove the removal.

src/main/java/io/nats/service/ServiceBuilder.java
```diff
@@ -129,16 +128,6 @@ public ServiceBuilder infoDispatcher(Dispatcher infoDispatcher) {
...
-    /**
-     * Optional dispatcher for the schema service
-     * @param schemaDispatcher the dispatcher
-     * @return the ServiceBuilder
-     */
-    public ServiceBuilder schemaDispatcher(Dispatcher schemaDispatcher) {
-        this.schemaDispatcher = schemaDispatcher;
-        return this;
-    }
-
```
scottf, line 141: Again, because of semver and not wanting to major version, you can deprecate this and make it a no-op, but unfortunately cannot remove a public api.

scottf (changes requested):
Minor changes due to public api change would require a major version change due to semver

PR #1270, opened 2025-01-15
fix: properly propagate `MaxRequestBatch` error in pull subscriptions

### LINKED ISSUE #1269: `MaxRequestBatch` Error Incorrectly Handled as Warning in Pull Subscriptions
### Observed behavior

When using JetStream Pull subscriptions in the Java client, requests that exceed the consumer's configured `maxBatch` limit are incorrectly handled as warnings rather than errors. 

Specifically, when executing `sub.fetch(BATCH_SIZE, Duration.ofSeconds(1))` with a `BATCH_SIZE` greater than the consumer's configured `maxBatch`.

```java
// Configure consumer with maxBatch = 50
ConsumerConfiguration consumerConfig = ConsumerConfiguration.builder()
        .durable(DURABLE_NAME)
        .maxBatch(MAX_BATCH)  // MAX_BATCH = 50
        .build();

// Attempt to fetch with larger batch size
JetStreamSubscription sub = js.subscribe(SUBJECT, 
    PullSubscribeOptions.builder().durable(DURABLE_NAME).build());
List<Message> messages = sub.fetch(BATCH_SIZE, Duration.ofSeconds(1));  // BATCH_SIZE = 100
```

The server correctly returns a 409 conflict with "Exceeded MaxRequestBatch of 50" message, but this error is being suppressed in [PullMessageManager.java](https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/impl/PullMessageManager.java) due to being caught in the generic "Exceeded Max" condition block:

```java
if (statMsg.startsWith("Exceeded Max")
    || statMsg.equals(SERVER_SHUTDOWN)
    || statMsg.equals(LEADERSHIP_CHANGE)
) {
    if (raiseStatusWarnings) {
        conn.executeCallback((c, el) -> el.pullStatusWarning(c, sub, status));
    }
    return STATUS_HANDLED;  // Error suppression occurs here
}
```

### Expected behavior

When attempting to fetch more messages (BATCH_SIZE=100) than the consumer's configured maximum (MAX_BATCH=50), the client should propagate the server's `409` error as a `JetStreamStatusException`. 

This maintains consistency with the Go client's behaviour and allows applications to properly handle batch size violations.

The error should be propagated through the exception handling mechanism:

```java
try {
    List<Message> messages = sub.fetch(BATCH_SIZE, Duration.ofSeconds(1));
} catch (JetStreamStatusException e) {
    // Should receive: "409 Exceeded MaxRequestBatch of 50"
    System.err.println("Error during fetch: " + e.getMessage());
}
```

### Server and client version

- NATS Server: v2.10.14 
- Java Client: 2.20.6
- Go Client: v1.37.0
- CLI Version: v0.0.35

### Host environment

_No response_

### Steps to reproduce

1. Start NATS server with JetStream enabled:
   ```bash
   nats-server -js
   ```
2. Create a JetStream consumer with maxBatch configuration:
```java
ConsumerConfiguration consumerConfig = ConsumerConfiguration.builder()
           .durable("testDurable")
           .maxBatch(50)  // Set maxBatch limit
           .build();
```
3. Attempt to fetch more messages than maxBatch limit:
```java
   JetStreamSubscription sub = js.subscribe(SUBJECT, 
       PullSubscribeOptions.builder().durable("testDurable").build());
   List<Message> messages = sub.fetch(100, Duration.ofSeconds(1));  // Attempt to fetch 100 messages
   ```
4. Expected: Should throw `JetStreamStatusException` with "Exceeded MaxRequestBatch of 50"
   Actual: Error is suppressed and handled as a warning

5. Full reproduction code :

```java
import io.nats.client.*;
import io.nats.client.api.ConsumerConfiguration;
import io.nats.client.api.StreamConfiguration;
import io.nats.client.api.StreamInfo;

import java.time.Duration;
import java.util.List;

public class NatsMaxBatchTest {
    private static final String STREAM_NAME = "testStream";
    private static final String SUBJECT = "testSubject";
    private static final String DURABLE_NAME = "testDurable";
    private static final int BATCH_SIZE = 100; // Batch size larger than maxBatch
    private static final int MAX_BATCH = 50;  // maxBatch set in consumer configuration

        try (Connection nc = Nats.connect("nats://localhost:4222")) {
            // JetStream management and context

            // Create stream
            StreamConfiguration streamConfig = StreamConfiguration.builder()
                    .name(STREAM_NAME)
                    .subjects(SUBJECT)
                    .build();
            StreamInfo streamInfo = jsm.addStream(streamConfig);

            // Create consumer with maxBatch = 50 and matching filterSubject
            ConsumerConfiguration consumerConfig = ConsumerConfiguration.builder()
                    .durable(DURABLE_NAME)
                    .maxBatch(MAX_BATCH)
                    .filterSubject(SUBJECT) // Set filterSubject to match the stream's subject
                    .build();
            jsm.addOrUpdateConsumer(STREAM_NAME, consumerConfig);

            // Publish messages
            for (int i = 0; i < 200; i++) {
                js.publish(SUBJECT, ("Message " + i).getBytes());
            }

            // Subscribe and fetch messages
            JetStreamSubscription sub = js.subscribe(SUBJECT, PullSubscribeOptions.builder()
                    .durable(DURABLE_NAME)
                    .build());

            // Attempt to fetch more than maxBatch
            try {
                List<Message> messages = sub.fetch(BATCH_SIZE, Duration.ofSeconds(1));
                messages.forEach(msg -> {
                    System.out.println(new String(msg.getData()));
                    msg.ack();
                });
            } catch (Exception e) {
                System.err.println("Error during fetch: " + e.getMessage());
            } finally {
                // Clean up
                sub.unsubscribe();
                jsm.deleteStream(STREAM_NAME);
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

somratdutta:
## Problem
The Java client currently suppresses the "Exceeded MaxRequestBatch" error by handling it as a warning in the generic "Exceeded Max" condition block. This leads to inconsistent behavior with other NATS clients (notably Go) and prevents proper error handling at the application level.

## Solution
Modified error handling in `PullMessageManager` to:
1. Specifically handle "Exceeded MaxRequestBatch" as an error condition
2. Properly propagate the error through `JetStreamStatusException`
3. Maintain existing warning handling for other "Exceeded Max" scenarios

### Changes
The `PullMessageManager.java` class is refactored to propagate the Message as proper error instead of suppressing as a warning.

## Impact & Backwards Compatibility
-  No API changes
- Behavior change: MaxRequestBatch errors now properly propagated
- Maintains existing warning handling for other conditions
- Consistent with Go client behavior

## Related Issues
Fixes #1269 

## Future Improvements
Handle other 409 conflicts as errors instead of warning.

### DIFF
--- a/src/main/java/io/nats/client/impl/PullMessageManager.java
+++ b/src/main/java/io/nats/client/impl/PullMessageManager.java
@@ -161,6 +161,11 @@ class PullMessageManager extends MessageManager {
             case CONFLICT_CODE:
                 // sometimes just a warning
                 String statMsg = status.getMessage();
+                //properly handle error & propagate
+                if (statMsg.startsWith("Exceeded MaxRequestBatch")) {
+                        conn.executeCallback((c, el) -> el.pullStatusError(c, sub, status));
+                    return STATUS_ERROR;
+                }
                 if (statMsg.startsWith("Exceeded Max")
                     || statMsg.equals(SERVER_SHUTDOWN)
                     || statMsg.equals(LEADERSHIP_CHANGE)


scottf:
This is already a warning. Changing to an error would actually be an api change as it is behavior that already exists.

I think it might be useful to understand why this is a warning and not an error already, because I understand it does seem like an error. After all, I was just using the consumer, I didn't remove it, why is it gone? But in reality, this should never actually happen, it just that the server allows it to happen and that isn't going to change. But it's a warning since the consumer could be deleted, then it could be re-created and then the consuming would re-start. Both the delete and re-create operations are done by a completely different application than the one that is using it, but this is all legal in NATS, hence the warning.

In regard to flink where this was seen, the running example actually yanks the stream out from under the flink env when the flink env is not finished shutting down. This is a similar problem when we write examples, but forget to sleep while background threads are consuming.

PR #1273, opened 2025-02-10
Ignore Ipv6

aditya-agarwal-groww:


scottf:
@aditya-agarwal-groww 

1. All ip addresses that are sent to connect are already resolved. NatsConnection line 217 calls resolveHost (line 1903). resolveHost calls the existing NatsServerPool.resolveHostToIps.
2. If there is a change to be made to ignore ipv6, it must be an explicit Options setting where the default behavior is the existing behavior. Once there is an option, it could be applied inside the existing NatsServerPool.resolveHostToIps.

Where do the IPV6 addresses come from? Do they come from the bootstrap (unlikely) so they probably come from discovery. A good time to "ignore" these is when discovered addresses are ingested (NatsServerPool.acceptDiscoveredUrls). Also there is already an options setting `ignoreDiscoveredServers`

aditya-agarwal-groww:
@scottf 
1. We are using noResolvedHostNames as true. Reason being we are trying to connect on a wss connection. If the we are resolving the host before hand , we are getting error in upgradeToSecure. While upgradeToSecure the DNS name is being needed.

2. Why are we ignoring ipv6 :- almost 10% of our users are getting connection timeout issues on ipv6 addresses. That why we are trying to add this filter.   Yes we can ignore the ipv6 by adding a flag in options , and it can be taken care in the NatsServerPool.resolveHostToIps. But that needs a fix issue 1. I can help you with the exact error we get on it

Here is the error if I don't use option **noResolvedHostNames**.

javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
	at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:131)
	at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:117)
	at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:347)
	at java.base/sun.security.ssl.Alert$AlertConsumer.consume(Alert.java:293)
	at java.base/sun.security.ssl.TransportContext.dispatch(TransportContext.java:186)
	at java.base/sun.security.ssl.SSLTransport.decode(SSLTransport.java:172)
	at java.base/sun.security.ssl.SSLSocketImpl.decode(SSLSocketImpl.java:1511)
	at java.base/sun.security.ssl.SSLSocketImpl.readHandshakeRecord(SSLSocketImpl.java:1421)
	at java.base/sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:456)
	at java.base/sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:427)
	at io.nats.client.impl.SocketDataPort.upgradeToSecure(SocketDataPort.java:136)
	at io.nats.client.impl.SocketDataPort.connect(SocketDataPort.java:94)
	at io.nats.client.impl.SocketDataPortWithWriteTimeout.connect(SocketDataPortWithWriteTimeout.java:72)
	at io.nats.client.impl.SocketDataPort.connect(SocketDataPort.java:59)
	at io.nats.client.impl.NatsConnection.tryToConnect(NatsConnection.java:545)
	at io.nats.client.impl.NatsConnection.connectImpl(NatsConnection.java:229)
	at io.nats.client.impl.NatsConnection.connect(NatsConnection.java:183)
	at io.nats.client.impl.NatsImpl.createConnection(NatsImpl.java:29)
	at io.nats.client.Nats.createConnection(Nats.java:352)
	at io.nats.client.Nats.connect(Nats.java:250)
	at io.nats.examples.ConnectTime.main(ConnectTime.java:85)
NATS connection exception occurred

src/main/java/io/nats/client/impl/SocketDataPort.java
```diff
@@ -81,7 +83,7 @@ public void connect(NatsConnection conn, NatsUri nuri, long timeoutNanos) throws
             socket.setTcpNoDelay(true);
             socket.setReceiveBufferSize(2 * 1024 * 1024);
             socket.setSendBufferSize(2 * 1024 * 1024);
-            socket.connect(new InetSocketAddress(host, port), (int) timeout);
+            socket.connect(new InetSocketAddress(getIpV4Addresses(nuri).get(0).getHost(), port), (int) timeout);
```
scottf, line 86: No. This is not backward compatible, meaning it could cause unexpected behavior for a user.
1. This must be an option where the default behavior is the same.
2. This should be applied in NatsServerPool.resolveHostToIps

scottf (changes requested):

scottf:
@aditya-agarwal-groww Please see my latest review comments. I can probably make the change faster and let you review.

src/main/java/io/nats/client/impl/SocketDataPort.java
aditya-agarwal-groww: @scottf thanks for the review suggestions 

Regarding point 2
If I make this change in NatsServerPool.resolveHostToIps , I would have to use noResolvedHostNames as false.

But using that option we get the following error

javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:131)
at java.base/sun.security.ssl.Alert.createSSLException(Alert.java:117)
at java.base/sun.security.ssl.TransportContext.fatal(TransportContext.java:347)
at java.base/sun.security.ssl.Alert$AlertConsumer.consume(Alert.java:293)
at java.base/sun.security.ssl.TransportContext.dispatch(TransportContext.java:186)
at java.base/sun.security.ssl.SSLTransport.decode(SSLTransport.java:172)
at java.base/sun.security.ssl.SSLSocketImpl.decode(SSLSocketImpl.java:1511)
at java.base/sun.security.ssl.SSLSocketImpl.readHandshakeRecord(SSLSocketImpl.java:1421)
at java.base/sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:456)
at java.base/sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:427)
at io.nats.client.impl.SocketDataPort.upgradeToSecure(SocketDataPort.java:136)
at io.nats.client.impl.SocketDataPort.connect(SocketDataPort.java:94)
at io.nats.client.impl.SocketDataPortWithWriteTimeout.connect(SocketDataPortWithWriteTimeout.java:72)
at io.nats.client.impl.SocketDataPort.connect(SocketDataPort.java:59)
at io.nats.client.impl.NatsConnection.tryToConnect(NatsConnection.java:545)
at io.nats.client.impl.NatsConnection.connectImpl(NatsConnection.java:229)
at io.nats.client.impl.NatsConnection.connect(NatsConnection.java:183)
at io.nats.client.impl.NatsImpl.createConnection(NatsImpl.java:29)
at io.nats.client.Nats.createConnection(Nats.java:352)
at io.nats.client.Nats.connect(Nats.java:250)
at io.nats.examples.ConnectTime.main(ConnectTime.java:85)
NATS connection exception occurred

This is probably because we now the ips are bing passed to the SocketDataPort , which does hinder upgradeToSecure

aditya-agarwal-groww: Hey @scottf Sorry If I was clear in the start

So we are trying to make a secure websocket connnection. 
Lets say host is wss://abc.in

Now [here](https://github.com/nats-io/nats.java/blob/a48f61eaee0dff4564100255942e4d393c008c2c/src/main/java/io/nats/client/impl/NatsConnection.java#L217) if I resolve my host , i will get the list of Ips to connect to.

Now I try to connect to all ips.

In the call to socketDataPort [here](https://github.com/nats-io/nats.java/blob/a48f61eaee0dff4564100255942e4d393c008c2c/src/main/java/io/nats/client/impl/SocketDataPort.java#L67),  we are just passing the resolved ip now  which sets [this](https://github.com/nats-io/nats.java/blob/a48f61eaee0dff4564100255942e4d393c008c2c/src/main/java/io/nats/client/impl/SocketDataPort.java#L42) to ip
 
Now once the socket connection is made , next step is this https://github.com/nats-io/nats.java/blob/a48f61eaee0dff4564100255942e4d393c008c2c/src/main/java/io/nats/client/impl/SocketDataPort.java#L120. 

Here we are getting the above mentioned error. This could be due to the host passed is the IP,  being used [here](https://github.com/nats-io/nats.java/blob/a48f61eaee0dff4564100255942e4d393c008c2c/src/main/java/io/nats/client/impl/SocketDataPort.java#L127) 

It don't get context of the domain. 

That is why I was making the change in SocketDataPort itself. So that the [host](https://github.com/nats-io/nats.java/blob/a48f61eaee0dff4564100255942e4d393c008c2c/src/main/java/io/nats/client/impl/SocketDataPort.java#L42) var remains wss://abc.in

scottf: So the actual problem might still be related to ipv6, but it's mostly that the wss protocol is lost when resolving the ip. We should fix that problem.

scottf: @aditya-agarwal-groww So the actual problem might still be related to ipv6, but it's mostly that the wss protocol is lost when resolving the ip. We should fix that problem.

aditya-agarwal-groww: OK So I should fix 2 things right ?
1. Wss getting lost on ip resolving 
2. Option to ignore IPv6 while resolving the Ips. 

For 2 instead of ignoring ipv6 , should we look into implementing this : -

We can try parallel connection on the the resolved Ips we get, if any one connects we drop the others and move ahead.
This could be a more generic solve ? [Happy Eyeballs Algorithm ](https://en.wikipedia.org/wiki/Happy_Eyeballs)

scottf:
@aditya-agarwal-groww What happens if you just use the `Options.noResolveHostnames()`

PR #1276, opened 2025-02-17
Feature: Support adding service endpoint after construction (more)

scottf:
Additional testing. Minor tweaks.

### DIFF
--- a/src/main/java/io/nats/service/InfoResponse.java
+++ b/src/main/java/io/nats/service/InfoResponse.java
@@ -16,7 +16,10 @@ package io.nats.service;
 import io.nats.client.support.JsonUtils;
 import io.nats.client.support.JsonValue;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 
 import static io.nats.client.support.ApiConstants.DESCRIPTION;
 import static io.nats.client.support.ApiConstants.ENDPOINTS;
@@ -33,13 +36,21 @@ public class InfoResponse extends ServiceResponse {
     private final String description;
     private final List<Endpoint> endpoints;
 
-    InfoResponse(String id, String name, String version, Map<String, String> metadata, String description, Collection<ServiceEndpoint> serviceEndpoints) {
+    InfoResponse(String id, String name, String version, Map<String, String> metadata, String description) {
         super(TYPE, id, name, version, metadata);
         this.description = description;
         this.endpoints = new ArrayList<>();
-        for (ServiceEndpoint se : serviceEndpoints) {
-            addServiceEndpoint(se);
-        }
+    }
+
+    InfoResponse addServiceEndpoint(ServiceEndpoint se) {
+        endpoints.add(new Endpoint(
+            se.getName(),
+            se.getSubject(),
+            se.getQueueGroup(),
+            se.getMetadata()
+        ));
+        serialized.set(null);
+        return this;
     }
 
     InfoResponse(byte[] jsonBytes) {
@@ -73,19 +84,6 @@ public class InfoResponse extends ServiceResponse {
      */
     public List<Endpoint> getEndpoints() {
         return endpoints;
-    }
-
-    /**
-     * Adds a service endpoint to the list of endpoints.
-     * @param se the service endpoint to be added
-     */
-    public void addServiceEndpoint(ServiceEndpoint se) {
-        endpoints.add(new Endpoint(
-                se.getName(),
-                se.getSubject(),
-                se.getQueueGroup(),
-                se.getMetadata()
-        ));
     }
 
     @Override
--- a/src/main/java/io/nats/service/Service.java
+++ b/src/main/java/io/nats/service/Service.java
@@ -63,32 +63,26 @@ public class Service {
         dInternals = new ArrayList<>();
         startStopLock = new ReentrantLock();
 
+        // build responses first. info needs to be available when adding service endpoints.
+        pingResponse = new PingResponse(id, b.name, b.version, b.metadata);
+        infoResponse = new InfoResponse(id, b.name, b.version, b.metadata, b.description);
+
         // set up the service contexts
-        // ? do we need an internal dispatcher for any user endpoints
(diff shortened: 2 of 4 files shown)


src/main/java/io/nats/service/InfoResponse.java
```diff
@@ -33,13 +36,21 @@ public class InfoResponse extends ServiceResponse {
...
+
+    InfoResponse addServiceEndpoint(ServiceEndpoint se) {
+        endpoints.add(new Endpoint(
+            se.getName(),
+            se.getSubject(),
+            se.getQueueGroup(),
+            se.getMetadata()
+        ));
+        serialized.set(null);
+        return this;
```
scottf, line 53: 1. this method should not be public. Only the service code is supposed to build this.
2. Clear the lazy serialization. when an endpoint is added

src/main/java/io/nats/service/InfoResponse.java
```diff
@@ -75,19 +86,6 @@ public List<Endpoint> getEndpoints() {
...
-     */
-    public void addServiceEndpoint(ServiceEndpoint se) {
-        endpoints.add(new Endpoint(
-                se.getName(),
-                se.getSubject(),
-                se.getQueueGroup(),
-                se.getMetadata()
-        ));
-    }
-
```
scottf, line 90: Moved to the top of the file

src/main/java/io/nats/service/Service.java
```diff
@@ -63,32 +63,26 @@ public class Service {
         dInternals = new ArrayList<>();
         startStopLock = new ReentrantLock();
 
+        // build responses first. info needs to be available when adding service endpoints.
+        pingResponse = new PingResponse(id, b.name, b.version, b.metadata);
+        infoResponse = new InfoResponse(id, b.name, b.version, b.metadata, b.description);
```
scottf, line 68: build responses first. info needs to be available when adding service endpoints.

src/main/java/io/nats/service/Service.java
```diff
@@ -63,32 +63,26 @@ public class Service {
...
             dInternals.add(dTemp);
         }
-        else {
-            dTemp = null;
-        }
 
         discoveryContexts = new ArrayList<>();
         addDiscoveryContexts(SRV_PING, pingResponse, b.pingDispatcher, dTemp);
-        addDynamicDiscoveryContexts(SRV_INFO, infoResponse, b.infoDispatcher, dTemp);
+        addDiscoveryContexts(SRV_INFO, infoResponse, b.infoDispatcher, dTemp);
```
scottf, line 85: can do this because of the lazy serialization

src/main/java/io/nats/service/Service.java
```diff
@@ -97,9 +91,12 @@ public class Service {
      * @param se the service endpoint to be added
      */
     public void addServiceEndpoint(ServiceEndpoint se) {
-        Dispatcher dTemp = null == dInternals || dInternals.isEmpty() ? null : dInternals.get(0);
-        EndpointContext ctx = null;
+        // do this first so it's available on start
+        infoResponse.addServiceEndpoint(se);
```
scottf, line 95: do this first so it's available on start

src/main/java/io/nats/service/Service.java
```diff
@@ -109,19 +106,17 @@ public void addServiceEndpoint(ServiceEndpoint se) {
...
         try {
             if (runningIndicator != null) {
                 ctx.start();
             }
-        } finally {
-            startStopLock.unlock();
         }
-
-        if (null != infoResponse) {
-            infoResponse.addServiceEndpoint(se);
```
scottf, line 122: infoResponse will never be null since it's initialized first in the constructor.

src/main/java/io/nats/service/Service.java
```diff
@@ -138,19 +133,6 @@ private void addDiscoveryContexts(String discoveryName, Dispatcher dUser, Dispat
...
-     * @param discoveryName the name of the discovery
-     * @param dUser the user dispatcher
-     * @param dInternal the internal dispatcher
-     * @param handler the service message handler
-     */
-    private void addDynamicDiscoveryContexts(String discoveryName, ServiceResponse sr, Dispatcher dUser, Dispatcher dInternal) {
-        ServiceMessageHandler handler = smsg -> smsg.respond(conn, sr.serialize());
-        addDiscoveryContexts(discoveryName, dUser, dInternal, handler);
-    }
-
```
scottf, line 153: removed since lazy serialization

src/main/java/io/nats/service/ServiceResponse.java
```diff
@@ -59,6 +63,16 @@ protected ServiceResponse(String type, JsonValue jv) {
...
+        serialized = new AtomicReference<>();
+    }
+
+    @Override
+    public byte[] serialize() {
+        // lazy since endpoints can be added after creation
+        if (serialized.get() == null) {
+            serialized.set(toJson().getBytes(StandardCharsets.UTF_8));
+        }
+        return serialized.get();
```
scottf, line 75: This method is part of the JsonSerializable interface, which has a default, always execute implementation. This is why I originally applied the serializer in the add discovery, so it didn't need to be executed every time. But since we can add endpoints, the serialization needs to be updated when endpoints are added.

src/test/java/io/nats/service/ServiceTests.java
```diff
@@ -1107,7 +1171,7 @@ public void testServiceResponsesConstruction() {
         endMeta.put("foo", "bar");
         Endpoint ep = new Endpoint("endfoo", endMeta);
         ServiceEndpoint se = new ServiceEndpoint(ep, m -> {}, null);
-        InfoResponse ir1 = new InfoResponse("id", "name", "0.0.0", metadata, "desc", Collections.singletonList(se));
+        InfoResponse ir1 = new InfoResponse("id", "name", "0.0.0", metadata, "desc").addServiceEndpoint(se);
```
scottf, line 1174: reflects move the endpoints out of the constructor.

aricart (approved):
Lgtm

PR #1277, opened 2025-02-18
Service user endpoints can be run without queue groups.

scottf:
Bug fix: internal endpoints should not be run in a queue group.


rexf (approved):
Looks great!

PR #1288, opened 2025-03-15
Replace net.i2p.crypto with BouncyCastle

scottf:
net.i2p.crypto has a CVE https://nvd.nist.gov/vuln/detail/CVE-2020-36843 and is not being maintained. BouncyCastle is an industry standard well maintained library.

### DIFF
--- a/src/main/java/io/nats/client/NKey.java
+++ b/src/main/java/io/nats/client/NKey.java
@@ -13,13 +13,9 @@
 
 package io.nats.client;
 
-import net.i2p.crypto.eddsa.EdDSAEngine;
-import net.i2p.crypto.eddsa.EdDSAPrivateKey;
-import net.i2p.crypto.eddsa.EdDSAPublicKey;
-import net.i2p.crypto.eddsa.spec.EdDSANamedCurveSpec;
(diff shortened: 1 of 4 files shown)


src/main/java/io/nats/client/NKey.java
```diff
@@ -208,17 +203,16 @@ static int crc16(byte[] bytes) {
         return crc;
     }
 
-
-    private static boolean checkValidPublicPrefixByte(int prefix) {
+    private static boolean notValidPublicPrefixByte(int prefix) {
```
scottf, line 206: Intellij pointed out that this was always negated upon calling.

src/main/java/io/nats/client/NKey.java
```diff
@@ -281,8 +275,7 @@ static char[] encodeSeed(Type type, byte[] src) throws IOException {
 
     static byte[] decode(char[] src) {
         byte[] raw = base32Decode(src);
-
-        if (raw == null || raw.length < 4) {
+        if (raw.length < 4) {
```
scottf, line 278: base32Decode never returns null

omarsmak:
Thanks @scottf . Are we doing a release soon that contains the fix for the CVE that has reported?

PR #1302, opened 2025-04-10
ConsumeOptions creation from json should use default, not minimum

scottf:
This is a recently added feature where you can create ConsumeOptions from json. The value for the expires in field was not properly set to the default value when not supplied in the json.


MauriceVanVeen (approved):
LGTM

PR #1305, opened 2025-04-17
Update account AccountTier with reserved memory and reserved storage

scottf:


src/main/java/io/nats/client/api/AccountTier.java
```diff
@@ -24,10 +24,10 @@
...
 public class AccountTier {
 
-    private final int memory;
-    private final int storage;
-    private final int reservedMemory;
-    private final int reservedStorage;
+    private final long memory;
+    private final long storage;
+    private final long reservedMemory;
+    private final long reservedStorage;
```
scottf, line 30: The reserveds of type int have not been released so this is not breaking anything.

arondi (approved):
LGTM

PR #1310, opened 2025-04-30
KV Limit Marker

scottf:
Accept a per message ttl when creating a kv. Implements [ADR-43](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-43.md)

```
public long create(String key, byte[] value, MessageTtl messageTtl) throws IOException, JetStreamApiException
```

### DIFF
--- a/src/main/java/io/nats/client/api/StreamConfiguration.java
+++ b/src/main/java/io/nats/client/api/StreamConfiguration.java
@@ -1081,12 +1081,26 @@ public class StreamConfiguration implements JsonSerializable {
 
         /**
          * The time delete marker TTL duration. Server accepts 1 second or more.
-         * CLIENT DOES NOT VALIDATE
          * @param subjectDeleteMarkerTtl the TTL duration
          * @return The Builder
          */
         public Builder subjectDeleteMarkerTtl(Duration subjectDeleteMarkerTtl) {
-            this.subjectDeleteMarkerTtl = subjectDeleteMarkerTtl;
+            this.subjectDeleteMarkerTtl = validateDurationNotRequiredGtOrEqSeconds(1, subjectDeleteMarkerTtl, Duration.ZERO);
+            return this;
+        }
+
+        /**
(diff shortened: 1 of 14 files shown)


src/main/java/io/nats/client/api/StreamConfiguration.java
```diff
@@ -1081,12 +1081,26 @@ public Builder allowMessageTtl(boolean allowMessageTtl) {
 
         /**
          * The time delete marker TTL duration. Server accepts 1 second or more.
-         * CLIENT DOES NOT VALIDATE
```
scottf, line 1084: Changed my mind to now validate.

src/main/java/io/nats/client/api/StreamConfiguration.java
```diff
@@ -1081,12 +1081,26 @@ public Builder allowMessageTtl(boolean allowMessageTtl) {
...
+            this.subjectDeleteMarkerTtl = validateDurationNotRequiredGtOrEqSeconds(1, subjectDeleteMarkerTtl, Duration.ZERO);
+            return this;
+        }
+
+        /**
+         * The time delete marker TTL duration in milliseconds. Server accepts 1 second or more.
+         * @param subjectDeleteMarkerTtlMillis the TTL duration
+         * @return The Builder
+         */
+        public Builder subjectDeleteMarkerTtl(long subjectDeleteMarkerTtlMillis) {
```
scottf, line 1097: didn't originally have the ms api complement to the duration api

src/main/java/io/nats/client/api/StreamConfiguration.java
```diff
@@ -557,7 +557,7 @@ public static class Builder {
         private Map<String, String> metadata;
         private long firstSequence = 1;
         private boolean allowMessageTtl = false;
-        private Duration subjectDeleteMarkerTtl = Duration.ZERO;
+        private Duration subjectDeleteMarkerTtl;
```
scottf, line 560: null is better and the getter already noted could be null

PR #1314, opened 2025-05-13
Ability to supply a name for an ordered consumer.

scottf:
* Original Ordered Push Consumers can supply a name.
* Key Value and Object Store watchers can supply a name
* Simplified ordered consumers can supply a prefix in OrderedConsumerConfiguration


MauriceVanVeen (approved):
LGTM

Issue #1320, opened 2025-06-03
NATS Client "Authorization Violation" after server restart


ajax-surovskyi-y:
### Observed behavior

During reconnect, UNSUB messages  may be enqueued in the reconnectOutgoing queue if they are issued after the client’s connection status check but before the transition into reconnect mode. Under a race condition, UNSUB commands can be batched and transmitted prior to the mandatory CONNECT message. This causes the server to interpret the messages as coming from an unauthorized, not-yet-authenticated client, leading to Authorization Violation errors.

### Expected behavior

Reconnect is successful

### Server and client version

- NATS Server version: 2.10.24
- NATS Client library/version: 2.21.0
- Authentication method: Token Authentication
- Deployment: Multiple NATS server instances

### Steps to reproduce

The analysis identified that if context switching occurs, threads performing unsubscribe operations might pass a connection check that becomes outdated once the reconnect process starts. As a result, UNSUB messages are sent with the wrong ordering relative to CONNECT, and the server responds with errors. The logs show that multiple errors can occur if several messages are sent in this state.

| T1 (NatsConnection.unsubscribe)                                                                                                                                 | T2 NatsConnection.closeSocket()                                                                                                                              |
|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|
| ![Image](https://github.com/user-attachments/assets/e30900cf-0c69-423a-91cd-c2c5283bd086)                                                                       | ![Image](https://github.com/user-attachments/assets/1c79a9e1-0b7d-4910-8538-60d7684d7fdf)                                                                    |
| ![Image](https://github.com/user-attachments/assets/2ed78a0d-3d40-4f0d-85c0-795fa7754800) The status is being checked, which will change in the next step in T2 |                                                                                                                                                              |
|                                                                                                                                                                 | ![Image](https://github.com/user-attachments/assets/3d34a637-4aac-4547-8f4c-277277d2bf7f)                                                                    |
|                                                                                                                                                                 | ![Image](https://github.com/user-attachments/assets/c46393b9-f7b4-4da2-b051-b48f35388059)                                                                    |
|                                                                                                                                                                 | ![Image](https://github.com/user-attachments/assets/b09cb662-3d85-4c2f-8b39-ee7bc7b8eab7) writer.setReconnectMode() toggles the NatsConnectionWriter queue |
[...]

PR #1321, opened 2025-06-03
Fix race condition during reconnect sends UNSUB messages

ajax-surovskyi-y:
This is a PR for https://github.com/nats-io/nats.java/issues/1320

Race condition during reconnect sends unsent UNSUB messages before CONNECT, triggering Authorization Violation errors

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -638,7 +638,7 @@ class NatsConnection implements Connection {
                 }
 
                 this.currentServer = cur;
-                this.serverAuthErrors.remove(resolved); // reset on successful connection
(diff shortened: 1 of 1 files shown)


scottf:
@ajax-surovskyi-y I'm inclined to accept this portion:
```
this.serverAuthErrors.clear(); // reset on successful connection
```
but not the other change.

I think what we need to do is figure out how to know that we are trying to transmit between disconnect and connect. Maybe a second check?

What would be useful is some repeatable test to be able to test this. If we have to use some sort of mock or simulation, that would be extremely helpful.

ajax-surovskyi-y:
@scottf Thanks! I've updated the PR to include only the 

> this.serverAuthErrors.clear();

 change.

I'll also try to come up with a repeatable test to reproduce the issue reliably.

Will follow up once I have something concrete.

PR #1323, opened 2025-06-04
Nats-Marker-Reason must be mapped to a Key Value Operation

scottf:
- "Remove" map to DEL
- "MaxAge" and "Purge" maps to PURGE


MauriceVanVeen (approved):
LGTM

PR #1328, opened 2025-06-10
Add test for auth violations during reconnect

ajax-surovskyi-y:
Introduce a new test case to reproduce issue #1320, focusing on handling authorization violations during reconnections. It also makes the `closeSocketLock` in `NatsConnection` package-private to facilitate testing.


ajax-surovskyi-y:
> @ajax-surovskyi-y What is the answer to the question I posed about you re-subscribing?
> 
> > In your original code, are you trying to re-subscribe manually as this example does? The client is supposed to re-subscribe for you.

Our original code contains logic that can trigger a mass unsubscribe. Under realistic conditions this could remove on the order of 10,000 subscriptions at once. If this coincides with reconnect timing, it can significantly increase the likelihood of the issue occurring.

PR #1335, opened 2025-06-24
Replace Timer with scheduled tasks

scottf:
* Reduces thread usage when many pull consumers
* Scheduler uses nano time making it better immune to time shifts


MauriceVanVeen (approved):
LGTM

PR #1336, opened 2025-06-24
[BUG] Fix UNSUBs after disconnect can cause auth violations

### LINKED ISSUE #1320: NATS Client "Authorization Violation" after server restart

scottf:
Fixes #1320

An Authorization Violation could occur if an UNSUB was sent between a disconnect and reconnect, because the UNSUB got queued and sent before the CONNECT protocol finished. This PR prevents that from happening and also addresses not needing to send unsubs after a disconnect.

1. Puts UNSUBS in the normal (not internal/reconnect) queue. This by itself is enough to solve the bug, since the normal queue is never sent while not connected.

2. On a disconnect, PINGS and PONGS were already filtered out of the normal outgoing queue because they are invalid anyway. 
    * This PR filters out UNSUB messages, since they are unnecessary traffic; subs are closed on the server anyway because of the disconnect.
    * This PR filters out SUB messages. By the time the request is made to send the actual SUB message to the server, the sub has been registered internally. This means that if there is a disconnect, the client is going to resub, so not filtering would actually create a double sub. This is a rare edge case. 

---
If you want to verify, run the test it passes. Then go into NatsConnection and change line 1062 to
```
queueInternalOutgoing(new ProtocolMessage(bab));
```
and then run the test, it will fail.

---
Thanks @ajax-surovskyi-y


src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1527,7 +1527,7 @@ public Duration RTT() throws IOException {
         pongQueue.add(pongFuture);
         try {
             long time = NatsSystemClock.nanoTime();
-            writer.queueInternalMessage(new ProtocolMessage(OP_PING_BYTES));
+            writer.queueInternalMessage(new ProtocolMessage(PONG_PROTO));
```
scottf: Use the premade PONG_PROTO for faster construction

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -105,10 +106,7 @@ Future<Boolean> stop() {
             try {
                 this.outgoing.pause();
                 this.reconnectOutgoing.pause();
-                // Clear old ping/pong requests
-                this.outgoing.filter((msg) ->
-                    msg.isProtocol() &&
-                        (msg.getProtocolBab().equals(OP_PING_BYTES) || msg.getProtocolBab().equals(OP_PONG_BYTES)));
+                this.outgoing.filter(NatsMessage::isProtocolFilterOnStop);
```
scottf, line 109: Leverage the new flag for faster filtering instead of doing byte compares on proto bytes

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -151,7 +149,7 @@ void sendMessageBatch(NatsMessage msg, DataPort dataPort, StatisticsCollector st
                 sendBuffer[sendPosition++] = CR;
                 sendBuffer[sendPosition++] = LF;
 
-                if (!msg.isProtocol()) { // because a protocol message does not have headers
+                if (!msg.isProtocol()) { // because a protocol message does not have headers or data
```
scottf, line 152: This clarification was important.

src/main/java/io/nats/client/impl/ProtocolMessage.java
```diff
@@ -19,30 +19,39 @@
 // Protocol message is a special version of a NatsPublishableMessage extends NatsMessage
 // ----------------------------------------------------------------------------------------------------
 class ProtocolMessage extends NatsPublishableMessage {
+    final boolean filterOnStop;
```
scottf, line 22: must construct with this information now

src/main/java/io/nats/client/impl/ProtocolMessage.java
```diff
@@ -19,30 +19,39 @@
...
         protocolBab = babProtocol;
         sizeInBytes = controlLineLength = protocolBab.length() + 2; // CRLF, protocol doesn't have data
+        this.filterOnStop = filterOnStop;
     }
 
     ProtocolMessage(byte[] protocol) {
         super(false);
         protocolBab = new ByteArrayBuilder(protocol);
         sizeInBytes = controlLineLength = protocolBab.length() + 2; // CRLF, protocol doesn't have data
+        this.filterOnStop = true;
```
scottf, line 35: I looked through every use this constructor. 99% were test construction. The only production usage were these, the pre-made messages which speed up new ProtocolMessage creation by sharing the existing protocol buffer which never changes for these fixed size protocols. These are used to construct new ProtocolMessage, which is required since they have some state related to their queueing - you can only share / reuse the data, not the message itself.
```
private static final ProtocolMessage PING_PROTO = new ProtocolMessage(OP_PING_BYTES);
private static final ProtocolMessage PONG_PROTO = new ProtocolMessage(OP_PONG_BYTES);
```

MauriceVanVeen (approved):

src/main/java/io/nats/client/impl/ProtocolMessage.java
ajax-surovskyi-y: One concern I wanted to raise is the potential memory impact of adding additional state to each NatsMessage.
In high-throughput systems (we’re handling tens of thousands of messages per second), even a small increase in object size can become noticeable.

ajax-surovskyi-y: On second thought, I may have overestimated the impact — this change applies to protocol messages rather than the actual PUB data, so the volume is relatively low.
That likely means the memory overhead is negligible.

PR #1339, opened 2025-06-26
[BUG] MessageConsumer.isFinished() not set properly in certain condit…

### LINKED ISSUE #1337: MessageConsumer.isFinished() after .stop() will never report true on empty stream or after seeing messages with no future messages in stream
### Observed behavior

When testing some of the graceful closure of MessageConsumer I came across MessageConsumerExample.
This tells me to call consumerContext.consume( handler ), and then when finished I can call consumer.stop() and wait for consumer.isFinished() to return true.

I've found instead that the message consumer will never return true unless there are more messages in the stream to consume, and it has already seen messages.
My scenario is that I'm shutting down one consumer last in a series of applications, and since its already consumed everything it always times out.
Based on a brief examination of the code it looks like the `AtomicBoolean finished` field is only set to true in the consumption of a message, rather than the time outs of all requests to the server.
This also occurs if no messages are received before consumer.stop() is called - any subsequent messages will be received but I will not see the consumer close.

The only time I can consistently successfully stop() is when the consumer sees some events before stop() and there are more messages in the stream.

### Expected behavior

Calling consumer.stop() on a stream which never receives messages should have isFinished() return true quickly.
Calling consumer.stop() on a stream post seeing messages but with no more messages in the stream should have isFinished() return true quickly

### Server and client version

nats-server 2.10.26
nats java 2.20.5

### Host environment

Seen locally on Mac OSX and on ubuntu servers in GCP.

### Steps to reproduce

Call .consume() on empty stream, call .stop() then wait for isFinished() == true which will infinitely loop.
Call .consume() on a stream, publish exactly X messages and then wait for a countdown latch for them to be received, then call .stop() and wait for isFinished() == true which will will infinitely loop.

scottf:
…ions.

Fixes #1337

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsFetchConsumer.java
+++ b/src/main/java/io/nats/client/impl/NatsFetchConsumer.java
@@ -67,8 +67,7 @@ class NatsFetchConsumer extends NatsMessageConsumerBase implements FetchConsumer
 
     @Override
     public void heartbeatError() {
-        stopped.set(true);
-        finished.set(true);
+        finish(true);
     }
 
     @Override
@@ -86,8 +85,7 @@ class NatsFetchConsumer extends NatsMessageConsumerBase implements FetchConsumer
                 if (m == null) {
                     // if there are no messages in the internal cache AND there are no more pending,
                     // they all have been read and we can go ahead and finish
-                    finished.set(true);
-                    lenientClose();
+                    finish(true);
                 }
                 return m;
             }
@@ -105,8 +103,7 @@ class NatsFetchConsumer extends NatsMessageConsumerBase implements FetchConsumer
                 Message m = sub._nextUnmanagedNoWait(pullSubject);
                 if (m == null) {
                     // no message and no time left, go ahead and finish
-                    finished.set(true);
-                    lenientClose();
+                    finish(true);
                 }
                 return m;
             }
--- a/src/main/java/io/nats/client/impl/NatsMessageConsumer.java
+++ b/src/main/java/io/nats/client/impl/NatsMessageConsumer.java
@@ -53,9 +53,14 @@ class NatsMessageConsumer extends NatsMessageConsumerBase implements PullManager
     public void heartbeatError() {
         try {
             // just close the current sub and make another one.
-            // this could go on endlessly
-            lenientClose();
-            doSub();
+            // this could go on endlessly - unless the user had called stop
+            if (stopped.get()) {
+                finish(true);
+            }
+            else {
+                lenientClose();
+                doSub();
+            }
         }
         catch (JetStreamApiException | IOException e) {
             pmm.resetTracking();
@@ -67,7 +72,7 @@ class NatsMessageConsumer extends NatsMessageConsumerBase implements PullManager
         MessageHandler mh = userMessageHandler == null ? null : msg -> {
             userMessageHandler.onMessage(msg);
             if (stopped.get() && pmm.noMorePending()) {
-                finished.set(true);
+                finish(false);
             }
         };
         super.initSub(subscriptionMaker.subscribe(mh, userDispatcher, pmm, null));
@@ -78,7 +83,12 @@ class NatsMessageConsumer extends NatsMessageConsumerBase implements PullManager
 
     @Override
     public void pendingUpdated() {
-        if (!stopped.get() && (pmm.pendingMessages <= thresholdMessages || (pmm.trackingBytes && pmm.pendingBytes <= thresholdBytes)))
+        if (stopped.get()) {
+            if (pmm.noMorePending()) {
+                finish(false);
+            }
+        }
+        else if (pmm.pendingMessages <= thresholdMessages || (pmm.trackingBytes && pmm.pendingBytes <= thresholdBytes))
         {
             repull();
         }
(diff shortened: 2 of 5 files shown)


src/main/java/io/nats/client/impl/NatsFetchConsumer.java
```diff
@@ -67,8 +67,7 @@ public void pendingUpdated() {}
 
     @Override
     public void heartbeatError() {
-        stopped.set(true);
-        finished.set(true);
+        finishAndClose();
```
scottf, line 70: a heartbeat error on a fetch is terminal

src/main/java/io/nats/client/impl/NatsFetchConsumer.java
```diff
@@ -86,8 +85,7 @@ public Message nextMessage() throws InterruptedException, JetStreamStatusChecked
                 if (m == null) {
                     // if there are no messages in the internal cache AND there are no more pending,
                     // they all have been read and we can go ahead and finish
-                    finished.set(true);
-                    lenientClose();
+                    finishAndClose();
```
scottf, line 88: no more messages on a fetch is terminal

src/main/java/io/nats/client/impl/NatsFetchConsumer.java
```diff
@@ -105,17 +103,15 @@ public Message nextMessage() throws InterruptedException, JetStreamStatusChecked
                 Message m = sub._nextUnmanagedNoWait(pullSubject);
                 if (m == null) {
                     // no message and no time left, go ahead and finish
-                    finished.set(true);
-                    lenientClose();
+                    finishAndClose();
```
scottf, line 106: no more messages on a fetch is terminal

src/main/java/io/nats/client/impl/NatsFetchConsumer.java
```diff
@@ -105,17 +103,15 @@ public Message nextMessage() throws InterruptedException, JetStreamStatusChecked
...
                 }
                 return m;
             }
 
             Message m = sub._nextUnmanaged(timeLeftNanos, pullSubject);
             if (m == null && isNoWaitNoExpires) {
                 // no message and no wait, go ahead and finish
-                finished.set(true);
-                lenientClose();
+                finishAndClose();
```
scottf, line 114: no more messages on a fetch is terminal

src/main/java/io/nats/client/impl/NatsMessageConsumer.java
```diff
@@ -67,7 +72,7 @@ void doSub() throws JetStreamApiException, IOException {
         MessageHandler mh = userMessageHandler == null ? null : msg -> {
             userMessageHandler.onMessage(msg);
             if (stopped.get() && pmm.noMorePending()) {
-                finished.set(true);
+                finishAndClose();
```
scottf, line 75: We have no more pending and the user has called stop, so we are finished with this consume

src/main/java/io/nats/client/impl/NatsMessageConsumer.java
```diff
@@ -78,7 +83,12 @@ void doSub() throws JetStreamApiException, IOException {
 
     @Override
     public void pendingUpdated() {
-        if (!stopped.get() && (pmm.pendingMessages <= thresholdMessages || (pmm.trackingBytes && pmm.pendingBytes <= thresholdBytes)))
+        if (stopped.get()) {
```
scottf, line 86: We have no more pending and the user has called stop, so we are finished with this consume

src/test/java/io/nats/client/impl/SimplificationTests.java
```diff
@@ -1303,20 +1303,20 @@ public void testOverflowFetch() throws Exception {
...
-                .expiresIn(2000)
+                .expiresIn(3000)
                 .group(group)
                 .minAckPending(5)
                 .build();
 
             FetchConsumeOptions fcoOverB = FetchConsumeOptions.builder()
                 .maxMessages(5)
-                .expiresIn(2000)
+                .expiresIn(3000)
```
scottf, line 1319: this test was flapping, so added a little extra time

src/main/java/io/nats/client/impl/NatsMessageConsumer.java
```diff
@@ -53,9 +53,14 @@ class NatsMessageConsumer extends NatsMessageConsumerBase implements PullManager
...
-            lenientClose();
-            doSub();
+            // this could go on endlessly - unless the user had called stop
+            if (stopped.get()) {
+                finishAndClose();
+            }
+            else {
+                lenientClose();
+                doSub();
+            }
```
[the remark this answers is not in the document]
scottf: The body of the else is the original code. The if before the else is new

src/main/java/io/nats/client/impl/NatsMessageConsumer.java
```diff
@@ -53,9 +53,14 @@ class NatsMessageConsumer extends NatsMessageConsumerBase implements PullManager
...
         try {
             // just close the current sub and make another one.
-            // this could go on endlessly
-            lenientClose();
-            doSub();
+            // this could go on endlessly - unless the user had called stop
+            if (stopped.get()) {
+                finishAndClose();
+            }
+            else {
```
scottf, line 60: If a heartbeat error occurs, we would usually just try again. But if the user had called stop, they've told us that we are finished.

PR #1345, opened 2025-07-01
Debug unit test hanging 01-001

scottf:


src/main/java/io/nats/client/impl/DispatcherFactory.java
```diff
@@ -16,10 +16,10 @@
...
 
 /**
- * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
- * WARNING: THIS CLASS IS PUBLIC BUT ITS API IS NOT GUARANTEED
- * TO BE BACKWARD COMPATIBLE AS IT IS PURELY AN INTERNAL CLASS
- * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! *
+ * WARNING: THIS CLASS IS PUBLIC BUT ITS API IS NOT GUARANTEED TO *
+ * BE BACKWARD COMPATIBLE AS IT IS INTENDED AS AN INTERNAL CLASS  *
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! *
```
scottf, line 22: Just updating this comment to match the other one I added. No production code.

src/main/java/io/nats/client/impl/MessageManager.java
```diff
@@ -42,11 +44,10 @@ public enum ManageResult {MESSAGE, STATUS_HANDLED, STATUS_TERMINUS, STATUS_ERROR
...
     // heartbeat stuff
-    protected boolean hb;
-    protected long idleHeartbeatSetting;
-    protected long alarmPeriodSettingNanos;
-    protected ScheduledTask heartbeatTask;
-    protected final AtomicLong currentAlarmPeriodNanos;
+    protected final AtomicBoolean hb;
+    protected final AtomicLong idleHeartbeatSettingMillis;
+    protected final AtomicLong alarmPeriodSettingNanos;
+    protected final AtomicReference<ScheduledTask> heartbeatTask;
```
scottf, line 50: Just made everything atomic since these value could change from different threads.

src/main/java/io/nats/client/impl/MessageManager.java
```diff
@@ -137,26 +139,20 @@ protected void updateLastMessageReceived() {
...
-                if (currentAlarmPeriodNanos.get() == alarmPeriodSettingNanos) {
-                    updateLastMessageReceived();
-                    return;
-                }
-
-                // Replace timer since settings have changed
-                shutdownHeartbeatTimer();
+            ScheduledTask hbTask = heartbeatTask.get();
+            if (hbTask != null) {
+                hbTask.shutdown();
```
scottf, line 144: The gist of the PR is that I think that the scheduled heartbeat task was still running causing essentially an infinite loop. Some of the original code here was to reuse timer tasks because they were much heavier (had their own thread). So I had code to reuse them. But this is so much more light weight, just adding a task to the scheduler, so I simplified the code to just close the open task (shutdownHeartbeatTimer does that) and make a new one.

src/main/java/io/nats/client/impl/MessageManager.java
```diff
@@ -137,26 +139,20 @@ protected void updateLastMessageReceived() {
...
 
             // replacement or new comes here
-            this.currentAlarmPeriodNanos.set(alarmPeriodSettingNanos);
-            heartbeatTask = new ScheduledTask(conn.getScheduledExecutor(), alarmPeriodSettingNanos, TimeUnit.NANOSECONDS,
+            heartbeatTask.set(new ScheduledTask(conn.getScheduledExecutor(), alarmPeriodSettingNanos.get(), TimeUnit.NANOSECONDS,
                 () -> {
                     long sinceLast = NatsSystemClock.nanoTime() - lastMsgReceivedNanoTime.get();
-                    if (sinceLast > currentAlarmPeriodNanos.get()) {
+                    if (sinceLast > alarmPeriodSettingNanos.get()) {
+                        shutdownHeartbeatTimer(); // a new one will get started when needed.
```
scottf, line 152: The second part here is that when I get a heartbeat alarm, I make sure I shutdown the timer/task. I end up calling handleHeartbeatError (see next line of code) which is raises the error to the code that made the subscription to begin with (simplified consumer for instance). It's their problem to know what to do, for instance simplified will just try to make another sub.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -601,7 +601,7 @@ void tryToConnect(NatsUri cur, NatsUri resolved, long now) {
 
                 if (pingMillis > 0) {
                     pingTask = new ScheduledTask(scheduledExecutor, pingMillis, () -> {
-                        if (isConnected()) {
+                        if (isConnected() && !isClosing()) {
```
scottf, line 604: When figuring this out I considered to see if I had to somehow make the runnable part of the task aware that it should stop, similar to to a keepGoing flag or handling an interrupt. I noticed that the tasks I care about are short lived. And then I noticed that there is no point in pinging if the connection is currently open but being closed so I changed this.

src/main/java/io/nats/client/support/ScheduledTask.java
```diff
@@ -17,50 +17,113 @@
...
+
 /**
- * This is a utility class for a task being scheduled
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! *
+ * WARNING: THIS CLASS IS PUBLIC BUT ITS API IS NOT GUARANTEED TO *
+ * BE BACKWARD COMPATIBLE AS IT IS INTENDED AS AN INTERNAL CLASS  *
+ * !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! *
  */
 public class ScheduledTask implements Runnable {
+    private static final AtomicLong ID_GENERATOR = new AtomicLong();
```
scottf, line 31: I added methods to support testing and make this a little bit more robust.

MauriceVanVeen (approved):
LGTM

Issue #1346, opened 2025-07-02
Fastfallback Algorithm for faster connections with dual stack ip configuration


jitinsharma:
### Proposed change

**Introduction**
We are currently using nats java client in our android application. Over time we got user queries, where they were unable to connect to nats server and after some debugging we realised, some of our users connecting ipv6 addresses were facing this problem.
Issues with ipv6 with dual ip setup are well documented and are expected not to occur, but due to faulty hardware it does happen.

Here's an RFC which is implemented by clients such as okhttp, chrome, Firefox etc
https://datatracker.ietf.org/doc/html/rfc6555

### Use case

We are currently using nats as a web socket client for our apps. We have enabled `noResolveHostnames` property for it.
When this property is true, `SocketDataPort` uses following line of code to make tcp connection.
```java
socket.connect(new InetSocketAddress(host, port), (int) timeout);
```
where in `new InetSocketAddress` returns first ip address from list of addresses dns has resolved.
In cases, where ipv6 connectivity is error prone, this would lead to persistent errors.

The proposal is to create another `DataPort` class, which can make speculative connection to each ip address with a delay and connect to fastest one. This can be enabled using another boolean property in `Options`.
To give an example, here's how okhttp does it.
https://github.com/square/okhttp/blob/master/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/connection/FastFallbackExchangeFinder.kt

Exact steps
1. Get all IP addresses from InetAddress.
2. Create an executor with number of threads equal to number of ip addresses.
3. Using the order of ip addresses received, trigger socket connect to each ip address with a delay of 250ms.
4. If any of these succeed, cancel others and return successful one.

### Contribution

Since NatsClient allows overriding of `DataPort` class using Options, we have already built this algorithm and tested it in production where it is working fine and has led to reduction in timeout errors. 
I'll be happy to contribute the same to the repo, if this proposal looks fine.

PR #1351, opened 2025-07-07
Implement Fast fallback algorithm in SocketDataPort

jitinsharma:
#1346

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -529,6 +529,12 @@ public class Options {
      */
     public static final String PROP_READ_LISTENER_CLASS = "read.listener.class";
 
+    /**
+     * Property used to enable fast fallback algorithm for socket connection.
+     * {@link Builder#enableFastFallback() enableFastFallback}.
+     */
+    public static final String PROP_FAST_FALLBACK = PFX + "fast.fallback";
+
     // ----------------------------------------------------------------------------------------------------
     // PROTOCOL CONNECT OPTION CONSTANTS
     // ----------------------------------------------------------------------------------------------------
@@ -950,6 +956,7 @@ public class Options {
             booleanProperty(props, PROP_USE_TIMEOUT_EXCEPTION, b -> this.useTimeoutException = b);
             booleanProperty(props, PROP_USE_DISPATCHER_WITH_EXECUTOR, b -> this.useDispatcherWithExecutor = b);
             booleanProperty(props, PROP_FORCE_FLUSH_ON_REQUEST, b -> this.forceFlushOnRequest = b);
+            booleanProperty(props, PROP_FAST_FALLBACK, b -> this.enableFastFallback = b);
 
             classnameProperty(props, PROP_SERVERS_POOL_IMPLEMENTATION_CLASS, o -> this.serverPool = (ServerPool) o);
             classnameProperty(props, PROP_DISPATCHER_FACTORY_CLASS, o -> this.dispatcherFactory = (DispatcherFactory) o);
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -678,6 +678,7 @@ public class Options {
 
     private final List<java.util.function.Consumer<HttpRequest>> httpRequestInterceptors;
     private final Proxy proxy;
+    private final boolean enableFastFallback;
 
     static class DefaultThreadFactory implements ThreadFactory {
         String name;
@@ -802,6 +803,7 @@ public class Options {
         private char[] truststorePassword;
         private String tlsAlgorithm = DEFAULT_TLS_ALGORITHM;
         private String credentialPath;
+        private boolean enableFastFallback = false;
 
         /**
          * Constructs a new Builder with the default values.
@@ -1763,6 +1765,15 @@ public class Options {
         }
 
         /**
+         * Whether to enable Fast fallback algorithm for socket connect
+         * @return the Builder for chaining
+         */
+        public Builder enableFastFallback() {
+            this.enableFastFallback = true;
+            return this;
+        }
+
+        /**
          * Build an Options object from this Builder.
          *
          * <p>If the Options builder was not provided with a server, a default one will be included
@@ -1996,6 +2007,7 @@ public class Options {
 
             this.serverPool = o.serverPool;
             this.dispatcherFactory = o.dispatcherFactory;
+            this.enableFastFallback = o.enableFastFallback;
         }
     }
 
@@ -2064,6 +2076,7 @@ public class Options {
 
         this.serverPool = b.serverPool;
         this.dispatcherFactory = b.dispatcherFactory;
+        this.enableFastFallback = b.enableFastFallback;
     }
 
     // ----------------------------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/main/java/io/nats/client/impl/SocketDataPort.java
```diff
@@ -187,4 +183,70 @@ protected static boolean isWebsocketScheme(String scheme) {
...
+     * Implements the "Happy Eyeballs" algorithm as described in RFC 6555,
+     * which attempts to connect to multiple IP addresses in parallel to reduce
+     * connection setup delays.
+     */
+    private Socket connectToFastestIp(Options options, String hostname, int port,
+                                      int timeoutMillis) throws IOException {
+        // Get all IP addresses for the hostname
+        List<InetAddress> ips = Arrays.asList(InetAddress.getAllByName(hostname));
+
+        ExecutorService executor = Executors.newFixedThreadPool(ips.size());
```
scottf: Developers can provide their own executors in options and these must be used.

src/main/java/io/nats/client/Options.java
```diff
@@ -678,6 +678,7 @@ public class Options {
 
     private final List<java.util.function.Consumer<HttpRequest>> httpRequestInterceptors;
     private final Proxy proxy;
+    private final boolean enableFastFallback;
```
scottf, line 694: This property needs an environment/property variable. The goal is that every single option can be set via a property. Look at PROP_TLS_FIRST for an example

src/main/java/io/nats/client/impl/SocketDataPort.java
```diff
@@ -187,4 +183,70 @@ protected static boolean isWebsocketScheme(String scheme) {
...
+
+    /**
+     * Implements the "Happy Eyeballs" algorithm as described in RFC 6555,
+     * which attempts to connect to multiple IP addresses in parallel to reduce
+     * connection setup delays.
+     */
+    private Socket connectToFastestIp(Options options, String hostname, int port,
+                                      int timeoutMillis) throws IOException {
+        // Get all IP addresses for the hostname
+        List<InetAddress> ips = Arrays.asList(InetAddress.getAllByName(hostname));
```
scottf, line 195: This concerns me. I thought that ip addresses were already resolved by the time they make it here.

src/main/java/io/nats/client/impl/SocketDataPort.java
```diff
@@ -187,4 +183,70 @@ protected static boolean isWebsocketScheme(String scheme) {
...
+
+        ExecutorService executor = Executors.newFixedThreadPool(ips.size());
+        long CONNECT_DELAY_MILLIS = 250;
```
scottf, line 198: Why was 250 millis chosen? Should this be configurable?

src/main/java/io/nats/client/impl/SocketDataPort.java
```diff
@@ -187,4 +183,70 @@ protected static boolean isWebsocketScheme(String scheme) {
...
+    private Socket createSocket(Options options) throws SocketException {
+        Socket socket;
+        if (options.getProxy() != null) {
+            socket = new Socket(options.getProxy());
+        } else {
+            socket = new Socket();
+        }
+        socket.setTcpNoDelay(true);
+        socket.setReceiveBufferSize(2 * 1024 * 1024);
+        socket.setSendBufferSize(2 * 1024 * 1024);
```
scottf, line 242: these settings should always apply once you've figured out the fastest ip

src/main/java/io/nats/client/impl/SocketDataPort.java
```diff
@@ -72,16 +71,13 @@ public void connect(NatsConnection conn, NatsUri nuri, long timeoutNanos) throws
...
+                socket = createSocket(options);
+                socket.connect(new InetSocketAddress(host, port), (int) timeout);
             }
-            else {
-                socket = new Socket();
-            }
-            socket.setTcpNoDelay(true);
-            socket.setReceiveBufferSize(2 * 1024 * 1024);
-            socket.setSendBufferSize(2 * 1024 * 1024);
-            socket.connect(new InetSocketAddress(host, port), (int) timeout);
```
scottf, line 84: These setting should always apply

scottf (changes requested):

src/main/java/io/nats/client/impl/SocketDataPort.java
jitinsharma: There are 2 scenarios where ips are not resolved.
- when `noResolveHostnames` is enabled.
- when websocket protocol is used.
In other cases, `NatsConnection` class already does sequential tries on all ip addresses, where this would not be needed.

src/main/java/io/nats/client/impl/SocketDataPort.java
jitinsharma: Got this from okhttp, other clients like Chrome, Firefox use 300ms. Not really needed to be configurable.

src/main/java/io/nats/client/impl/SocketDataPort.java
jitinsharma: Is it safe to apply these after `socket.connect` is called, since figuring out fastest ip requires socket connection?

src/main/java/io/nats/client/impl/SocketDataPort.java
jitinsharma: done

src/main/java/io/nats/client/Options.java
jitinsharma: done

src/main/java/io/nats/client/impl/SocketDataPort.java
jitinsharma: Also updated the code to no longer shutdown the executor since it's custom provided by developers. Tasks will be cancelled automatically.

src/main/java/io/nats/client/impl/SocketDataPort.java
scottf: Maybe we can turn off resolve when the new setting is used. Like pair them together to you resolve them. And yeah, I remember now that I already don't resolve before trying a ws/s connection and I remember this being important specifically to not resolve since part of the protocol includes the host name for matching purposes, something like that.

src/main/java/io/nats/client/impl/SocketDataPort.java
scottf: Ok then these need to be part of every connection you make, for instance TcpNoDelay is a requirement that came from the inventor of NATS. Unless there is a very good reason to not do this on Android. In which case, maybe we need to rename this to Android mode which also has no resolve.

src/main/java/io/nats/client/impl/SocketDataPort.java
scottf: Let's make this configurable, since you seem to have a reason to choose something other than 300 to match the browsers. Again, I feel like I'm make a collection of Android specific settings.

jitinsharma: Sure, but I feel mostly devs using this property won't find the need to change this as I've yet to see any other client make it configurable and the RFC doesn't define a constant interval.

src/main/java/io/nats/client/impl/SocketDataPort.java
jitinsharma: I don't see any good reason for changing this behaviour on Android, hence I've kept all these properties same as before. Just moved all of these into a function to avoid code repetition in if-else block.

src/main/java/io/nats/client/impl/SocketDataPort.java
jitinsharma: Got it, when I'll do the changes in code where resolve happens to honour the new setting.

src/main/java/io/nats/client/impl/SocketDataPort.java
scottf: Ok I see you delegate to create socket. Marking resolved.

PR #1354, opened 2025-07-15
[Fix] Simplified Ordered Consumer - Getting name early can cause NPE

scottf:
Also test app and example cleanup.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsMessageConsumerBase.java
+++ b/src/main/java/io/nats/client/impl/NatsMessageConsumerBase.java
@@ -43,6 +43,7 @@ class NatsMessageConsumerBase implements MessageConsumer {
 
     void initSub(NatsJetStreamPullSubscription sub) {
         this.sub = sub;
+        this.consumerName = sub.getConsumerName();
         pmm = (PullMessageManager)sub.manager;
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/io/nats/client/impl/NatsMessageConsumerBase.java
```diff
@@ -43,6 +43,7 @@ void setConsumerName(String consumerName) {
 
     void initSub(NatsJetStreamPullSubscription sub) {
         this.sub = sub;
+        this.consumerName = sub.getConsumerName();
```
scottf, line 46: Directly setting the consumer name.

src/main/java/io/nats/client/impl/NatsMessageConsumerBase.java
```diff
@@ -65,7 +66,7 @@ public boolean isFinished() {
      */
     @Override
     public String getConsumerName() {
-        if (consumerName == null) {
+        if (consumerName == null && cachedConsumerInfo != null) {
```
scottf, line 69: found the bug when cachedConsumerInfo was null, early in the lifecycle. Fixed the problem above but added the guard anyway.

MauriceVanVeen (approved):
LGTM

PR #1363, opened 2025-07-17
Fix core unsubscribe by subject

### LINKED ISSUE #1232: Dispatcher unsubscribe should throw if the wrong method is used
### Proposed change

The Dispatcher interface provides overloaded `unsubscribe` methods, unfortunately it is quite easy for someone to choose the incorrect method and this doesn't produce any kind of error but has the potentially dangerous side-effect of not actually closing the subscription. Please consider throwing an exception or at least logging a warning in the library when unsubscribe is used this way and the subscription is not found on the default handler.

```java
Options.Builder builder = new Options.Builder().server("nats://localhost:4222");
try (Connection conn = Nats.connect(builder.build())) {
    Dispatcher d = conn.createDispatcher();
    d.subscribe("test", msg -> {});
    ...
    d.unsubscribe("test");
}
```

### Use case

In the example above, the unsubscribe call does not throw but also it does not actually unsubscribe. In applications that manage subscriptions dynamically, this can result in a huge number of subscriptions being quietly left open on the server resulting in performance degradation and eventual OOM. The mistake is also easily missed in code review and by static analysis tools. The behavior is also difficult to validate in tests because the library does not provide a way to check the number of active subscriptions.

### Contribution

_No response_

scottf:
Fixes #1232

### Context
The dispatcher api allows subscribe 
1. with a subject and no message handler, `subscribe("foo");`, in which case it will use the message handler provided when the dispatcher was created.
2. with a subject plus and a message handler, `subscribe("foo", nonDefaultHandler);`.

### Problem
If you `unsubscribe("foo")`, the dispatcher currently only unsubscribes the subscription with the default handler, meaning if you had subscribed with a non default handler, that subscriptions did not get unsubscribed. (That was the behavior reported in #1232)

### Desired Behavior
The dispatcher should unsubscribe for all handlers, default or otherwise, for a subject.

### Considerations
You can only have one subscription for a subject with the default handler (otherwise you just get multiple copies of the same message to that handler) but you can have as many subscriptions for a subject if you provide a non default handler.

So if you do this: 
```
subscribe("foo"); 
subscribe("foo"); 
subscribe("foo", handler1); 
subscribe("foo", handler2);
```
you end up with 3 subscriptions.

### Solution
 
Unsubscribe all subscriptions for the subject regardless of the handler.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,6 @@ plugins {
 def jarVersion = "2.21.5"
 
 def buildEvent = System.getenv("BUILD_EVENT")
-def isLocal = buildEvent == null
 def isRelease = buildEvent == "release"
 def brn = System.getenv("BRANCH_REF_NAME")
 def snap = brn == null || brn.equals("") ? "-SNAPSHOT" : "." + brn + "-SNAPSHOT"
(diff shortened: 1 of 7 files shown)


MauriceVanVeen (approved):
LGTM

PR #1373, opened 2025-07-26
Feature/gradle8.14 8-in-1

magicprinc:
1) Gradle 8.14.2 (latest)

2) Google error-prone https://errorprone.info/ 
Error Prone is a static analysis tool and Java compiler plugin, open-sourced by Google, that detects common programming mistakes and bugs during compilation and can suggest or automatically apply fixes to improve code quality and prevent errors before runtime

3) Lombok https://projectlombok.org/
Project Lombok is a java library that automatically plugs into your editor and build tools, spicing up your java.
Never write another getter or equals method again, with one annotation your class has a fully featured builder, Automate your logging variables, and much more.

4) reproducible build https://reproducible-builds.org/
Reproducible builds are a set of software development practices that create an independently-verifiable path from source to binary code.

5) https://jitpack.io compatible (easy to get builds: similar to Go Modules)

6) Amazing .editorconfig from your nats-io/spring-nats

7) runs on modern JDK 21 (LTS), but builds Java-8-byte-code binaries thanks to --release 
https://stackoverflow.com/questions/43102787/what-is-the-release-flag-in-the-java-9-compiler/43103038

8) fixed broken NonNull contract (one, which I found)


scottf:
> PS: "Look, Mister...I'm not really a welder. I just found the mask." So, don't judge too hard and be sure to check everything.
> 
> It works on my Windows 10 machine: everything except publishing to sonatype of course 😅

I don't understand this comment. Also, it's good that you can't publish to sonatype

scottf:
1\. So the first thing in regard to this PR is it addresses too many things. When you submit a PR it should be targeted to a specific functionality.

I could see make a pr to change gradle and the build jdk since they are both build related.

Want to add static tools, fine, make it's own PR and we can look at it.

2\. **lombok** There are two reasons this isn't used. 1. The original developers of the library intentionally did not include 3rd party libraries unless absolutely required. 2. Code generation is a religious argument in regard to it's usefulness and impact on the build.

3.\ **reproducible build**
We can take a look at this independently as noted in item 1.

4.\ **jitpack** I don't understand the point of this, we already have a publish mechanism. Snapshots are published every time code is merged to main. We have the ability to publish a branch snapshot on as as needed basis. With recent sonatype upgrades snapshots are available immediately after build an release within 30 minutes. The location is fully documented in the readme.

5\. **editorconfig** Ok, but in it's own PR.


6\. **broken NonNull contract** In it's own PR.

Also, regarding PRs. All commits must be verified.

magicprinc:
It requires a full-time job to satisfy all the requirements. 
I am really sorry, but I don't have so much free time 😭
I spent several hours to make this PR.
I hope you will find time to investigate it later and use anything you find useful 🤝

magicprinc:
> I don't understand this comment. 

This was a phrase from a joke, it means (in my environment): "I am not a real expert in this field, so please review and verify".

magicprinc:
A little hobby of mine: 

Memory efficient JsonValue with inheritance instead of dozens of fields

I hope you will find this interesting.
No PR of course 🤝

https://github.com/magicprinc/nats.java/commit/1d3c7bd19f77d744526f062a1fa60b140b56d8fc?w=1

scottf:
I reviewed the part about non null contract broken. If you would like you can submit a PR for just this. Since it will be your first submission, I will relax the requirement for verified commit.

magicprinc:
I try to figure out how to make my commits verified. A good thing, I need this.

Should I make new compact PR with only one commit and one fix?

magicprinc:
> JsonValue could certainly be optimized. Values could also be initialized retrieved in a more lazy fashion and only pulled from the jv/map as needed.


Actually, I am a big fan of vert.x (https://vertx.io/)
JsonArray - https://github.com/eclipse-vertx/vert.x/blob/master/vertx-core/src/main/java/io/vertx/core/json/JsonArray.java
and
JsonObject - https://github.com/eclipse-vertx/vert.x/blob/master/vertx-core/src/main/java/io/vertx/core/json/JsonObject.java

I have read a book about "Data Oriented Programming".
And JsonArray/JsonObject are almost as in this book, except they are not immutable.

I write this only to cause your curiosity if you haven't seen them.
Maybe it can help you somehow 🙏

PR #1374, opened 2025-07-27
Draft Concept: usual Gradle layout

magicprinc:
More radical approach than https://github.com/nats-io/nats.java/pull/1373

Currently, nats.java has unusual layout with examples inside core library 😱

This PR splits monolith project into two subprojects: core library and examples 👍

Everything works as expected. All JARs are more compact and easy identifiable.

👀👉 Publishing to sonatype IS NOT tested and probably needs refactoring 🤷‍♀️ (this is the reason of "Draft Concept" in the subject).


scottf:
In general, I agree that the examples should be split off to their own project, but I have to be cognizant of the fact that it's been this way for a long time. It also makes it easy to just use them without need a full project that always has to keep up to date on the latest released library.
The build handles this. I think this gets filed under if it ain't broke don't fix it.

PR #1376, opened 2025-07-28
Bugfix NonNull contract and two javadoc links

magicprinc:
The changes we talked about
Verified commits 🤝
https://github.com/nats-io/spring-nats/blob/main/.editorconfig added to strictly comply with your source code formatting standard (I had a sad experience when IDEA replaced spaces with TABs (my default IDE setting) and the project owner had to fix the PR manually)


magicprinc:
About
```
public @NonNull String toJson()
```

JSpecify `@NonNull` is a type annotation: it applies to types.
So Oracle and IDEA are insisting on putting it close to the type 🤷‍♀️

magicprinc:
There is a way to semi-automatize nullability contract verification.
https://github.com/uber/NullAway
https://www.baeldung.com/java-nullaway

I personally haven't used it, but Spring says they use it intensively https://spring.io/blog/2025/03/10/null-safety-in-spring-apps-with-jspecify-and-null-away to develop Spring 7 (Nov 2025)

scottf:
1. editor config should be it's own pr.
2. What is with the Options class? github shows the entire class has changed. This is no good.
3. I use IntelliJ and the annotation placements are perfectly fine, and do not need changing. I have no idea why you are having issues.
4. Nullaway can be in it's own PR, but I have 94% code coverage and nullaway is of extremely low importance to me.

Last and most important. Name can be null and must stay as null. I realize this is not obvious. As part of creating a stream info, name can be null. I thought the problem was that I had marked it as non-null and thought that was what you were going to fix.

As a looked deeper I realized that some of the code, specifically StreamInfo creation needed it to be nullable. So I fixed that. I and fixed the nullability. See #1377

PR #1381, opened 2025-08-04
Use String.replace instead of String.replaceAll

scottf:


src/examples/java/io/nats/examples/stability/StabilityPub.java
```diff
@@ -79,7 +79,7 @@ public static void main(String args[]) {
                     Instant finish = Instant.now();
                     System.out.printf("Running for %s\n", Duration.between(start, finish).toString()
                                                         .substring(2)
-                                                        .replaceAll("(\\d[HMS])(?!$)", "$1 ")
+                                                        .replace("(\\d[HMS])(?!$)", "$1 ")
```
magicprinc: Isn't it a regexp? 
String.replace treats it as plain string 😱

src/examples/java/io/nats/examples/stability/StabilitySub.java
```diff
@@ -71,7 +71,7 @@ public static void main(String args[]) {
                             Instant finish = Instant.now();
                             System.out.printf("Running for %s\n", Duration.between(start, finish).toString()
                                                                 .substring(2)
-                                                                .replaceAll("(\\d[HMS])(?!$)", "$1 ")
+                                                                .replace("(\\d[HMS])(?!$)", "$1 ")
```
magicprinc: Isn't it a regexp? 
String.replace treats it as plain string 😱

PR #1385, opened 2025-08-10
Headers +toString(), fixes

magicprinc:
**Headers.java**

1. I actually came to add toString(), required for debugging and logging;

Then I saw and:

2. Fixed nullability issues.

3. Replaced inefficient keySet() + Map.get() pattern with entrySet().

4. Fixed broken forEach loop; added a failing test to confirm the issue before the fix.

5. Replaced val.chars() with a more efficient for-loop and charAt().

6. Removed "redundant" `private final boolean readOnly` field; replaced with another check.

7. Finally, I have added missing tests.

---

I really don't have time for several one-line PRs 😭🙏
This PR is obvious and not bigger than your recent PRs.
I open to your questions, critic, and recommendations for improvement 🤝

-----

PS: Some java quirks (why I have changed those lines)

1) Collections.unmodifiableMap isn't completely unmodifiable 
`java.util.Collections.UnmodifiableMap#forEach`
```
m.forEach(action); // 😭🤷‍♀️
``` 

2) `java.util.Objects#hash`

```
public static int hash(Object... values) {
  return Arrays.hashCode(values);
}
```
it is overkill for ONE value

For one value, it is recommended to use
`java.util.Objects#hashCode`
```
public static int hashCode(Object o) {
  return o != null ? o.hashCode() : 0;
}
```

3) The Message of the thrown exception is probably unexpected
~ "Header value has invalid character: 42911"
c has type int ⇒ "+ c " creates decimal digit representation
but "+ (char) c" can be problematic too: invisible characters, special characters e.g. 
0x200B// Zero Width space
0x200C// Zero Width non-joiner
0x200D// Zero Width joiner
```
val.chars().forEach(c -> {
  if (c > 127 || c == 10 || c == 13) {
    throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS + c);
```

### DIFF
--- a/src/main/java/io/nats/client/impl/Headers.java
+++ b/src/main/java/io/nats/client/impl/Headers.java
@@ -1,4 +1,4 @@
-// Copyright 2020 The NATS Authors
+// Copyright 2020-2025 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at:
@@ -18,10 +18,7 @@ import org.jspecify.annotations.Nullable;
 
 import java.util.*;
 import java.util.function.BiConsumer;
-import java.util.stream.Collectors;
-
 import static io.nats.client.support.NatsConstants.*;
-import static java.nio.charset.StandardCharsets.US_ASCII;
 
 /**
  * An object that represents a map of keys to a list of values. It does not accept
@@ -403,8 +400,9 @@ public class Headers {
 	 * removed during iteration
 	 */
 	public void forEach(BiConsumer<String, List<String>> action) {
-		valuesMap.forEach((key, values) ->
-				action.accept(key, Collections.unmodifiableList(values)));
+		for (Map.Entry<String, List<String>> entry : valuesMap.entrySet()) {
+			action.accept(entry.getKey(), Collections.unmodifiableList(entry.getValue()));
+		}
 	}
 
 	/**
@@ -506,6 +504,7 @@ public class Headers {
 		dest[destPosition++] = CR;
 		dest[destPosition] = LF;
 
+		//to do update serialized and/or dataLength?
 		return serializedLength();
 	}
 
@@ -525,7 +524,8 @@ public class Headers {
 		for (int idx = 0; idx < len; idx++) {
 			char c = key.charAt(idx);
 			if (c < 33 || c > 126 || c == ':') {
-				throw new IllegalArgumentException(KEY_INVALID_CHARACTER + "'" + c + "'");
+				throw new IllegalArgumentException(KEY_INVALID_CHARACTER +"0x"+ Integer.toHexString(c)
+						+" at "+ idx +" in "+ key);
 			}
 		}
 	}
@@ -541,7 +541,8 @@ public class Headers {
 		for (int i = 0, len = val.length(); i < len; i++) {
 			int c = val.charAt(i);
 			if (c > 127 || c == 10 || c == 13) {
-				throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS + Integer.toHexString(c));
+				throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS +"0x"+ Integer.toHexString(c)
+						+" at "+ i +" in "+ val);
 			}
 		}
 	}
@@ -595,18 +596,21 @@ public class Headers {
 		return Objects.hashCode(valuesMap);
 	}
 
+	@SuppressWarnings("deprecation")
 	@Override
 	public String toString() {
-		return valuesMap.entrySet().stream()
-				.filter(e -> e.getValue() != null)
-				.sorted(Map.Entry.comparingByKey())
-				.map(e -> {
-					String headerName = e.getKey();
-					List<String> values = e.getValue();
-					return headerName +": "+ (values.size() == 1 ? values.get(0)
-							: String.join(", ", values)
-					);
-				})
-				.collect(Collectors.joining("; "));
+		byte[] b = getSerialized();
+		if (b.length <= HVCRLF_BYTES + 2){
+			return "";
+		}
+
+		for (int i = 0, len = b.length; i < len; i++) {
+			if (b[i] == LF) {
+				b[i] = ' ';
+			} else if (b[i] == CR) {
+				b[i] = ';';
+			}
+		}
+		return new String(b, 0, HVCRLF_BYTES, b.length - HVCRLF_BYTES - 3);
 	}
 }
--- a/src/main/java/io/nats/client/impl/Headers.java
+++ b/src/main/java/io/nats/client/impl/Headers.java
@@ -30,8 +30,8 @@ import static io.nats.client.support.NatsConstants.*;
 public class Headers {
 
 	private static final String KEY_CANNOT_BE_EMPTY_OR_NULL = "Header key cannot be null.";
-	private static final String KEY_INVALID_CHARACTER = "Header key has invalid character: ";
-	private static final String VALUE_INVALID_CHARACTERS = "Header value has invalid character: ";
+	private static final String KEY_INVALID_CHARACTER = "Header key has invalid character: 0x";
+	private static final String VALUE_INVALID_CHARACTERS = "Header value has invalid character: 0x";
 
 	private final Map<String, List<String>> valuesMap;
 	private final Map<String, Integer> lengthMap;
@@ -523,8 +523,7 @@ public class Headers {
 		for (int idx = 0; idx < len; idx++) {
 			char c = key.charAt(idx);
 			if (c < 33 || c > 126 || c == ':') {
-				throw new IllegalArgumentException(KEY_INVALID_CHARACTER +"0x"+ Integer.toHexString(c)
-						+" at "+ idx +" in "+ key);
+				throw new IllegalArgumentException(KEY_INVALID_CHARACTER + Integer.toHexString(c));
 			}
 		}
 	}
@@ -540,8 +539,7 @@ public class Headers {
 		for (int i = 0, len = val.length(); i < len; i++) {
 			int c = val.charAt(i);
 			if (c > 127 || c == 10 || c == 13) {
-				throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS +"0x"+ Integer.toHexString(c)
-						+" at "+ i +" in "+ val);
+				throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS + Integer.toHexString(c));
 			}
 		}
 	}
--- a/src/main/java/io/nats/client/impl/Headers.java
+++ b/src/main/java/io/nats/client/impl/Headers.java
@@ -38,6 +38,7 @@ public class Headers {
 
 	private final Map<String, List<String>> valuesMap;
 	private final Map<String, Integer> lengthMap;
+	private final boolean readOnly;
 	private byte @Nullable [] serialized;
 	private int dataLength;
 
@@ -70,6 +71,7 @@ public class Headers {
 				}
 			}
 		}
+		this.readOnly = readOnly;
 		if (readOnly) {
 			valuesMap = Collections.unmodifiableMap(tempValuesMap);
 			lengthMap = Collections.unmodifiableMap(tempLengthMap);
@@ -92,7 +94,7 @@ public class Headers {
 	 *         -or- if any value contains invalid characters
 	 */
 	public Headers add(String key, String... values) {
-		if (isReadOnly()) {
+		if (readOnly) {
 			throw new UnsupportedOperationException();
 		}
 		if (values == null || values.length == 0) {
@@ -113,7 +115,7 @@ public class Headers {
 	 *         -or- if any value contains invalid characters
 	 */
 	public Headers add(String key, Collection<String> values) {
-		if (isReadOnly()) {
+		if (readOnly) {
 			throw new UnsupportedOperationException();
 		}
 		if (values == null || values.isEmpty()) {
@@ -153,7 +155,7 @@ public class Headers {
 	 *         -or- if any value contains invalid characters
 	 */
 	public Headers put(String key, String... values) {
-		if (isReadOnly()) {
+		if (readOnly) {
 			throw new UnsupportedOperationException();
 		}
 		if (values == null || values.length == 0) {
@@ -174,7 +176,7 @@ public class Headers {
 	 *         -or- if any value contains invalid characters
 	 */
 	public Headers put(String key, Collection<String> values) {
-		if (isReadOnly()) {
+		if (readOnly) {
 			throw new UnsupportedOperationException();
 		}
 		if (values == null || values.isEmpty()) {
@@ -191,7 +193,7 @@ public class Headers {
 	 * @return the Headers object
 	 */
 	public Headers put(Map<String, List<String>> map) {
-		if (isReadOnly()) {
+		if (readOnly) {
 			throw new UnsupportedOperationException();
 		}
 		if (map == null || map.isEmpty()) {
@@ -228,7 +230,7 @@ public class Headers {
 	 * @param keys the key or keys to remove
 	 */
 	public void remove(String... keys) {
-		if (isReadOnly()) {
+		if (readOnly) {
 			throw new UnsupportedOperationException();
 		}
 		for (String key : keys) {
@@ -243,7 +245,7 @@ public class Headers {
 	 * @param keys the key or keys to remove
 	 */
 	public void remove(Collection<String> keys) {
-		if (isReadOnly()) {
+		if (readOnly) {
 			throw new UnsupportedOperationException();
 		}
 		for (String key : keys) {
@@ -282,7 +284,7 @@ public class Headers {
 	 * Removes all the keys The object map will be empty after this call returns.
 	 */
 	public void clear() {
-		if (isReadOnly()) {
+		if (readOnly) {
 			throw new UnsupportedOperationException();
 		}
 		valuesMap.clear();
@@ -577,7 +579,7 @@ public class Headers {
 	 * @return the read only state
 	 */
 	public boolean isReadOnly() {
-		return !(valuesMap instanceof HashMap);
+		return readOnly;
 	}
 
 	@Override
--- a/src/main/java/io/nats/client/impl/Headers.java
+++ b/src/main/java/io/nats/client/impl/Headers.java
@@ -16,6 +16,7 @@ package io.nats.client.impl;
 import io.nats.client.support.ByteArrayBuilder;
 import org.jspecify.annotations.Nullable;
 
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.function.BiConsumer;
 import static io.nats.client.support.NatsConstants.*;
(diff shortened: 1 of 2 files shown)


magicprinc:
PS: I think
```
private final Map<String, Integer> lengthMap;
private int dataLength;
```
can be optimized (removed).

If you accept this PR and give me the go-ahead: I can do this optimizations with a benchmark.


Anyway, not cached serializeToArray is used in `io.nats.client.impl.NatsMessage#copyNotEmptyHeaders` 😭😱

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -92,7 +92,7 @@ public Headers(Headers headers, boolean readOnly, String[] keysNotToCopy) {
 	 *         -or- if any value contains invalid characters
 	 */
 	public Headers add(String key, String... values) {
-		if (readOnly) {
+		if (isReadOnly()) {
```
scottf: Please revert this. See comment below about just use the variable, it's not redundant, it's fine.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -113,7 +113,7 @@ public Headers add(String key, String... values) {
 	 *         -or- if any value contains invalid characters
 	 */
 	public Headers add(String key, Collection<String> values) {
-		if (readOnly) {
+		if (isReadOnly()) {
```
scottf: Revert

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -153,7 +153,7 @@ private Headers _add(String key, Collection<String> values) {
 	 *         -or- if any value contains invalid characters
 	 */
 	public Headers put(String key, String... values) {
-		if (readOnly) {
+		if (isReadOnly()) {
```
scottf: Revert

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -174,7 +174,7 @@ public Headers put(String key, String... values) {
 	 *         -or- if any value contains invalid characters
 	 */
 	public Headers put(String key, Collection<String> values) {
-		if (readOnly) {
+		if (isReadOnly()) {
```
scottf: Revert

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -191,14 +191,14 @@ public Headers put(String key, Collection<String> values) {
 	 * @return the Headers object
 	 */
 	public Headers put(Map<String, List<String>> map) {
-		if (readOnly) {
+		if (isReadOnly()) {
```
scottf: Revert

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -191,14 +191,14 @@ public Headers put(String key, Collection<String> values) {
...
 			throw new UnsupportedOperationException();
 		}
 		if (map == null || map.isEmpty()) {
 			return this;
 		}
-		for (String key : map.keySet() ) {
-			_put(key, map.get(key));
+		for (Map.Entry<String, List<String>> entry : map.entrySet()) {
+			_put(entry.getKey(), entry.getValue());
```
scottf, line 201: This is done for performance reasons, correct?

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -228,7 +228,7 @@ private Headers _put(String key, Collection<String> values) {
 	 * @param keys the key or keys to remove
 	 */
 	public void remove(String... keys) {
-		if (readOnly) {
+		if (isReadOnly()) {
```
scottf: Revert

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -243,7 +243,7 @@ public void remove(String... keys) {
 	 * @param keys the key or keys to remove
 	 */
 	public void remove(Collection<String> keys) {
-		if (readOnly) {
+		if (isReadOnly()) {
```
scottf: Revert

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -282,7 +282,7 @@ public boolean isEmpty() {
 	 * Removes all the keys The object map will be empty after this call returns.
 	 */
 	public void clear() {
-		if (readOnly) {
+		if (isReadOnly()) {
```
scottf: Revert

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -331,7 +331,7 @@ public Set<String> keySet() {
 	 * @return a read-only set of keys (in lowercase) contained in this map
 	 */
 	public Set<String> keySetIgnoreCase() {
-		HashSet<String> set = new HashSet<>();
+		HashSet<String> set = new HashSet<>(valuesMap.size()*4/3 + 1);
```
scottf: Please explain how this formula was arrived at. What's wrong with making it the same size? It's not like a byte array boundary that we might want to keep to a power of 2

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -401,7 +401,8 @@ public List<String> getIgnoreCase(String key) {
 	 * removed during iteration
 	 */
 	public void forEach(BiConsumer<String, List<String>> action) {
-		Collections.unmodifiableMap(valuesMap).forEach(action);
+		valuesMap.forEach((key, values) ->
+				action.accept(key, Collections.unmodifiableList(values)));
```
scottf: Actually, completely remove the syntactic sugar / mapping, since on very small collections, the overhead is expensive and just plain old code is actually faster.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -535,11 +536,12 @@ private void checkKey(String key) {
 	private void checkValue(String val) {
 		// Like rfc822 section 3.1.2 (quoted in ADR 4)
 		// The field-body may be composed of any US-ASCII characters, except CR or LF.
-		val.chars().forEach(c -> {
+		for (int i = 0, len = val.length(); i < len; i++) {
+			int c = val.charAt(i);
 			if (c > 127 || c == 10 || c == 13) {
-				throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS + c);
+				throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS + Integer.toHexString(c));
```
scottf, line 545: Please prefix it with `0x` so it's clear is hex and not some string. We probably should do the same for the `KEY_INVALID_CHARACTER` message

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -36,8 +38,7 @@ public class Headers {
 
 	private final Map<String, List<String>> valuesMap;
 	private final Map<String, Integer> lengthMap;
-	private final boolean readOnly;
```
scottf, line 39: No. Use the variable.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -575,19 +577,34 @@ boolean hasValues() {
 	 * @return the read only state
 	 */
 	public boolean isReadOnly() {
-		return readOnly;
+		return !(valuesMap instanceof HashMap);
```
scottf: Revert

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -575,19 +577,34 @@ boolean hasValues() {
...
+				.filter(e -> e.getValue() != null)
+				.sorted(Map.Entry.comparingByKey())
+				.map(e -> {
+					String headerName = e.getKey();
+					List<String> values = e.getValue();
+					return headerName +": "+ (values.size() == 1 ? values.get(0)
+							: String.join(", ", values)
+					);
+				})
+				.collect(Collectors.joining("; "));
```
scottf: how about just
```
return new String(getSerialized());
```

This has the benefit of preparing the serialization, so if no other values are added, it's ready to go across the wire, plus the toString exactly matches what goes over the wire. You could add a replace("\r\n", ";") if you don't want a multi line string.

scottf:
The size of this pr is fine, since it's one class and all related. And yes I know my PR's are large a lot, but that's the nature of working something that affects many things, like nullability. It's a fine line and I'm sure I cross it sometimes.

src/main/java/io/nats/client/impl/Headers.java
magicprinc: 1) Performance: Fewer actions under the hood
2) in recent java versions (with var) the source code is also as compact as keySet()+get version
3) in concurrent environment (not this case, but anyway) entrySet version is more safe/consistent
4) IDEA suggests it as refactoring

src/main/java/io/nats/client/impl/Headers.java
magicprinc: TL;DR:
*4/3 (int) == * 1/0.75 (float/double), where 0.75 is default load factor
+1 is extra for small sizes, e.g: 0

Longer:
new HashSet<>(valuesMap.size() * 4 / 3 + 1) is better than new HashSet<>(valuesMap.size())
because it accounts for HashSet's internal capacity and load factor.
(HashSet ctor argument is actually called not size, but initialCapacity)

- HashSet in Java is backed by a HashMap with a default load factor of 0.75.
- HashSet automatically resizes when the number of elements exceeds capacity * load factor.
- If you set the initial capacity exactly to the number of elements (valuesMap.size()), the backing HashMap may resize internally during insertion because it triggers resizing after reaching 75% of capacity.
- To avoid resizing and improve performance, you provide a capacity larger than the expected number of elements

→ valuesMap.size() * 4/3 + 1 capacity: Avoids resizing by providing enough buckets upfront according to load factor.

src/main/java/io/nats/client/impl/Headers.java
magicprinc: You mean: 
1) Simple for-each using entrySet()-iterator
2) and Collections.unmodifiableList(values) is ok
?

src/main/java/io/nats/client/impl/Headers.java
magicprinc: Great idea 👍
I'm in favor of replacing \r and \n: if used for logging, they break text logs formatting

magicprinc:
I have applied all your recommendations 

Plus: new serializeToArray() is 3.7 times faster than previous one

src/main/java/io/nats/client/impl/Headers.java
scottf: @magicprinc Please explain.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -522,7 +523,8 @@ private void checkKey(String key) {
 		for (int idx = 0; idx < len; idx++) {
 			char c = key.charAt(idx);
 			if (c < 33 || c > 126 || c == ':') {
-				throw new IllegalArgumentException(KEY_INVALID_CHARACTER + "'" + c + "'");
+				throw new IllegalArgumentException(KEY_INVALID_CHARACTER +"0x"+ Integer.toHexString(c)
+						+" at "+ idx +" in "+ key);
```
scottf: Just this please.
```
private static final String KEY_INVALID_CHARACTER = "Header key has invalid character: 0x";
throw new IllegalArgumentException(KEY_INVALID_CHARACTER + Integer.toHexString(c));
```

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -535,11 +537,13 @@ private void checkKey(String key) {
 	private void checkValue(String val) {
 		// Like rfc822 section 3.1.2 (quoted in ADR 4)
 		// The field-body may be composed of any US-ASCII characters, except CR or LF.
-		val.chars().forEach(c -> {
+		for (int i = 0, len = val.length(); i < len; i++) {
+			int c = val.charAt(i);
 			if (c > 127 || c == 10 || c == 13) {
-				throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS + c);
+				throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS +"0x"+ Integer.toHexString(c)
+						+" at "+ i +" in "+ val);
```
scottf:
```
private static final String VALUE_INVALID_CHARACTERS = "Header value has invalid character: 0x";
throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS + Integer.toHexString(c)));
```

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -581,13 +585,31 @@ public boolean isReadOnly() {
...
+		}
+
+		for (int i = 0, len = b.length; i < len; i++) {
+			if (b[i] == LF) {
+				b[i] = ' ';
+			} else if (b[i] == CR) {
+				b[i] = ';';
+			}
+		}
+		return new String(b, 0, HVCRLF_BYTES, b.length - HVCRLF_BYTES - 3);
```
scottf: I'm fully on board with squeezing every nano second out, but we can't use a deprecated method. So just simplify this. We are talking maybe nano seconds on an operation that probably isn't used all that much. Headers have been around for years and no one has asked for this, users probably write their own anyway.

```
@Override
public String toString() {
	byte[] b = getSerialized();
	for (int i = 0, len = b.length; i < len; i++) {
		switch (b[i]) {
			case CR: b[i] = ';'; break;
			case LF: b[i] = ' '; break;
		}
	}
	return new String(b, US_ASCII);
}
```

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -479,22 +480,21 @@ public ByteArrayBuilder appendSerialized(ByteArrayBuilder bab) {
 	 * @param dest the byte array to write to
 	 * @return the length of the header
 	 */
+	@SuppressWarnings("deprecation")
```
scottf, line 486: No on deprecated methods. I do like the part of assigning the key instead of calling `entry.getKey()` multiple times.

scottf (changes requested):

scottf:
> Plus: new serializeToArray() is 3.7 times faster than previous one
Explain this like I'm bad at math.

magicprinc:
I see... Github didn't publish my comments
Now they are published.

image

src/main/java/io/nats/client/impl/Headers.java
magicprinc: new String(b, US_ASCII)→new String(b, ISO_8859_1) ⇒ 3x times faster

Yes, in this case deprecated String ctor isn't faster than new String(b, ISO_8859_1) 👍

src/main/java/io/nats/client/impl/Headers.java
magicprinc: I agree with assigning the key.

But not with deprecated methods:
It is deprecated for years, still exists in JDK 24 and used for example in offical MSSQL jdbc driver.
In this case, it allows serializing 3.7 times faster:
instead of a "slow" US_ASCII encoder into new array
getBytes() simply copies low bytes of chars into your array.

Keys and values are US_ASCII: the input is checked for that, so both methods generate same bytes as output.

The result is absolutely the same, but the speed is much faster.

magicprinc:
> > Plus: new serializeToArray() is 3.7 times faster than previous one
> > Explain this like I'm bad at math.

In this case, it allows serializing 3.7 times faster:
instead of a "slow" US_ASCII encoder into new array
getBytes() simply copies low bytes of chars into your array.

Keys and values are US_ASCII: the input is checked for that, so both methods generate same bytes as output.

The result is absolutely the same, but the speed is much faster.

scottf:
> 6. serializeToArray uses deprecated getBytes to achieve 3.7x speed improvement, see p3 and explanation in comments above
>    JDK 24 has this method, Microsoft uses this method, and... 3.7x is 3.7x

> JDK25-26 has it: 
    public String(byte ascii[], int hibyte, int offset, int count) {

> @deprecated This method does not properly convert bytes into characters.

I'd actually be fine with a manual implementation that is similar to deprecated code but is tuned to recognize that we have actually already validated the key in checkKey, so we know that every char/byte c is 33 <= c <= 126

src/main/java/io/nats/client/impl/Headers.java
scottf: 1\. You could also make the initial capacity and the load factor 1. Seems like we know the size, it's never going to grow bigger, and we should instruct the code to do as close to that as we can.

2\. Code like this is fancy without adding a whole lot of value. We need to keep it simple, we don't need to squeeze water from a stone. Maintainability is just as important as anything else. I'm a firm no on the fancy formula.

magicprinc: I see 4 possible solutions:

0) a simpler formula: size * 4 / 3 like in JDK

1) (best) a utility function in a utility class ~
```
/** long explanation from native english speaker and link to JDK with example */
int capacityForSize(int size) {
  return size * 4 / 3;
}
```
then `new HashMap<>(capacityForSize(srcMap.size()))`

2) load factor 1: benchmarks are required! It can make everything worse.
HashMap uses closed addressing (!)
- When two or more keys hash to the same bucket, a collision occurs.

- Java HashMap resolves collisions by storing entries in a linked list (or, starting from Java 8, a balanced binary tree) at each bucket.

- This means multiple key-value pairs can be stored at the same bucket location.

- When inserting or retrieving a key, HashMap uses the key's hash code to find the bucket and then uses the key's equals() method to distinguish between entries in the bucket.

- If the number of entries in a bucket grows beyond a threshold, the linked list is replaced with a balanced binary tree to improve performance from O(n) to O(log n) in worst cases.

With fill factor 1 collisions are more likely, so memory consumption will probably increase → tests and benchmarks

3) use size without *4/3, but again benchmarks are required! It can make everything worse for small src-Maps

PS: if you open the sources of JDK:
In HashMap(Map) ctor, you see the same formula
```
  final void putMapEntries(Map<? extends K, ? extends V> m, boolean evict) {
        int s = m.size();
        if (s > 0) {
            if (table == null) { // pre-size
                double dt = Math.ceil(s / (double)loadFactor);//👈 the same 👍
                int t = ((dt < (double)MAXIMUM_CAPACITY) ?
                         (int)dt : MAXIMUM_CAPACITY);
                if (t > threshold)
          
```

magicprinc:
A copy of my thoughts about HashMap(capacity formula) to be sure you see it


I see 4 possible solutions:

0) a simpler formula: size * 4 / 3 like in JDK

1) (best) a utility function in a utility class ~

/** long explanation from native english speaker and link to JDK with example */
int capacityForSize(int size) {
  return size * 4 / 3;
}
then new HashMap<>(capacityForSize(srcMap.size()))

2) load factor = 1: benchmarks are required! It can make everything worse.
HashMap uses closed addressing (!)

- When two or more keys hash to the same bucket, a collision occurs.

- Java HashMap resolves collisions by storing entries in a linked list (or, starting from Java 8, a balanced binary tree) at each bucket.

- This means multiple key-value pairs can be stored at the same bucket location.

- When inserting or retrieving a key, HashMap uses the key's hash code to find the bucket and then uses the key's equals() method to distinguish between entries in the bucket.

- If the number of entries in a bucket grows beyond a threshold, the linked list is replaced with a balanced binary tree to improve performance from O(n) to O(log n) in worst cases.

- With fill factor 1 collisions are more likely, so memory consumption will probably increase → tests and benchmarks

3) use srcMap.size() without *4/3 as capacity (it will be two internal hashtable-array allocations), but again benchmarks are required! It can make everything worse for small src-Maps

PS: if you open the sources of JDK:
In HashMap(Map) ctor, you see the same formula
```
final void putMapEntries(Map<? extends K, ? extends V> m, boolean evict) {
  int s = m.size();
  if (s > 0) {
    if (table == null) { // pre-size
      double dt = Math.ceil(s / (double)loadFactor);//👈 the same 👍
      int t = ((dt < (double)MAXIMUM_CAPACITY) ? (int)dt : MAXIMUM_CAPACITY);
```

scottf:
> about HashMap
It would be so rare if there is more than 1 entry. Yes, people can put multiple values in, but this is an edge case. Just leave it as the default and we can move on.

magicprinc:
> > about HashMap
> > It would be so rare if there is more than 1 entry. Yes, people can put multiple values in, but this is an edge case. Just leave it as the default and we can move on.

The command is not understood. Please specify 🙏

magicprinc:
> > put checkKey and checkValue into Checker class:
> 
> Sounds good.

Done! 👍

scottf (approved):
LGTM

PR #1387, opened 2025-08-11
Ensuring nullability contracts

scottf:
- ConsumerInfo
- StreamInfo
- Unit tests reflect tightening
- Address IDE review notes

### DIFF
--- a/src/main/java/io/nats/client/ConsumeOptions.java
+++ b/src/main/java/io/nats/client/ConsumeOptions.java
@@ -17,7 +17,7 @@ package io.nats.client;
  * Consume Options are provided to customize the consume operation.
  */
 public class ConsumeOptions extends BaseConsumeOptions {
-    public static ConsumeOptions DEFAULT_CONSUME_OPTIONS = ConsumeOptions.builder().build();
+    public static final ConsumeOptions DEFAULT_CONSUME_OPTIONS = ConsumeOptions.builder().build();
 
     private ConsumeOptions(Builder b) {
         super(b);
--- a/src/main/java/io/nats/client/FetchConsumeOptions.java
+++ b/src/main/java/io/nats/client/FetchConsumeOptions.java
@@ -26,7 +26,7 @@ import static io.nats.client.support.JsonValueUtils.readLong;
  * Fetch Consume Options are provided to customize the fetch operation.
  */
 public class FetchConsumeOptions extends BaseConsumeOptions {
-    public static FetchConsumeOptions DEFAULT_FETCH_OPTIONS = FetchConsumeOptions.builder().build();
+    public static final FetchConsumeOptions DEFAULT_FETCH_OPTIONS = FetchConsumeOptions.builder().build();
 
     private final boolean noWait;
 
--- a/src/main/java/io/nats/client/api/ApiResponse.java
+++ b/src/main/java/io/nats/client/api/ApiResponse.java
@@ -18,10 +18,11 @@ import io.nats.client.Message;
 import io.nats.client.support.*;
 import org.jspecify.annotations.Nullable;
 
+import java.time.ZonedDateTime;
+
 import static io.nats.client.support.ApiConstants.ERROR;
 import static io.nats.client.support.ApiConstants.TYPE;
-import static io.nats.client.support.JsonValueUtils.readString;
-import static io.nats.client.support.JsonValueUtils.readValue;
+import static io.nats.client.support.JsonValueUtils.*;
 
 public abstract class ApiResponse<T> {
 
@@ -31,7 +32,7 @@ public abstract class ApiResponse<T> {
     protected final JsonValue jv;
 
     private final String type;
-    private final Error error;
+    private Error error;
 
     public ApiResponse(Message msg) {
         this(parseMessage(msg));
@@ -50,6 +51,33 @@ public abstract class ApiResponse<T> {
                 .put(TYPE, PARSE_ERROR_TYPE)
                 .toJsonValue();
         }
+    }
+
+    protected <R> R invalidJson(R retVal) {
+        // only set the error if it's not already set.
+        // this can easily happen when the original response is a real error
+        // but the parsing continues
+        if (error == null) {
+            error = new Error(500, "Invalid JSON for " + getClass().getSimpleName());
+        }
+        return retVal;
+    }
+
+    protected String nullStringIsError(JsonValue jv, String key) {
+        String s = readString(jv, key);
+        return s == null ? invalidJson("") : s;
+    }
+
+    @SuppressWarnings("SameParameterValue")
+    protected ZonedDateTime nullDateIsError(JsonValue jv, String key) {
+        ZonedDateTime zdt = readDate(jv, key);
+        return zdt == null ? invalidJson(DateTimeUtils.DEFAULT_TIME) : zdt;
+    }
+
+    @SuppressWarnings("SameParameterValue")
+    protected JsonValue nullValueIsError(JsonValue jv, String key, JsonValue errorValue) {
+        JsonValue v = readValue(jv, key);
+        return v == null ? invalidJson(errorValue) : v;
     }
 
     public ApiResponse(JsonValue jsonValue) {
--- a/src/main/java/io/nats/client/api/ConsumerInfo.java
+++ b/src/main/java/io/nats/client/api/ConsumerInfo.java
@@ -14,6 +14,7 @@
 package io.nats.client.api;
 
 import io.nats.client.Message;
+import io.nats.client.support.DateTimeUtils;
 import io.nats.client.support.JsonValue;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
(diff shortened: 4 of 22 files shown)


src/main/java/io/nats/client/ConsumeOptions.java
```diff
@@ -17,7 +17,7 @@
  * Consume Options are provided to customize the consume operation.
  */
 public class ConsumeOptions extends BaseConsumeOptions {
-    public static ConsumeOptions DEFAULT_CONSUME_OPTIONS = ConsumeOptions.builder().build();
+    public static final ConsumeOptions DEFAULT_CONSUME_OPTIONS = ConsumeOptions.builder().build();
```
scottf, line 20: Should have been final

src/main/java/io/nats/client/FetchConsumeOptions.java
```diff
@@ -26,7 +26,7 @@
  * Fetch Consume Options are provided to customize the fetch operation.
  */
 public class FetchConsumeOptions extends BaseConsumeOptions {
-    public static FetchConsumeOptions DEFAULT_FETCH_OPTIONS = FetchConsumeOptions.builder().build();
+    public static final FetchConsumeOptions DEFAULT_FETCH_OPTIONS = FetchConsumeOptions.builder().build();
```
scottf, line 29: Should have been final

src/main/java/io/nats/client/api/ApiResponse.java
```diff
@@ -31,7 +32,7 @@ public abstract class ApiResponse<T> {
     protected final JsonValue jv;
 
     private final String type;
-    private final Error error;
+    private Error error;
```
scottf, line 35: make this not final to support the subclass constructor setting it.
This is part of the effort to ensure that ConsumerInfo and StreamInfo can indicate they are an error if they don't parse correctly, because they can't throw an exception during construction, they didn't before.

src/main/java/io/nats/client/api/ConsumerInfo.java
```diff
@@ -51,26 +53,46 @@ public ConsumerInfo(Message msg) {
...
-        numPending = readLong(jv, NUM_PENDING, 0);
-        numWaiting = readLong(jv, NUM_WAITING, 0);
-        paused = readBoolean(jv, PAUSED, false);
-        pauseRemaining = readNanos(jv, PAUSE_REMAINING);
-
-        clusterInfo = ClusterInfo.optionalInstance(readValue(jv, CLUSTER));
-        pushBound = readBoolean(jv, PUSH_BOUND);
-
-        timestamp = readDate(jv, TIMESTAMP);
+        if (hasError()) {
```
scottf, line 56: When it's an already error there is no need to parse, but since vars are final we do need to set them.

src/main/java/io/nats/client/api/ConsumerInfo.java
```diff
@@ -51,26 +53,46 @@ public ConsumerInfo(Message msg) {
...
+            numWaiting = -1;
+            paused = false;
+            pauseRemaining = null;
+            clusterInfo = null;
+            pushBound = false;
+            timestamp = null;
+        }
+        else {
+            JsonValue jvConfig = nullValueIsError(jv, CONFIG, JsonValue.EMPTY_MAP) ;
+            configuration = ConsumerConfiguration.builder().jsonValue(jvConfig).build();
```
scottf, line 75: can't be null, but turns the object into an error. This would never happen except with data from somewhere other than the server, and all this is to just match the nullability contracts. Same with stream/name/created below.

src/main/java/io/nats/client/api/KeyValuePurgeOptions.java
```diff
@@ -20,7 +20,7 @@ public class KeyValuePurgeOptions {
     /**
      * The default time in millis that is used for as the threshold to keep markers.
      */
-    public static long DEFAULT_THRESHOLD_MILLIS = Duration.ofMinutes(30).toMillis();
+    public static final long DEFAULT_THRESHOLD_MILLIS = Duration.ofMinutes(30).toMillis();
```
scottf, line 23: should have been final

src/main/java/io/nats/client/api/ServerInfo.java
```diff
@@ -64,11 +65,11 @@ public ServerInfo(String json) {
...
         }
 
-        serverId = readString(jv, SERVER_ID, "N/A");
-        serverName = readString(jv, SERVER_NAME, "N/A");
+        serverId = readString(jv, SERVER_ID, UNDEFINED);
+        serverName = readString(jv, SERVER_NAME, UNDEFINED);
         version = readString(jv, VERSION, "0.0.0");
         go = readString(jv, GO, "0.0.0");
-        host = readString(jv, HOST, "N/A");
+        host = readString(jv, HOST, UNDEFINED);
```
scottf, line 72: made a reusable constant. Also will never happen on actual server data.

src/main/java/io/nats/client/api/SourceInfoBase.java
```diff
@@ -24,7 +24,7 @@
 import static io.nats.client.support.JsonValueUtils.*;
 
 abstract class SourceInfoBase {
-    protected JsonValue jv;
+    protected final JsonValue jv;
```
scottf, line 27: the IDE let me know, I checked it that no one else writes it and it was always the intention for only the base class to initialize it so I fixed it. jv is protected so it can be accessed by subclasses.

src/main/java/io/nats/client/api/StreamInfo.java
```diff
@@ -46,15 +48,31 @@ public StreamInfo(@NonNull Message msg) {
...
-        createTime = readDate(jv, CREATED);
-        JsonValue jvConfig = readValue(jv, CONFIG); // null when it's an error
-        config = jvConfig == null ? null : StreamConfiguration.instance(jvConfig);
-        streamState = new StreamState(readValue(jv, STATE));
-        clusterInfo = ClusterInfo.optionalInstance(readValue(jv, CLUSTER));
-        mirrorInfo = MirrorInfo.optionalInstance(readValue(jv, MIRROR));
-        sourceInfos = SourceInfo.optionalListOf(readValue(jv, SOURCES));
-        alternates = StreamAlternate.optionalListOf(readValue(jv, ALTERNATES));
-        timestamp = readDate(jv, TIMESTAMP);
+        if (hasError()) {
```
scottf, line 51: When it's an already error there is no need to parse, but since vars are final we do need to set them.

src/main/java/io/nats/client/api/StreamInfo.java
```diff
@@ -46,15 +48,31 @@ public StreamInfo(@NonNull Message msg) {
...
+            mirrorInfo = null;
+            sourceInfos = null;
+            alternates = null;
+            timestamp = null;
+        }
+        else {
+            JsonValue jvConfig = nullValueIsError(jv, CONFIG, JsonValue.NULL);
+            config = (jvConfig == JsonValue.NULL)
+                ? StreamConfiguration.builder().name(UNDEFINED).build()
+                : StreamConfiguration.instance(jvConfig);
```
scottf, line 65: can't be null, but turns the object into an error. This would never happen except with data from somewhere other than the server, and all this is to just match the nullability contracts. Same with create time below.

src/main/java/io/nats/client/impl/MessageManager.java
```diff
@@ -41,7 +41,7 @@ public enum ManageResult {MESSAGE, STATUS_HANDLED, STATUS_TERMINUS, STATUS_ERROR
 
     protected long lastStreamSeq;
     protected long lastConsumerSeq;
-    protected AtomicLong lastMsgReceivedNanoTime;
+    protected final AtomicLong lastMsgReceivedNanoTime;
```
scottf, line 44: should have been final

src/main/java/io/nats/client/impl/NatsKeyValueWatchSubscription.java
```diff
@@ -49,27 +50,33 @@ private void kvWatchInit(NatsKeyValue kv, List<String> keyPatterns, KeyValueWatc
...
             readSubjects.add(kv.readSubject(keyPattern.trim()));
         }
 
-        finishInit(kv, readSubjects, deliverPolicy, headersOnly, fromRevision, handler, watcher.getConsumerNamePrefix());
+        finishInit(kv,
+            readSubjects,
+            deliverPolicy,
+            headersOnly,
+            fromRevision,
+            getHandler(watcher, !ignoreDeletes),
```
scottf, line 64: The ide suggested and then helped me refactor the handler to a method

src/main/java/io/nats/client/impl/NatsObjectStoreWatchSubscription.java
```diff
@@ -44,21 +45,27 @@ public NatsObjectStoreWatchSubscription(NatsObjectStore os, ObjectStoreWatcher w
...
-                        sendEndOfData();
-                    }
-                }
-            };
+        finishInit(os,
+            Collections.singletonList(os.rawAllMetaSubject()),
+            deliverPolicy,
+            headersOnly,
+            ULONG_UNSET,
+            getHandler(watcher, !ignoreDeletes),
```
scottf, line 53: The ide suggested and then helped me refactor the handler to a method

PR #1389, opened 2025-08-12
JsonValueUtils: optimization, compaction, nullability

magicprinc:
1) replace JsonValueSupplier lambda with fast direct calls (faster, more readable 👍) 
2) remove unused ArrayBuilder
3) fix nullablity contracts

### DIFF
--- a/src/main/java/io/nats/client/support/JsonValueUtils.java
+++ b/src/main/java/io/nats/client/support/JsonValueUtils.java
@@ -1,4 +1,4 @@
-// Copyright 2023 The NATS Authors
+// Copyright 2023-2025 The NATS Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at:
@@ -14,50 +14,49 @@
 package io.nats.client.support;
 
 import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.ZonedDateTime;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.function.Function;
 
 import static io.nats.client.support.Encoding.base64BasicDecode;
-import static io.nats.client.support.JsonValue.*;
+import static io.nats.client.support.JsonValue.Type;
 
 /**
  * Internal json value helpers.
  */
 public abstract class JsonValueUtils {
-
     private JsonValueUtils() {} /* ensures cannot be constructed */
 
-    public interface JsonValueSupplier<T> {
-        T get(JsonValue v);
-    }
-
-    public static <T> T read(JsonValue jsonValue, String key, JsonValueSupplier<T> valueSupplier) {
-        JsonValue v = jsonValue == null || jsonValue.map == null ? null : jsonValue.map.get(key);
-        return valueSupplier.get(v);
-    }
-
-    public static JsonValue readValue(JsonValue jsonValue, String key) {
-        return read(jsonValue, key, v -> v);
+    public static @Nullable JsonValue readValue(@Nullable JsonValue jsonValue, String key) {
+        return jsonValue == null || jsonValue.map == null ? null
+                : jsonValue.map.get(key);
     }
 
     public static JsonValue readObject(JsonValue jsonValue, String key) {
-        return read(jsonValue, key, v -> v == null ? EMPTY_MAP : v);
+        JsonValue v = readValue(jsonValue,key);
+        return v == null ? JsonValue.EMPTY_MAP : v;
     }
 
     public static List<JsonValue> readArray(JsonValue jsonValue, String key) {
-        return read(jsonValue, key, v -> v == null ? EMPTY_ARRAY.array : v.array);
+        JsonValue v = readValue(jsonValue,key);
+        return v == null ? JsonValue.EMPTY_ARRAY.array : v.array;
     }
 
-    public static Map<String, String> readStringStringMap(JsonValue jv, String key) {
+    public static @Nullable Map<String, String> readStringStringMap(JsonValue jv, String key) {
         JsonValue o = readObject(jv, key);
         if (o.type == Type.MAP && o.map.size() > 0) {
-            Map<String, String> temp = new HashMap<>();
+            Map<String, String> temp = new HashMap<>(o.map.size() * 4/3 + 1);
             for (String k : o.map.keySet()) {
                 String value = readString(o, k);
                 if (value != null) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/nats/client/support/JsonValueUtils.java
```diff
@@ -14,50 +14,49 @@
...
  */
 public abstract class JsonValueUtils {
-
     private JsonValueUtils() {} /* ensures cannot be constructed */
 
-    public interface JsonValueSupplier<T> {
-        T get(JsonValue v);
-    }
-
-    public static <T> T read(JsonValue jsonValue, String key, JsonValueSupplier<T> valueSupplier) {
```
scottf: You cannot delete a public api method without a semver major version.

src/main/java/io/nats/client/support/JsonValueUtils.java
```diff
@@ -14,50 +14,49 @@
...
-        JsonValue v = jsonValue == null || jsonValue.map == null ? null : jsonValue.map.get(key);
-        return valueSupplier.get(v);
-    }
-
-    public static JsonValue readValue(JsonValue jsonValue, String key) {
-        return read(jsonValue, key, v -> v);
+    public static @Nullable JsonValue readValue(@Nullable JsonValue jsonValue, String key) {
+        return jsonValue == null || jsonValue.map == null ? null
+                : jsonValue.map.get(key);
```
scottf, line 53: I mean this is fine, but is it really necessary? Water from a stone.

src/main/java/io/nats/client/support/JsonValueUtils.java
```diff
@@ -150,59 +148,62 @@ public static <T> List<T> listOf(JsonValue v, Function<JsonValue, T> provider) {
...
     public static List<String> readStringList(JsonValue jsonValue, String key) {
-        return read(jsonValue, key, v -> listOf(v, jv -> jv.string));
+        JsonValue v = readValue(jsonValue,key);
+        return listOf(v, jv -> jv.string);
     }
 
     public static List<String> readStringListIgnoreEmpty(JsonValue jsonValue, String key) {
-        return read(jsonValue, key, v -> listOf(v, jv -> {
+        JsonValue v = readValue(jsonValue,key);
+        return listOf(v, jv -> {
```
scottf, line 173: Do no change formatting without function.

src/main/java/io/nats/client/support/JsonValueUtils.java
```diff
@@ -319,68 +321,28 @@ public MapBuilder put(String s, Object o) {
...
+                jv.map.put(key, mb.jv);
+                jv.mapOrder.add(key);
             }
             return this;
         }
 
         @Override
-        @NonNull
-        public String toJson() {
+        public @NonNull String toJson() {
```
scottf: I considered both ways and I prefer these where they where, on a separate line before the method.

src/main/java/io/nats/client/support/JsonValueUtils.java
```diff
@@ -319,68 +321,28 @@ public MapBuilder put(String s, Object o) {
...
-        }
-
-        @Deprecated
-        public JsonValue getJsonValue() {
+        public @NonNull JsonValue toJsonValue() {
             return jv;
         }
-    }
 
-    public static ArrayBuilder arrayBuilder() {
```
scottf, line 352: Cannot remove public api.

src/test/java/io/nats/client/support/JsonParsingTests.java
```diff
@@ -414,9 +415,14 @@ public void testGetMapped() {
         assertNotNull(readValue(v, "string"));
         assertNull(readValue(v, "na"));
         assertEquals(JsonValue.EMPTY_MAP, readObject(v, "na"));
-        assertNull(read(null, "na", vv -> vv));
-        assertNull(read(JsonValue.NULL, "na", vv -> vv));
-        assertNull(read(JsonValue.EMPTY_MAP, "na", vv -> vv));
```
scottf, line 419: Put these back when you put back the read method.

src/test/java/io/nats/client/support/JsonParsingTests.java
```diff
@@ -876,31 +888,6 @@ public void testValueUtilsInstanceArray() {
         validateArray(true, false, instance(list));
     }
 
-    @Test
-    public void testValueUtilsArrayBuilder() {
```
scottf: revert

src/test/java/io/nats/client/support/JsonParsingTests.java
```diff
@@ -950,4 +937,21 @@ public void testReadStringStringMap() {
         assertEquals("A", stringString.get("a"));
         assertEquals("B", stringString.get("b"));
     }
+
+    @Test
+    void testReadArray() {
```
scottf, line 971: Are you sure this isn't covered elsewhere?

scottf (changes requested):
So there are diminishing returns on these PRs and it's becoming very time consuming to address all these changes when there is very little gained.

scottf:
So I ran this test against the current code and the pr. Over 15 million operations the new was 8 seconds faster. 8 seconds = 8_000_000_000 nanoseconds (8 billion). This is a savings of 533 nanos per operations. Not nothing. But is it worth the time and effort both of us had to spend? I'm not so sure.

```
@Test
public void testReadSpeed() {
    JsonValue jvSI = JsonParser.parseUnchecked(dataAsString("StreamInfo.json"));
    JsonValue jvCC = JsonParser.parseUnchecked(dataAsString("ConsumerConfiguration.json"));
    NatsMessage accStatMsg = getDataMessage(dataAsString("AccountStatistics.json"));

    long start = System.currentTimeMillis();
    for (int x = 0; x < 5_000_000; x++) {
        new StreamInfo(jvSI);
        ConsumerConfiguration.builder().jsonValue(jvCC);
        new AccountStatistics(accStatMsg);
    }
    long elapsed = System.currentTimeMillis() - start;
    System.out.println("Elapsed: " + elapsed);
}
```

src/test/java/io/nats/client/support/JsonParsingTests.java
magicprinc: IDEA coverage shows it as uncovered 🤷‍♀️

magicprinc:
reverted 👍

IDEA shows as not covered
```
io.nats.client.support.JsonValueUtils#readLongList

io.nats.client.support.JsonValueUtils#readNanosList(io.nats.client.support.JsonValue, java.lang.String, boolean)

io.nats.client.support.JsonValueUtils#readBytes

io.nats.client.support.JsonValueUtils#readBase64

io.nats.client.support.JsonValueUtils.MapBuilder#put(java.lang.String, java.util.Map<java.lang.String,java.lang.String>) // patially: internal loop
```

src/main/java/io/nats/client/support/JsonValueUtils.java
```diff
@@ -69,65 +78,64 @@ public static Map<String, String> readStringStringMap(JsonValue jv, String key)
         return null;
     }
 
-    public static String readString(JsonValue jsonValue, String key) {
-        return read(jsonValue, key, v -> v == null ? null : v.string);
+    public static @Nullable String readString(@Nullable JsonValue jsonValue, String key) {
+        JsonValue v = readValue(jsonValue,key);
+        return v == null ? null : v.string;// todo number can be converted to string
```
scottf, line 83: Is this todo correct?

magicprinc: This is an idea: I don't know all cases where JsonValue is used.
In the wild, you can find JSON like
"someField":42
and
"someField":"42"

So if you parse JSON from an unknown source, and you really want this value, it is a good idea to fallback.
Like you do in case of int: "int then long" and long: "long then int" 
(BTW double and float can be checked too 👍)

In some code I work with, there is an option to serialize long as string (no idea why)

scottf:
> Here is (not fair, not scientific) benchmark https://github.com/magicprinc/MagicBenchmarks/blob/main/jnats/src/main/java/jnats/LambdaBenchmark.java where plain old java code is 2.9 times faster then new-API-and-lambda

My benchmark was fair and scientific enough. I just ran the code multiple times with nano second granularity, as much as possible in java and a Windows OS. 70 million total executions seems a large enough sample size to get an accurate pictures. I ran that sample size multiple times and many multiple times with smaller sample sizes. My machine had as little as possible open and running, with plenty of memory to spare. I could hear my cpu fan engage, it comes on pretty quick.

Maybe your results had to do with your jvm/os combination. But I actually think my slow machine is a fine judge of real world conditions where apps are deployed on the cloud in shared instances.

Is it perfect? Nope. It's good enough. There is no point in arguing the minutiae, it's not productive. I think I've been very accommodating, very accepting. I accepted the using a deprecated method so as not to be excessively formal, to be flexible. I've made an effort to understand everything you've put forward. There are bigger fish to fry and we need to move on.

If you want to submit a PR with complete and correct nullability annotations, great, or add test fine, but I'm closing this one.

magicprinc:
PS:
if you change implementation so, that
`public static List<Duration> readNanosList(@Nullable JsonValue jsonValue, String key)` is main function
and 
`public static @Nullable List<Duration> readNanosList(@Nullable JsonValue jsonValue, String key, boolean nullIfEmpty)` calls it 
(as I did) then
IDEA can understand its logic and doesn't require 

`//noinspection DataFlowIssue readNanosList with false ensures not null;`
before
`backoff(readNanosList(jsonValue, BACKOFF, false).toArray(new Duration[0]));`
in `io.nats.client.api.ConsumerConfiguration.Builder#jsonValue`

scottf:
I went so far as to ask Claude

```
● Yes, that's absolutely correct and a great observation! Modern Java performance is much more sophisticated than many developers realize.

  Why your benchmark showed minimal improvement:

  1. JIT Compiler Optimizations: The HotSpot JVM's C2 compiler is extremely aggressive at inlining, especially for:
    - Small methods (< 35 bytecodes by default)
    - Simple lambdas that compile to single expressions
    - Method references to static methods
    - Frequently called code paths ("hot" methods)
  2. Lambda Optimization: Since Java 8, lambda expressions often compile to invokedynamic bytecode, which the JVM can optimize very efficiently:
    - Simple lambdas like v -> v.string get inlined automatically
    - The lambda "allocation" overhead is often eliminated entirely
    - Method references are typically as fast as direct calls after warmup
  3. Escape Analysis: The JVM can detect when lambda objects don't escape their method scope and eliminate the allocation entirely.
  4. Profile-Guided Optimization: The JIT compiler profiles actual usage patterns and optimizes the hot paths accordingly.

  When manual inlining still matters:
  - Cold code paths that don't get JIT compiled
  - Very early in application startup before warmup
  - Extremely latency-sensitive code where even JIT compilation time matters
  - Complex lambdas that exceed inlining thresholds

  Your experience is typical - microbenchmarks often show that well-written lambda code performs identically to manually inlined code after JIT warmup. The readability and
  maintainability benefits of lambdas usually outweigh the theoretical performance concerns.
```

PR #1394, opened 2025-08-14
Not PR: architectural concept: compact JsonValue

magicprinc:
This is not an actual PR, because it breaks public JsonValue API 🤷‍♀️

The size of the current JsonValue (only object itself) is ~ 96 bytes.

The size of new JsonValue is ~ 24 bytes ⇒ 4 times smaller.
(see `io.nats.client.support.JsonParsingTests#usedMemory`)

The speed of working with the object does not seem to have changed (within the margin of error)
(see `io.nats.client.support.JsonParsingTests#testReadSpeed`)

If there can be significant amounts of objects in memory, then the difference will be noticeable.
E.g: for 50_000_000 objects
old 4.77 G
new 1.332 G

It passes the unit tests.

You might probably find this useful as a proof of concept or a source of inspiration 🤝

### DIFF
--- a/src/examples/java/io/nats/examples/service/ServiceExample.java
+++ b/src/examples/java/io/nats/examples/service/ServiceExample.java
@@ -16,6 +16,7 @@ package io.nats.examples.service;
 import io.nats.client.*;
 import io.nats.client.support.JsonSerializable;
 import io.nats.client.support.JsonValue;
+import io.nats.client.support.JsonValueObject;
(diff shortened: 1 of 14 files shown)


magicprinc:
I have a two-week vacation starting tomorrow without a laptop, so I sent this "PR" to summarize all my findings in jnats. 
Unfortunately, I will be unavailable for the next few days.

Thanks again for the great project!

PR #1399, opened 2025-08-20
ObjectMeta metadata

scottf:
The java client has ObjectMeta Headers, which is a map of string (key) to list of strings (value), but does not have the additional, almost identical field of metadata, which is a map of string (key) to string (value).


scottf (commented):
It looks like a lot, but I added some missing api doc and also added some null annotations and checking, but the functionality added is just metadata.

PR #1412, opened 2025-09-04
Expose connection outgoingPendingMessageCount and outgoingPendingBytes

scottf:
MessageQueue is used in multiple ways, for both incoming and outgoing messages, but this PR focuses on NatsConnectionWriter which is used for outgoing messages.

NatsConnectionWriter is used by NatsConnection to publish messages. (The concept of Producer is a Kafka concept and is more analogous to your specific application component that publishes messages. NATS itself has no concept of a Producer, and for that matter using broker is confusing as well as that is also a Kafka concept.)

The MessageQueue's
length is a count of items [messages] in the queue. sizeInBytes is the number of bytes calculated from the serialized form of all the messages in the queue. 

The change to NatsConnectionWriter exposes the MessageQueue's length and sizeInBytes with two new internal (packaged scoped) messages that specifically expose the Outgoing MessageQueue's information 

```
long outgoingPendingMessageCount() {
    return outgoing.length();
}

long outgoingPendingBytes() {
    return outgoing.sizeInBytes();
}
```

The change to the Connection interface and the NatsConnection implementation exposes publicly the length and sizeInBytes from the NatsConnectionWriter.

```
public long outgoingPendingMessageCount() {
    return writer.outgoingPendingMessageCount();
}

public long outgoingPendingBytes() {
    return writer.outgoingPendingBytes();
}
```

### DIFF
--- a/src/main/java/io/nats/client/Connection.java
+++ b/src/main/java/io/nats/client/Connection.java
@@ -788,7 +788,7 @@ public interface Connection extends AutoCloseable {
 
     /**
      * Get the number of messages in the outgoing queue for this connection.
-     * This value is volatile in the sense that it changes often.
+     * This value is volatile in the sense that it changes often and may be adjusted by more than one message.
      * It changes every time a message is published (put in the outgoing queue)
      * and every time a message is removed from the queue to be written over the socket
      * @return the number of messages in the outgoing queue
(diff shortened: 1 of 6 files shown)


jnmoyne (approved):
LGTM

Issue #1426, opened 2025-09-18
NATS Client "Authorization Violation" during server rolling updates


ajax-surovskyi-y:
### Observed behavior

During NATS server rolling updates, a client connection issue occurs that leads to permanent connection failure.

Impact is critical: Complete connection failure during rolling updates

### Expected behavior

When reconnectMode is false but reconnectOutgoing queue is not empty (due to previous timeout), the writer prioritizes messages from reconnectOutgoing queue over the normal outgoing queue. This prevents the CONNECT command from being sent first during a new connection establishment.

### Server and client version

jnats: 2.21.4
nats server: 2.11.4

### Steps to reproduce

1. Client is connected to NATS node 1
2. Node 1 is restarted (rolling update)
3. Client reconnects to node 2
4. During flush on subscription restoration, a timeout occurs due to large number of subscriptions
5. Timeout error is ignored, connection is established, but only partial subscriptions are restored
6. Subscription restoration commands remain in reconnectOutgoing queue
7. reconnectMode flag is set to false, allowing normal operation
8. Node 2 is restarted (continuing rolling update)
9. Client attempts to connect to node 3
10. Due to non-empty reconnectOutgoing queue, CONNECT command is not sent as first packet
11. Server responds with "authorization violation"
12. Connection permanently fails after multiple retry attempts

ajax-surovskyi-y:
I've written a test to reproduce this issue in #1427. The test confirms that authorization violations can occur during reconnect when flush timeout happens.

I'm considering introducing a handshake mode to isolate the two-protocol message flow (connect + ping) that establishes the connection. This would help prevent such race conditions by clearly separating the connection establishment phase from normal operation.

I'm also thinking about adding an option to optionally disable reconnect termination on isAuthenticationError. In our use case, this would be perfectly acceptable since our NATS cluster operates in a private network where all connections are controlled. This would provide more flexibility for environments where authentication errors are expected to be transient.
If this approach sounds reasonable, I'd be happy to create a PR with these changes. What are your thoughts on this solution?

scottf:
Would it be easier to just clear the reconnect outgoing queue before a connect starts?

scottf:
Are you sure?
> Due to non-empty reconnectOutgoing queue, CONNECT command is not sent as first packet

scottf:
I'm publishing a snapshot, 2.22.1.clearqueue-SNAPSHOT, can you try with that, or get the `clearqueue` branch

[...]

PR #1427, opened 2025-09-18
Test to reproduce #1426

ajax-surovskyi-y:
Provide handshake mode


ajax-surovskyi-y:
I cleaned up the code and left only the test

PR #1431, opened 2025-09-24
Exposed StreamName

roeschter:
Exposed StreamName to NatsJetstreamSubscription and extended ErrorListener to log it.

The change is fairly simple and has not hidden dependencies. The NatsJetstreamSubscription was only implemented once. The streamName was already present (but protected).


scottf (approved):
LGTM

PR #1438, opened 2025-09-26
Improve ConnectionListener

scottf:
A new method with a default implementation is added to ConnectionListener. The caller of connection listeners now calls the new method allowing implementors to have extra details about the uri related to this connection event. This information was not always available from the Connection object server info, for instance on reconnect, the connection will be stale, and the details contain information about the uri trying to be connected to and the host uri if the try to connect uri is an ip address resolved from the host uri.

```
default void connectionEvent(Connection conn, Events type, String uriDetails) {
    connectionEvent(conn, type);
}
```

### DIFF
--- a/src/main/java/io/nats/client/ConnectionListener.java
+++ b/src/main/java/io/nats/client/ConnectionListener.java
@@ -70,12 +70,26 @@ public interface ConnectionListener {
     }
 
     /**
+     * @deprecated use new api that gives additional details
      * Connection related events that occur asynchronously in the client code are
      * sent to a ConnectionListener via a single method. The ConnectionListener can
      * use the event type to decide what to do about the problem.
-     * 
+     *
      * @param conn the connection associated with the error
      * @param type the type of event that has occurred
      */
+    @Deprecated
     void connectionEvent(Connection conn, Events type);
+
+    /**
+     * Connection related events that occur asynchronously in the client code are
+     * sent to a ConnectionListener via a single method. The ConnectionListener can
+     * use the event type to decide what to do about the problem.
+     * @param conn the connection associated with the error
+     * @param type the type of event that has occurred
+     * @param details extra details about the event
+     */
+    default void connectionEvent(Connection conn, Events type, String details) {
+        connectionEvent(conn, type);
+    }
 }
--- a/src/main/java/io/nats/client/ConnectionListener.java
+++ b/src/main/java/io/nats/client/ConnectionListener.java
@@ -87,9 +87,9 @@ public interface ConnectionListener {
      * use the event type to decide what to do about the problem.
      * @param conn the connection associated with the error
      * @param type the type of event that has occurred
-     * @param details extra details about the event
+     * @param uriDetails extra details about the uri related to this connection event
      */
-    default void connectionEvent(Connection conn, Events type, String details) {
+    default void connectionEvent(Connection conn, Events type, String uriDetails) {
         connectionEvent(conn, type);
     }
 }
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -1922,13 +1922,13 @@ class NatsConnection implements Connection {
         }
     }
 
-    void processConnectionEvent(Events type, String details) {
+    void processConnectionEvent(Events type, String uriDetails) {
         if (!this.callbackRunner.isShutdown()) {
             try {
                 for (ConnectionListener listener : connectionListeners) {
                     this.callbackRunner.execute(() -> {
                         try {
-                            listener.connectionEvent(this, type, details);
+                            listener.connectionEvent(this, type, uriDetails);
                         } catch (Exception ex) {
                             this.statistics.incrementExceptionCount();
                         }
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -508,8 +508,7 @@ class NatsConnection implements Connection {
     // will wait for any previous attempt to complete, using the reader.stop and
     // writer.stop
     void tryToConnect(NatsUri cur, NatsUri resolved, long now) {
-        lastServer = currentServer;
-        currentServer = null;
+        clearCurrentServer();
 
         try {
             Duration connectTimeout = options.getConnectionTimeout();
@@ -672,6 +671,13 @@ class NatsConnection implements Connection {
                 statusLock.unlock();
             }
         }
+    }
+
+    private void clearCurrentServer() {
+        if (currentServer != null) {
+            lastServer = currentServer;
+        }
+        currentServer = null;
     }
 
     void checkVersionRequirements() throws IOException {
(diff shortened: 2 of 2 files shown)


src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -224,7 +225,7 @@ else if (cur.equals(first)) {
                 connectError.set(""); // new on each attempt
 
                 timeTraceLogger.trace("setting status to connecting");
-                updateStatus(Status.CONNECTING);
+                updateStatus(Status.CONNECTING, resolved, cur);
```
scottf, line 228: cur is the NatsUri from the server pool. resolved is either cur or a host resolved to an ip address

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -453,11 +454,12 @@ else if (first.equals(cur)) {
...
         try {
             this.flush(this.options.getConnectionTimeout());
-        } catch (Exception exp) {
+        }
+        catch (Exception exp) {
             this.processException(exp);
         }
 
-        processConnectionEvent(Events.RESUBSCRIBED);
+        processConnectionEvent(Events.RESUBSCRIBED, uriDetail(currentServer));
```
scottf, line 462: uri detail is a helper that converts the parameter to a string, making sure it's not null. Just a macro so I didn't have to have a lot of if == null checks

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -485,7 +487,7 @@ else if (remainingNanos > -1_000_000_000) { // less than -1 second
                 timeTraceLogger.trace(message + String.format(", %d (ms) beyond timeout", ms));
             }
             else {
-                double seconds = ((double)-remainingNanos) / 1_000_000_000.0;
+                double seconds = ((double) -remainingNanos) / 1_000_000_000.0;
```
scottf, line 490: Um, thanks intellij?

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -506,7 +508,7 @@ else if (remainingNanos < 1_000_000_000) {
     // will wait for any previous attempt to complete, using the reader.stop and
     // writer.stop
     void tryToConnect(NatsUri cur, NatsUri resolved, long now) {
-        currentServer = null;
+        clearCurrentServer();
```
scottf, line 511: clearCurrentServer sets current server to null after setting the lastServer variable, which helps reporting statuses during re/connect when current server needs to be null.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -569,7 +572,7 @@ void tryToConnect(NatsUri cur, NatsUri resolved, long now) {
                     // If the time appears too long, it might be related to
                     // https://github.com/nats-io/nats.java#linux-platform-note
                     timeTraceLogger.trace("TLS upgrade took: %.3f (s)",
-                            ((double) (NatsSystemClock.nanoTime() - start)) / NANOS_PER_SECOND);
+                        ((double) (NatsSystemClock.nanoTime() - start)) / NANOS_PER_SECOND);
```
scottf, line 575: again, thanks intellij

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -582,7 +585,8 @@ void tryToConnect(NatsUri cur, NatsUri resolved, long now) {
             Future<Object> future = this.connectExecutor.submit(connectTask);
             try {
                 future.get(timeoutNanos, TimeUnit.NANOSECONDS);
-            } finally {
+            }
+            finally {
```
scottf, line 589: I must have formatted this entire method.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -639,31 +643,43 @@ void tryToConnect(NatsUri cur, NatsUri resolved, long now) {
...
                 statusLock.unlock();
             }
         }
     }
 
+    private void clearCurrentServer() {
+        if (currentServer != null) {
+            lastServer = currentServer;
+        }
```
scottf, line 679: I know this looks weird, but this makes sure lastServer is retained when clearServer is called when currentServer is already null, it can happen.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1762,12 +1813,12 @@ void handleInfo(String infoJson) {
         List<String> urls = this.serverInfo.get().getConnectURLs();
         if (!urls.isEmpty()) {
             if (serverPool.acceptDiscoveredUrls(urls)) {
-                processConnectionEvent(Events.DISCOVERED_SERVERS);
+                processConnectionEvent(Events.DISCOVERED_SERVERS, urls.toString());
```
scottf, line 1817: now the connection event can see the discovered servers

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1762,12 +1813,12 @@ void handleInfo(String infoJson) {
...
         if (!urls.isEmpty()) {
             if (serverPool.acceptDiscoveredUrls(urls)) {
-                processConnectionEvent(Events.DISCOVERED_SERVERS);
+                processConnectionEvent(Events.DISCOVERED_SERVERS, urls.toString());
             }
         }
 
         if (serverInfo.isLameDuckMode()) {
-            processConnectionEvent(Events.LAME_DUCK);
+            processConnectionEvent(Events.LAME_DUCK, uriDetail(currentServer));
```
scottf, line 1822: can see the server being lame ducked

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1899,19 +1957,34 @@ void executeCallback(ErrorListenerCaller elc) {
...
+    }
+
+    void processConnectionEvent(Events type, String uriDetails) {
         if (!this.callbackRunner.isShutdown()) {
             try {
                 for (ConnectionListener listener : connectionListeners) {
                     this.callbackRunner.execute(() -> {
                         try {
-                            listener.connectionEvent(this, type);
+                            listener.connectionEvent(this, type, uriDetails);
```
scottf, line 1988: This is the call to the new method in the ConnectionListener

src/main/java/io/nats/client/ConnectionListener.java
```diff
@@ -70,12 +70,26 @@ public String toString() {
...
+     * Connection related events that occur asynchronously in the client code are
+     * sent to a ConnectionListener via a single method. The ConnectionListener can
+     * use the event type to decide what to do about the problem.
+     * @param conn the connection associated with the error
+     * @param type the type of event that has occurred
+     * @param uriDetails extra details about the uri related to this connection event
+     */
+    default void connectionEvent(Connection conn, Events type, String uriDetails) {
+        connectionEvent(conn, type);
+    }
```
scottf, line 94: default impl required for backward compatibility. I've verified that there is no issue, I have lambda implementations of connection listener in unit tests, examples, tons of my scratch classes that are not checked in, etc. so I know this is safe to do and won't break anyone.

PR #1444, opened 2025-10-03
Improved reconnect determining when to switch queues

scottf:


src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -119,13 +127,21 @@ boolean isRunning() {
...
     void sendMessageBatch(NatsMessage msg, DataPort dataPort, StatisticsCollector stats) throws IOException {
         writerLock.lock();
         try {
             int sendPosition = 0;
             int sbl = sendBufferLength.get();
 
             while (msg != null) {
+                if (msg == END_RECONNECT) {
+                    mode.set(Mode.Outgoing);
+                    return;
```
ajax-surovskyi-y:
```suggestion
                    msg = msg.next;
                    continue;
```

We can lose all messages in this NatsMessage linked list when on of them is END_RECONNECT.
I created a test in https://github.com/nats-io/nats.java/pull/1445 to demonstrate it

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -47,11 +45,12 @@ class MessageQueue {
...
-    // Poison pill is a graphic, but common term for an item that breaks loops or stop something.
+    // SPECIAL MARKER MESSAGES
+    // A simple == is used to resolve if any message is exactly the static pill object in question
+    // ----------
+    // 1. Poison pill is a graphic, but common term for an item that breaks loops or stop something.
     // In this class the poison pill is used to break out of timed waits on the blocking queue.
-    // A simple == is used to check if any message in the queue is this message.
-    // /\ /\ /\ /\ which is why it is now a static. It's just a marker anyway.
-    protected static final NatsMessage POISON_PILL = new NatsMessage(POISON, null, EMPTY_BODY);
+    protected static final NatsMessage POISON_PILL = new NatsMessage("_poison", null, EMPTY_BODY);
```
scottf, line 53: I was in this file looking at the marker pill and ended up doing some refactoring. No functionality was changed and I clarified the comment.

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -356,4 +351,15 @@ void filter(Predicate<NatsMessage> p) {
...
+    void clear() {
+        editLock.lock();
+        try {
+            this.queue.clear();
+            this.length.set(0);
+            this.sizeInBytes.set(0);
+        } finally {
+            editLock.unlock();
+        }
+    }
```
scottf, line 364: This allows clearing of a queue.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -452,18 +451,9 @@ else if (first.equals(cur)) {
             }
         });
 
-        try {
-            this.flush(this.options.getConnectionTimeout());
-        }
-        catch (Exception exp) {
-            this.processException(exp);
-        }
```
scottf, line 460: I believe that the original intention here was to know when the re-subscribe messages, queued just before this block, finished publishing, meaning the reconnect mode had completed and the reconnect mode could be cleared. The problem was that with very large number of subscriptions, the flush could fail, triggering a reconnect loop.
This has been replaced with a marker in the reconnect queue, set via `writer.enterWaitingForEndReconnectMode()` which the writer will then automatically switch the mode itself when it reaches the marker.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -452,18 +451,9 @@ else if (first.equals(cur)) {
...
-        catch (Exception exp) {
-            this.processException(exp);
-        }
+        writer.endReconnect();
 
         processConnectionEvent(Events.RESUBSCRIBED, uriDetail(currentServer));
-
-        // When the flush returns, we are done sending internal messages,
-        // so we can switch to the non-reconnect queue
-        this.writer.setReconnectMode(false);
```
scottf, line 466: replaced by `writer.enterWaitingForEndReconnectMode`

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1674,7 +1664,7 @@ public Duration RTT() throws IOException {
         pongQueue.add(pongFuture);
         try {
             long time = NatsSystemClock.nanoTime();
-            writer.queueInternalMessage(new ProtocolMessage(PING_PROTO));
+            writer.queue(new ProtocolMessage(PING_PROTO));
```
scottf, line 1667: This was changed after a review of the internal queuing and determined it was kinda pointless. If they make an rtt call while normally connected, this message just goes into the normal queue anyway. If they made the rtt call during reconnect, it would have gone into the internal queue, but so what, it's time is going to be skewed anyway.
This whole RTT may need to be put into a "bypass" queue of some sort so on normal connect, it goes directly to the front of the line.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -27,13 +27,16 @@
...
 
 import static io.nats.client.support.BuilderBase.bufferAllocSize;
-import static io.nats.client.support.NatsConstants.CR;
-import static io.nats.client.support.NatsConstants.LF;
+import static io.nats.client.support.NatsConstants.*;
 
 class NatsConnectionWriter implements Runnable {
+    enum Mode {
+        Normal, Reconnect, WaitingForEndReconnect
+    }
```
scottf, line 39: Normal - regularly connected
Reconnect - Used during reconnect for internal messages like connect, ping, and re-subscribe 
WaitingForEndReconnect - So we still pull from the reconnectQueue, but all messages published after this go into the normal queue

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -58,7 +61,12 @@ class NatsConnectionWriter implements Runnable {
         writerLock = new ReentrantLock();
 
         this.running = new AtomicBoolean(false);
-        this.reconnectMode = new AtomicBoolean(sourceWriter != null);
+        if (sourceWriter == null) {
+            mode = new AtomicReference<>(Mode.Normal);
```
scottf, line 65: source writer is null on a new connection - it's added later.
source writer is not null on a reconnect, because the old one is passed to the new connection so the new connection can get the old connection's normal queue

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -193,11 +207,11 @@ public void run() {
 
             while (running.get() && !Thread.interrupted()) {
                 NatsMessage msg;
-                if (this.reconnectMode.get()) {
-                    msg = this.reconnectOutgoing.accumulate(sendBufferLength.get(), Options.MAX_MESSAGES_IN_NETWORK_BUFFER, reconnectTimeout);
+                if (mode.get() == Mode.Normal) {
+                    msg = this.normalOutgoing.accumulate(sendBufferLength.get(), Options.MAX_MESSAGES_IN_NETWORK_BUFFER, outgoingTimeout);
```
scottf, line 211: In normal mode, pull from the normal queue

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -193,11 +207,11 @@ public void run() {
...
             while (running.get() && !Thread.interrupted()) {
                 NatsMessage msg;
-                if (this.reconnectMode.get()) {
-                    msg = this.reconnectOutgoing.accumulate(sendBufferLength.get(), Options.MAX_MESSAGES_IN_NETWORK_BUFFER, reconnectTimeout);
+                if (mode.get() == Mode.Normal) {
+                    msg = this.normalOutgoing.accumulate(sendBufferLength.get(), Options.MAX_MESSAGES_IN_NETWORK_BUFFER, outgoingTimeout);
                 }
                 else {
-                    msg = this.outgoing.accumulate(sendBufferLength.get(), Options.MAX_MESSAGES_IN_NETWORK_BUFFER, outgoingTimeout);
+                    msg = this.reconnectOutgoing.accumulate(sendBufferLength.get(), Options.MAX_MESSAGES_IN_NETWORK_BUFFER, reconnectTimeout);
```
scottf, line 214: not normal mode is both the others, we need to get all the messages from the reconnect queue until the end mark is found, and then we can switch to the normal mode.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -218,24 +232,31 @@ public void run() {
         }
     }
 
-    void setReconnectMode(boolean tf) {
-        reconnectMode.set(tf);
+    void enterReconnectMode() {
+        reconnectOutgoing.clear();
+        mode.set(Mode.Reconnect);
```
scottf, line 237: 1. clear the queue. Should probably already be empty, but just in case.
2. set the mode

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -218,24 +232,31 @@ public void run() {
...
-    void setReconnectMode(boolean tf) {
-        reconnectMode.set(tf);
+    void enterReconnectMode() {
+        reconnectOutgoing.clear();
+        mode.set(Mode.Reconnect);
+    }
+
+    void enterWaitingForEndReconnectMode() {
+        reconnectOutgoing.push(END_RECONNECT);
+        mode.set(Mode.WaitingForEndReconnect);
```
scottf, line 242: 1. mark the reconnect queue 
2. set the mode

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -218,24 +232,31 @@ public void run() {
...
+        return this.normalOutgoing.push(msg);
     }
 
     void queueInternalMessage(NatsMessage msg) {
-        if (this.reconnectMode.get()) {
-            this.reconnectOutgoing.push(msg);
-        } else {
-            this.outgoing.push(msg, true);
+        if (mode.get() == Mode.Reconnect) {
+            reconnectOutgoing.push(msg);
```
scottf, line 256: If mode is Reconnect, messages go in that queue.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -218,24 +232,31 @@ public void run() {
...
+        }
+        else {
+            normalOutgoing.push(msg, true);
```
scottf, line 259: If mode is normal put in the normal queue
If mode is waiting for end marker, this means that all the reconnect specific messages have been queued, but we still need to read from the reconnect queue, but write to the normal queue

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
scottf: Fixed by break instead of return

PR #1449, opened 2025-10-07
Optimize header reading

scottf:


src/main/java/io/nats/client/support/IncomingHeadersProcessor.java
```diff
@@ -42,17 +42,15 @@ public IncomingHeadersProcessor(byte[] serialized) {
         Token terminus = new Token(serialized, serializedLength, serializedLength - 2, TokenType.CRLF);
         Token token = new Token(serialized, serializedLength, HEADER_VERSION_BYTES_LEN, null);
 
-        boolean hadStatus = false;
```
scottf, line 45: variable was never used in the initHeader function that it was passed to and there are no other uses, so it was removed.

mtmk (approved):
LGTM

src/main/java/io/nats/client/support/IncomingHeadersProcessor.java
```diff
@@ -90,7 +88,7 @@ private void initHeader(byte[] serialized, int len, Token tCrlf, boolean hadStat
             if (headers == null) {
                 headers = new Headers();
             }
-            headers.add(tKey.getValue(), tVal.getValue());
+            headers.add(tKey.getValueCheckKnownKeys(), tVal.getValue());
```
scottf, line 91: use the new function to get the key, checking against known keys.

src/main/java/io/nats/client/support/NatsKeyValueUtil.java
```diff
@@ -32,7 +32,7 @@ private NatsKeyValueUtil() {} /* ensures cannot be constructed */
     public static final int KV_STREAM_PREFIX_LEN = KV_STREAM_PREFIX.length();
     public static final String KV_SUBJECT_PREFIX = "$KV.";
     public static final String KV_SUBJECT_SUFFIX = ".>";
-    public static final String KV_OPERATION_HEADER_KEY = "KV-Operation";
+    public static final String KV_OPERATION_HEADER_KEY = NatsJetStreamConstants.KV_OPERATION_HEADER_KEY;
```
scottf, line 35: Putting in NatsJetStreamConstants to be able to not depend on KV in non kv code

src/main/java/io/nats/client/support/Token.java
```diff
@@ -74,6 +83,7 @@ public Token(byte[] serialized, int len, int cur, TokenType required) {
             }
             hasValue = true;
         }
+        valueLength = hasValue ? end - start + 1 : 0;
```
scottf, line 86: holding length in a variable removes the need to calculate it repeatedly

src/main/java/io/nats/client/support/Status.java
```diff
@@ -18,10 +18,15 @@
...
 public class Status {
 
-    public static final String FLOW_CONTROL_TEXT = "FlowControl Request";
-    public static final String HEARTBEAT_TEXT = "Idle Heartbeat";
+    public static final String FLOW_CONTROL_TEXT  = "FlowControl Request";
+    public static final String HEARTBEAT_TEXT     = "Idle Heartbeat";
     public static final String NO_RESPONDERS_TEXT = "No Responders Available For Request";
-    public static final String EOB_TEXT = "EOB";
+    public static final String EOB_TEXT           = "EOB";
+    public static byte[] FLOW_CONTROL_TEXT_BYTES  = FLOW_CONTROL_TEXT.getBytes();
```
mtmk: * can these be made 'final'?
* can 'getBytes()' specify charset to be safe and not use system default?

scottf: fixed and fixed

PR #1451, opened 2025-10-08
Header fine tuning

scottf:


src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -142,19 +142,17 @@ public Headers add(String key, Collection<String> values) {
...
-				// update the data length with the additional len
-				// update the lengthMap for the key to the old length plus the new length
-				List<String> currentSet = valuesMap.computeIfAbsent(key, k -> new ArrayList<>());
-				currentSet.addAll(checked.list);
-				dataLength += checked.len;
-				int oldLen = lengthMap.getOrDefault(key, 0);
-				lengthMap.put(key, oldLen + checked.len);
-				serialized = null; // since the data changed, clear this so it's rebuilt
-			}
+		CheckState checked = check(key, values);
```
scottf, line 145: the check method is a refactor of all the checks and validations, plus bringing everything inline.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -514,66 +507,51 @@ public int serializeToArray(int destPosition, byte[] dest) {
...
-		// key cannot be null or empty and contain only printable characters except colon
+	static CheckState check(String key, Collection<String> values) {
 		if (key == null || key.isEmpty()) {
 			throw new IllegalArgumentException(KEY_CANNOT_BE_EMPTY_OR_NULL);
 		}
-
-		int len = key.length();
-		for (int idx = 0; idx < len; idx++) {
+		// Check the key to ensure it matches the specification for keys.
+		int keyLen = key.length();
```
scottf, line 515: instead of repeatedly calling .length

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -514,66 +507,51 @@ public int serializeToArray(int destPosition, byte[] dest) {
...
 
-	/**
-	 * Check the key to ensure it matches the specification for keys.
-	 * @throws IllegalArgumentException if the key is null, empty or contains
-	 *         an invalid character
-	 */
-	static void checkKey(String key) {
-		// key cannot be null or empty and contain only printable characters except colon
+	static CheckState check(String key, Collection<String> values) {
 		if (key == null || key.isEmpty()) {
```
scottf, line 511: the refactor does the key validation. Before, _put did this and then the old checker also did it, meaning it was checked twice. _add did it correctly in the checker.

src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -514,66 +507,51 @@ public int serializeToArray(int destPosition, byte[] dest) {
...
+					// The field-body may be composed of any US-ASCII characters, except CR or LF.
+					for (int i = 0; i < valLen; i++) {
+						int c = val.charAt(i);
+						if (c > 127 || c == 10 || c == 13) {
+							throw new IllegalArgumentException(VALUE_INVALID_CHARACTERS + Integer.toHexString(c));
 						}
 					}
 				}
+				list.add(val);
+				len += keyLen + valLen + 3; // 3 is for the colon, cr and lf
```
scottf, line 542: length is zero if there were only null values

src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -191,20 +192,20 @@ private PublishAck processPublishResponse(Message resp, PublishOptions options)
     }
 
     private Headers mergePublishOptions(Headers headers, PublishOptions opts) {
+        if (opts == null) {
+            return headers;
+        }
+
```
scottf, line 198: if there aren't any opts, just short circuit. This is probably the most common way publish is called.

src/main/java/io/nats/client/impl/NatsJetStream.java
```diff
@@ -219,7 +220,8 @@ private Headers _merge(Headers h, String key, String value) {
         if (h == null) {
             h = new Headers();
         }
-        return h.add(key, value);
+        // this is always an internal header with one value per key
+        return h.put(key, Collections.singletonList(value));
```
scottf, line 224: switch to put because there will only every be one of each key.

PR #1452, opened 2025-10-09
Additional test coverage

scottf:


src/main/java/io/nats/client/impl/Headers.java
```diff
@@ -212,25 +215,74 @@ public Headers put(Map<String, List<String>> map) {
...
-			lengthMap.put(key, checked.len);
+			dataLength = dataLength - lengthMap.getOrDefault(key, 0) + collected.length;
+			valuesMap.put(key, collected.values);
+			lengthMap.put(key, collected.length);
 			serialized = null; // since the data changed, clear this so it's rebuilt
 		}
 		return this;
 	}
 
+	static final class ValuesAndLength {
```
scottf, line 241: This block was just moved in the file closer to where it was used.

PR #1465, opened 2025-10-22
Reduce calls to stats by combining.

scottf:
Backward compatible. New interface methods have default implementations which call the old, deprecated methods.


MauriceVanVeen (approved):
LGTM

PR #1483, opened 2025-11-19
Properly unsubscribe from dispatcher subject

scottf:
* Dispatcher unsubscribe by subject would unsubscribe all subject, not just target subject

### DIFF
--- a/src/test/java/io/nats/client/SubscriberTests.java
+++ b/src/test/java/io/nats/client/SubscriberTests.java
@@ -16,10 +16,7 @@ package io.nats.client;
 import org.junit.jupiter.api.Test;
 
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashSet;
-import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeoutException;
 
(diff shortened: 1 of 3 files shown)


src/test/java/io/nats/client/SubscriberTests.java
```diff
@@ -491,38 +488,6 @@ public void testInvalidSubjectsAndQueueNames() throws Exception {
         }
     }
 
-    @Test
-    public void testDispatcherMultipleSubscriptionsBySubject() throws Exception {
```
scottf, line 495: Moved this to dispatcher test

src/test/java/io/nats/client/impl/DispatcherTests.java
```diff
@@ -33,6 +35,52 @@
 // the done message (or should) - wanted to note that somewhere
 
 public class DispatcherTests {
+    @Test
+    public void testDispatcherMultipleSubscriptionsBySubject() throws Exception {
```
scottf, line 39: updated test to dispatcher subscribe to multiple subjects, then unsubscribe from only one, making sure that only the one subject was unsubscribed.

Ran the test against the old code, it failed as expected, against the new code it passes.

PR #1491, opened 2025-11-28
Object Store purge on re-put

scottf:
Purge old chunks if you put over the same object name


src/test/java/io/nats/client/impl/ObjectStoreTests.java
```diff
@@ -238,6 +238,7 @@ private static Object[] getInput(int size) {
             if (f.isFile()) {
                 long flen = f.length();
                 if (flen == size) {
+                    foundLen = flen;
```
scottf, line 241: was wrong, but really inconsequential since I never find an exact match.

PR #1492, opened 2025-12-03
Lazy initialize Options executors 

scottf:
Lazy initialize executors so as not to keep making instances if the option getter is called repeatedly.


MauriceVanVeen (approved):
LGTM

PR #1499, opened 2025-12-14
[Replaced] Message Queue Splitting / Improvements Pass 2

scottf:
Split the queue into it's 2 functions since the behavior and locking is significantly different. This makes it easier to properly lock makes each queue, WriterMessageQueue and ConsumerMessageQueue specific to their purpose.

- Tuning of locks
- WriterMessageQueue pause/filter combined since they are **always** called together.
- WriterMessageQueue filter optimized


src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -38,19 +37,26 @@ class MessageQueue {
...
+    protected final Lock countLock;
     protected final int maxMessagesInOutgoingQueue;
     protected final boolean discardWhenFull;
     protected final long offerLockNanos;
     protected final long offerTimeoutNanos;
     protected final Duration requestCleanupInterval;
 
-    // SPECIAL MARKER MESSAGES
+    // Used for POISON_PILL and NatsConnectionWriter.END_RECONNECT
+    static class MarkerMessage extends NatsMessage {
```
scottf: MarkerMessage class is made so I can use instanceof, instead of adding a boolean getter to the NatsMessage class. Claude says instanceof is actually significantly faster anyway.

MarkerMessages will not be counted for the message or bytes in reporting of length/sizeInBytes. MarkerMessages are NOT filtered out when stop is called, only ProtocolMessages are.

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -38,19 +37,26 @@ class MessageQueue {
     protected final AtomicInteger running;
     protected final boolean singleReaderMode;
     protected final LinkedBlockingQueue<NatsMessage> queue;
-    protected final Lock editLock;
+    protected final Lock pushLock;
+    protected final Lock countLock;
```
scottf: editLock is split into two. Count lock is for whenever the length/sizeInBytes is changed. pushLock is for whenever a message is pushed into the queue.

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -94,14 +101,27 @@ class MessageQueue {
...
+        countLock.lock();
         try {
             queue.drainTo(target.queue);
-            target.length.set(length.getAndSet(0));
-            target.sizeInBytes.set(sizeInBytes.getAndSet(0));
-        } finally {
-            editLock.unlock();
+            long messages = length.getAndSet(0);
+            long bytes = sizeInBytes.getAndSet(0);
+            target.count(messages, bytes);
```
scottf: split to ensure target counting is locked

src/main/java/io/nats/client/impl/MessageQueue.java
```diff
@@ -344,37 +382,43 @@ long sizeInBytes() {
...
-                    newQueue.add(cursor);
-                } else {
-                    sizeInBytes.addAndGet(-cursor.getSizeInBytes());
-                    length.decrementAndGet();
+            ArrayList<NatsMessage> tempQueue = new ArrayList<>();
+            this.queue.drainTo(tempQueue); // I realize this is a copy of a queue, but polling locks the queue every time
+            long removed = 0;
+            long bytesRemoved = 0;
+            for (NatsMessage msg : tempQueue) {
+                if (msg instanceof ProtocolMessage) {
```
scottf: All ProtocolMessages are filtered on stop, they are the only type of messages that are.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1604,7 +1604,7 @@ void sendConnect(NatsUri nuri) throws IOException {
             ByteArrayBuilder bab =
                 new ByteArrayBuilder(OP_CONNECT_SP_LEN + connectOptions.limit(), UTF_8)
                     .append(CONNECT_SP_BYTES).append(connectOptions);
-            queueInternalOutgoing(new ProtocolMessage(bab, false));
+            queueInternalOutgoing(new ProtocolMessage(bab));
```
scottf, line 1641: We used to say don't filter a connect message on stop, but if a stop happens when a connect is queued, the message should be discarded

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -114,7 +115,7 @@ Future<Boolean> stop() {
             try {
                 this.normalOutgoing.pause();
                 this.reconnectOutgoing.pause();
-                this.normalOutgoing.filter(NatsMessage::isProtocolFilterOnStop);
+                this.normalOutgoing.filterOnStop();
```
scottf: There was ever only one condition in this filtering which happens on stop.

src/main/java/io/nats/client/impl/ProtocolMessage.java
```diff
@@ -15,41 +15,35 @@
...
+import java.nio.charset.StandardCharsets;
+
 // ----------------------------------------------------------------------------------------------------
 // Protocol message is a special version of a NatsPublishableMessage extends NatsMessage
 // ----------------------------------------------------------------------------------------------------
 class ProtocolMessage extends NatsPublishableMessage {
-    final boolean filterOnStop;
-
-    ProtocolMessage(ByteArrayBuilder babProtocol, boolean filterOnStop) {
+    ProtocolMessage(ByteArrayBuilder babProtocol) {
```
scottf, line 24: simplified constructors since all protocol messages are filtered on stop

PR #1502, opened 2025-12-16
Call to error listener should have been done in a callback, not directly

scottf:


src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1791,7 +1791,7 @@ void queueOutgoing(NatsMessage msg) {
             throw new IllegalArgumentException("Control line is too long");
         }
         if (!writer.queue(msg)) {
-            options.getErrorListener().messageDiscarded(this, msg);
+            makeCallback(() -> options.getErrorListener().messageDiscarded(this, msg));
```
scottf, line 1794: the error listener should never be called directly, it always should be done in a task, which makeCallback handles

aricart (approved):

PR #1504, opened 2025-12-17
[Replaced] Message Queue Splitting Pass 3

scottf:
Split the queue into it's 2 functions since the behavior and locking is significantly different. This makes it easier to properly lock makes each queue, WriterMessageQueue and ConsumerMessageQueue specific to their purpose.

* Tuning of locks
* WriterMessageQueue pause/filter combined since they are **always** called together.
* WriterMessageQueue filter optimized

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnectionWriter.java
+++ b/src/main/java/io/nats/client/impl/NatsConnectionWriter.java
@@ -262,10 +262,8 @@ class NatsConnectionWriter implements Runnable {
         return this.normalOutgoing.push(msg);
     }
 
-    void queueInternalMessage(NatsMessage msg) {
+    void queueInternalMessage(ProtocolMessage msg) {
         if (mode.get() == Mode.Reconnect) {
-            // we can offer directly because we know
-            // this is a protocol message
             reconnectOutgoing.pushReconnect(msg);
         }
         else {


src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1136,7 +1136,7 @@ void sendUnsub(@NonNull NatsSubscription sub, int after) {
         if (after > 0) {
             bab.append(SP).append(after);
         }
-        queueOutgoing(new ProtocolMessage(bab, true));
+        queueOutgoing(new ProtocolMessage(bab));
```
scottf, line 1139: The true used to mean filter out when the writer is stop (it calls pause then filter), There was one case, the CONNECT message that used false. But I'm certain this was wrong and now all protocol messages are filtered out of the queue on writer stop. More in another comment.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1173,7 +1173,7 @@ String getNextSid() {
 
     String reSubscribe(NatsSubscription sub, String subject, String queueName) {
         String sid = getNextSid();
-        sendSubscriptionMessage(sid, subject, queueName, false);
+        sendSubscriptionMessage(sid, subject, queueName, true);
```
scottf, line 1176: just making this an internal message since, it is in face an internal message.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1829,7 +1829,7 @@ void queueOutgoing(NatsMessage msg) {
         }
     }
 
-    void queueInternalOutgoing(NatsMessage msg) {
+    void queueInternalOutgoing(ProtocolMessage msg) {
```
scottf, line 1832: All internal messages were ProtocolMessages, so I just changed the type to reflect this.

src/main/java/io/nats/client/impl/NatsConnection.java
```diff
@@ -1847,7 +1847,7 @@ void deliverMessage(NatsMessage msg) {
 
             NatsDispatcher d = sub.getNatsDispatcher();
             NatsConsumer c = (d == null) ? sub : d;
-            MessageQueue q = ((d == null) ? sub.getMessageQueue() : d.getMessageQueue());
+            ConsumerMessageQueue q = ((d == null) ? sub.getMessageQueue() : d.getMessageQueue());
```
scottf, line 1850: This part of the connection code is for handing off messages to the subscription or dispatcher, both of which are "Consumer" hence the name ConsumerMessageQueue

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -51,11 +53,11 @@ enum Mode {
 
     private byte[] sendBuffer;
     private final AtomicInteger sendBufferLength;
-
-    private final MessageQueue normalOutgoing;
-    private final MessageQueue reconnectOutgoing;
     private final long reconnectBufferSize;
 
+    private final WriterMessageQueue normalOutgoing;
+    private final WriterMessageQueue reconnectOutgoing;
```
scottf, line 59: The queue in the writer got named WriterMessageQueue

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -75,17 +77,30 @@ enum Mode {
...
-        reconnectOutgoing = new MessageQueue(true, options.getRequestCleanupInterval(),
-            sourceWriter == null ? null : sourceWriter.reconnectOutgoing);
         reconnectBufferSize = options.getReconnectBufferSize();
+
+        if (sourceWriter == null) {
+            normalOutgoing = new WriterMessageQueue(
+                options.getMaxMessagesInOutgoingQueue(),
+                options.isDiscardMessagesWhenOutgoingQueueFull(),
+                options.getRequestCleanupInterval());
+            reconnectOutgoing = new WriterMessageQueue();
```
scottf, line 87: There are really 2 variants of the WriterMessageQueue, one for the normal user flow while connected and one for reconnecting, which only handles Protocol Messages and is never bounded. So I made different constructors.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -75,17 +77,30 @@ enum Mode {
...
+            // So just take the queues from the source writer!
+            // This avoids instance creation, copying settings and data so is faster
+            // They have already been filterOnStop
+            // resume() makes sure they are in running mode, since they
+            // were stopped and filtered already.
+            normalOutgoing = sourceWriter.normalOutgoing;
+            normalOutgoing.resume();
+            reconnectOutgoing = sourceWriter.reconnectOutgoing;
+            reconnectOutgoing.resume();
+        }
```
scottf, line 103: I used to do a whole transfer of the queues, but looking at the workflow when this happens, the old writer is shut down and the new writer is being created. Just take the queues from the old writer and give them to the new writer. se the comment in the code.

src/main/java/io/nats/client/impl/NatsConnectionWriter.java
```diff
@@ -112,9 +127,8 @@ Future<Boolean> stop() {
             running.set(false);
             startStopLock.lock();
             try {
-                this.normalOutgoing.pause();
                 this.reconnectOutgoing.pause();
-                this.normalOutgoing.filter(NatsMessage::isProtocolFilterOnStop);
+                this.normalOutgoing.pause();
```
scottf, line 131: the code always caused pause then filter. This is the only run code that does it. So I combined the calls. Also, the whole "function" being passed again was only one actual use, and it always filtered protocol messages. So all this is combined and in the pause implementation.

src/main/java/io/nats/client/Options.java
```diff
@@ -154,6 +154,14 @@ public class Options {
      */
     public static final Duration DEFAULT_REQUEST_CLEANUP_INTERVAL = Duration.ofSeconds(5);
 
+    /**
+     * Default amount of time to lock waiting to offer a message to the outgoing write queue
+     * A timer is used to clean up futures that were handed out but never completed
+     * via a message, {@link #getQueueOfferLockWait() getQueueOfferLockWait()}.
+     * <p>This property is defined as 5 seconds.</p>
+     */
+    public static final Duration DEFAULT_QUEUE_OFFER_LOCK_WAIT = Duration.ofSeconds(5);
```
MauriceVanVeen, line 163: Can these 'queue offer lock wait' additions in this file be removed, it seems they're not used?

The `MessageQueue` (and now `WriterMessageQueue`) takes `options.getRequestCleanupInterval()` as an argument. And that argument was renamed from `requestCleanupInterval` to `queueOfferLockWait` as well. Perhaps that could also be reverted?

PR #1511, opened 2026-01-06
Fix line ends

scottf:
Ran this command:
```
git add --renormalize .
```

To see that it's only line ends in the github diff, while in the [changes page of the PR](https://github.com/nats-io/nats.java/pull/1511/changes), check Hide Whitespace. All the diffs will be empty.

image


MauriceVanVeen (approved):
LGTM

PR #1514, opened 2026-01-09
Enable path component when connecting via WebSocket

jbaeck:
When a WebSocket connection url is specfied via ws://HOST:IP/PATH the PATH will now be included when establishing the connection.

I need help implementing a Test as I have not found a way to define the path in the nats websocket configuration.


scottf:
@jbaeck For the test, we would have to mock a proxy of some sort? For now, as long as all the websocket tests pass confirming nothing is broken, and you manually confirm via a comment that you have tested your new pathways, I can live with that.

jbaeck:
@scottf - all manual tests against my environment worked as well as the existing Tests. 
I am really happy that in the end it’s a small change ! 

Anything else I could look into ?

PR #1518, opened 2026-01-14
WriterMessageQueue won't be blocked if discardWhenFull=true

DeviLab:
After the refactoring in [#1505](https://github.com/nats-io/nats.java/pull/1505), `io.nats.client.impl.MessageQueue` was split into `WriterMessageQueue` and `ConsumerMessageQueue`. During this change, the blocking behavior of `WriterMessageQueue.push` was altered.

In the original implementation, when `discardWhenFull` was `true`, the internal queue’s `offer` method was called without a timeout, ensuring the calling thread would not block. In the new `WriterMessageQueue`, `offer` is always invoked with a timeout, regardless of the `discardWhenFull` flag.

This behavior is problematic in scenarios where blocking the calling thread is undesirable. The change appears to be unintentional, so this PR restores the original non-blocking behavior when `discardWhenFull` is enabled.

### DIFF
--- a/src/main/java/io/nats/client/impl/WriterMessageQueue.java
+++ b/src/main/java/io/nats/client/impl/WriterMessageQueue.java
@@ -55,20 +55,20 @@ class WriterMessageQueue extends MessageQueueBase {
             long startNanos = NatsSystemClock.nanoTime();
             if (editLock.tryLock(pushTimeoutNanos, TimeUnit.NANOSECONDS)) {
                 try {
-                    long timeoutNanosLeft = Math.max(
-                        MIN_PUSH_TIMEOUT_NANOS,
-                        pushTimeoutNanos - (NatsSystemClock.nanoTime() - startNanos)
-                    );
-                   if (queue.offer(msg, timeoutNanosLeft, TimeUnit.NANOSECONDS)) {
-                        sizeInBytes.getAndAdd(msg.getSizeInBytes());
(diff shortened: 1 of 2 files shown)


src/main/java/io/nats/client/impl/WriterMessageQueue.java
```diff
@@ -55,20 +55,20 @@ boolean push(NatsMessage msg, boolean internal) {
...
-                    );
-                   if (queue.offer(msg, timeoutNanosLeft, TimeUnit.NANOSECONDS)) {
-                        sizeInBytes.getAndAdd(msg.getSizeInBytes());
-                        length.incrementAndGet();
-                        return true;
-                    }
-                    // internal or !discardWhenFull throws instead of returns
                     if (!internal && discardWhenFull) {
-                        return false;
+                        if (!queue.offer(msg)) return false;
```
scottf, line 59: Always bracket an if statement body
```
if (!queue.offer(msg)) {
    return false;
}
```

scottf:
The was done to address a problem with extremely fast producers filling up, and this gives more time for the existing messages to be sent out. Also, there is already blocking time to get the edit lock and the check (which I suppose might be compiler optimized). The time adding to the queue is not nothing, it's minimal.

I considered making this behavior driven by a connection option, and may reconsider this. I will certainly consider putting it back without an option. Can you give me more insight into your use case?

PR #1554, opened 2026-04-24
Finds while building J3

scottf:


scottf:
@claude review

src/test/java/io/nats/client/impl/JetStreamManagementTests.java
```diff
@@ -924,6 +924,7 @@ public void testValidConsumerUpdates() throws Exception {
             assertValidAddOrUpdate(jsm, cc);
 
             cc = prepForUpdateTest(jsm);
+            System.out.println("\n!!!");
```
Jarema: A leftover of some debuggning?

Jarema (changes requested):
Generally LGTM, but I think there is some leftover print that needs cleaning up before merge.

src/test/java/io/nats/client/impl/JetStreamManagementTests.java
scottf: yeah, claude had caught that too. fixing.

scottf:
> #### 1. Debug `println` left in test — must remove before merging

FIXED

> #### 2. Stale comment in `_nextUnmanagedWaitForever()`

FIXED

> #### 3. No positive test for the newly-allowed zero backoff

FIXED

Jarema (approved):
LGTM

PR #1562, opened 2026-05-03
2.14 Reset Consumer

scottf:


scottf:
@claude please review

scottf:
@claude please re-review considering the following:

> **1. Copy-paste error in Javadoc `@return` descriptions** 

FIXED

> **2. Missing `/** {@inheritDoc} */` on the no-arg override** 

FIXED

> **3. Inconsistent charset in `getBytes()`** 

FIXED with StandardCharsets.ISO_8859_1 since this string is known to be ascii

> **4. `atLeast2_14` skips `atLeast2_13`**

N/A There is no 2.13

> **5. Test does not fetch after `resetConsumer(stream, consumer, 1)`** 

FIXED

scottf:
Fixed reset to Reset in api doc

philpennock (approved):
Is it expected that a sequence of 0 is non-default but has no effect? The default sentinel is -1 but the conditional is `< 1`.

PR #1569, opened 2026-05-07
Change claude to always review

scottf:
except for when specifically told not to


Jarema (approved):
LGTM!

Though I think eventually we should put it upstream and make it default option.

PR #1572, opened 2026-05-11
Add Client and Orbit section, drop duplicate Orbit link

Jarema:
Explains the split between the core nats.java client and the Orbit utility artifacts (https://github.com/synadia-io/orbit.java): what each is for, what goes where, and a small layering diagram. Removes the now-redundant one-line Orbit reference from the intro list and adds a ToC entry. Mirrors the section added in nats.rs#1580.

Signed-off-by: Tomasz Pietrek <tomasz@synadia.com>


scottf (approved):
LGTM

Issue #1581, opened 2026-06-11
Address generic Timeout or no response waiting for NATS JetStream server


scottf:
## Problem

On requests, timeouts are seen, but often the actual cause is disguised. Fix this problem, which is discussed further below. All changes will have to be backward compatible.

### What actually causes "Timeout or no response waiting for NATS JetStream server" on a JS publish

Traced against the jnats source in this tree (`nats.java/src/main/java/io/nats/client/impl/`).

### The mechanism — why it's a catch-all

The exception is thrown in one place, on one condition:

```java
// NatsJetStreamImpl.responseRequired (line 257)
Message responseRequired(Message respMessage) throws IOException {
    if (respMessage == null) {
        throw new IOException("Timeout or no response waiting for NATS JetStream server");
    }
    return respMessage;
}
```

So the *only* trigger is `respMessage == null`. That null is manufactured one level down, where **three distinct failure modes are collapsed into the same null**:

```java
// NatsConnection.requestInternal (line 1316-1325)
CompletableFuture<Message> incoming = requestFutureInternal(...);
try {
    return incoming.get(timeout.toNanos(), TimeUnit.NANOSECONDS);
}
catch (TimeoutException | ExecutionException | CancellationException e) {
    return null;   // <-- all three look identical upstream
}
```

- **TimeoutException** — the PubAck genuinely didn't arrive within the deadline.
- **CancellationException** — the future was `cancel()`-ed by something else (reconnect cleanup, 503-in-CANCEL-mode, periodic cleanup of timed-out futures).
- **ExecutionException** — the future was completed exceptionally.

The text says "Timeout" but two of the three paths are not timeouts at all.

### Important: for the *publish* path, a 503 is NOT hidden here

`publishSyncInternal` passes `CancelAction.COMPLETE`:

```java
// NatsJetStream.publishSyncInternal (line 157)
Message resp = makeInternalRequestResponseRequired(
    subject, merged, data, getTimeout(), CancelAction.COMPLETE, conn.forceFlushOnRequest);
return processPublishResponse(resp, options);
```

With `COMPLETE`, a 503 No-Responders status comes back as a *real message* (not a cancel):

```java
// NatsConnection.deliverReply (line 1460-1471)
if (msg.isStatusMessage() && msg.getStatus().getCode() == 503) {
    switch (f.getCancelAction()) {
        case COMPLETE: f.complete(msg); break;                 // <-- publish path
[...]

PR #1582, opened 2026-06-11
Advanced Request Behavior Option

scottf:
Address #1581 

This adds an opt-in connection option, `advancedRequestBehavior`, that surfaces *why* a request came back without a response instead of the generic timeout. The detail is a classified `RequestFailureReason` — one of `TIMEOUT`, `CONNECTION_CLOSING`, `NO_RESPONDERS`, `PROTOCOL_ERROR`, or `CANCELLED` — plus the connection status, the last protocol error, how long the request waited, and the original exception chained as the cause, so you can tell a real timeout apart from a reconnect, a permissions error, or no stream on the subject.

It applies to both request styles:
- **JetStream** requests (publish, stream/consumer management) throw a `RequestFailureException` in place of the generic `IOException("Timeout or no response waiting for NATS JetStream server")`.
- **Core** `Connection.request(...)` returns a `RequestFailureMessage` (which *is* a `Message`) instead of `null`, which you can inspect for the same detail.

The option is off by default; existing behavior is unchanged unless you enable in the connection options.

```java
try {
    js.publish("subject", data);
}
catch (RequestFailureException e) {
    // e.getReason(), e.getConnectionStatus(), e.getLastError(), e.getCause()
    log.warn("publish failed: {} (status={})", e.getReason(), e.getConnectionStatus());
}
```

For a core request, the message will never be null. It will either be the response message or will be a RequestFailureMessage. You must inspect the returned message:

```java
Message m = nc.request("subject", data, Duration.ofSeconds(2));
if (m instanceof RequestFailureMessage) {
    RequestFailureMessage rfm = (RequestFailureMessage)m;
    log.warn("request failed: {} (status={})", rfm.getReason(), rfm.getConnectionStatus());
}
```

The two paths are distinct: a JetStream request **throws** `RequestFailureException` (which extends `IOException`, so existing `catch (IOException)` blocks keep working), while a core `request(...)` **returns** the `RequestFailureMessage` — it does not throw.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,12 @@
 gnatsd.log
 *.csv
 examples/**/Z*.java
+**/examples/z/**
+**/examples/testapp_mc/**
 **/Debug*.java
+.claude/**
+env.bat
+**/TestDebugger*
 
 # Compiled source #
 ###################
--- a/README.md
+++ b/README.md
@@ -185,10 +185,10 @@ Behavior and Benchmark over 37 million passes...
 
 This release adds an opt-in connection option, `advancedRequestBehavior`, that surfaces *why* a request
 came back without a response instead of the generic timeout. The detail is a classified
-`RequestFailureReason` — one of `TIMEOUT`, `CONNECTION_CLOSING`, `NO_RESPONDERS`, `PROTOCOL_ERROR`, or
-`CANCELLED` — plus the connection status, the last protocol error, how long the request waited, and the
-original exception chained as the cause, so you can tell a real timeout apart from a reconnect, a
-permissions error, or no stream on the subject.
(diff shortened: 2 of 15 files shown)


scottf:
@claude review

scottf:
@claude review again

scottf:
@claude one last review. Here are comments on your last items
> **1. `cause = e.getCause()` has no null guard** (`NatsConnection.java` ~line 1335)
NON-ISSUE. In this exact case, cause will never be null, because an ExecutionException by definition is a wrapper around a cause.

> **2. `RequestFailureMessage`'s public constructor exposes an internal type** 
NO. NatsRequestCompletableFuture is public. It's packaging means nothing.

> **3. Wildcard import in `Discovery.java`** (`src/main/java/io/nats/service/Discovery.java`)
No, Claude, your style guide is wrong.

> **4. PR description code example is now stale**
Fixed

Jarema (approved):
LGTM!

PR #1583, opened 2026-06-12
Optional individual thread factories for reader and writer

scottf:


scottf:
@claude review

scottf:
@claude re-review with these notes:

> 1. "Restarted across reconnects" wording is slightly misleading

Simplified comments

> 2. Shared-Options scenario silently shares the reader/writer executor

Addressed with notes in readme since this is an options wide behavior

> 3. No ExecutorService-direct variant for reader/writer

This is intentional. An executor service would provide a pool.

> 4. Integration test could assert thread names more precisely

No, I think this is the correct behavior.

> 5. No reconnect-path coverage

There are many tests that test reconnect, so this review point is not correct.

> NatsConnection.readerExecutorIsClosed() / writerExecutorIsClosed() helpers are absent

Fixed

> resolvedReaderExecutor / resolvedWriterExecutor fields at Options.java:949-950 are not volatile. They're always accessed under executorsLock, so this is fine — just calling it out for clarity.

Correct. Clarity not needed.

scottf:
@claude added some coverage, please review

build.gradle
```diff
@@ -11,7 +11,7 @@ plugins {
     id("signing")
 }
 
-def jarVersion = "2.60.0"
+def jarVersion = "2.26.0"
```
philpennock, line 14: Is it expected that this has gone down?

philpennock (approved):

build.gradle
scottf: This was a mistake and is being fixed here.

Issue #1579, opened 2026-06-11
Provide KV Configuration builder from Json


abishai:
### Proposed change

When working with jetstream, usually, we place stream configs in application resource directory. 
We can upsert a stream with 
```
StreamConfiguration.instance(jet.toJson())
```
, where jet is Json representation of corresponding stream config. This approach It is very easy and straightforward. 
We've tried KV to see if we can replace Redis, but KV Configuration doesn't have similar builder and primary constructor is package private. 

### Use case

Storing stream config in application resource directory. Since configuration can be saved as Json, reverse mechanism would be nice.

### Contribution

The change is very trivial, primary question if this proposal is acceptable at all.

scottf:
I think this is a reasonable thing to do, especially considering that it can be done for StreamConfiguration. I assume that you have your own wrapper or do this in a class in `io.nats.client.api` since `StreamConfiguration.instance` is package scoped.

Are you interested in giving this a go and making a PR? I think it is important to be able to validate this by calling build() (the same pattern in `StreamConfiguration.instance`)

abishai:
There is public method https://github.com/nats-io/nats.java/blob/main/src/main/java/io/nats/client/api/StreamConfiguration.java#L175 and I thought about something similar.

edit: I see, why there is no such feature - it would require rather big copy-paste from StreamConfiguration

PR #1585, opened 2026-06-16
Builder KV and OS Configuration from JSON

scottf:
Addressing https://github.com/nats-io/nats.java/issues/1579


scottf:
@claude review as if the first time. Do not care about wildcard imports

scottf:
@claude re-review

PR #1590, opened 2026-07-07
Tidy executor options and additional testing

scottf:
- Comments were added around code
- Some code changes are cosmetic, meaning I just reorganized the order of variables.
- A test was improved to not flap
- A test was improved to provide complete coverage of a recently added feature.
- Improvement (Claude identified as a bug, but really wasn't because it was a meaningless set of a variable. It was noticed in a review as an improvement to the readability of the code)

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -941,12 +941,12 @@ public class Options {
 
     private final ReentrantLock executorsLock;
 
-    private final ExecutorService userExecutor;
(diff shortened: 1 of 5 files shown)


scottf:
@claude review

scottf:
> ⚠️ Minor: double-close in try-with-resources (safe but slightly unusual)
> ⚠️ Minor: no test for the ordered-consumer next() fix

These were addressed

philpennock (approved):

PR #1592, opened 2026-07-13
Reset Reader Connection

scottf:
Allows a `NatsConnectionReader` to be repointed to a different `NatsConnection` without exposing internals.

- `NatsConnectionReader.connection`: `final` → `private volatile`, set via a new `setConnection(...)` so the reader thread sees the new target on its next read.
- New `protected NatsConnection.setReaderConnection(NatsConnection)` is the entry point; the reader's field stays private.
- The `ReadListener` is re-derived from the new connection's `Options`, rebuilt only when it actually changed.

The repoint is a single volatile write — visible to the reader thread but not atomic with respect to an in-flight message parse; the threading contract is documented on `setConnection`.

### DIFF
--- a/src/main/java/io/nats/client/Options.java
+++ b/src/main/java/io/nats/client/Options.java
@@ -501,7 +501,7 @@ public class Options {
      */
     public static final String PROP_RECONNECT_DELAY_HANDLER_CLASS = PFX + "reconnect.delay.handler.class";
     /**
-     * Property used to configure a builder from a Properties object. {@value}, see {@link Builder#reconnectDelayBehavior(ReconnectDelayBehavior)
+     * Property used to configure a builder from a Properties object. {@value}, see {@link Builder#reconnectDelayBehavior(Options.ReconnectDelayBehavior)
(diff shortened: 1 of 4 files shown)


scottf:
@claude review

scottf:
@claude. Addressed comments. Please re-review

mtmk (approved):
LGTM

scottf:
## Review — Reset Reader Connection

Core mechanism looks solid. Making `NatsConnectionReader.connection` a private `volatile` that's repointed through `setConnection(...)`, with a thin `protected NatsConnection.setReaderConnection(...)` wrapper, keeps the field encapsulated while letting a same-package caller redirect a live reader. Switching it from `final` to `volatile` is the correctness-critical bit, since the reader thread reads it on every delivered message. The `refreshReadListener` re-derivation (rebuild only when the configured `ReadListener` actually changed; the wrapper reads the current connection from the volatile field at dispatch) correctly avoids dispatching trace callbacks on a stale connection, and the invariant comments on `setConnection` are thorough.

PR #1594, opened 2026-07-13
Simplify set reader connection

scottf:
Claude left in a extra method that I don't need. The test didn't test it anyway


mtmk (approved):
LGTM

PR #1595, opened 2026-07-13
Count connect failure once per server, not once per resolved IP

stasimus:
## Problem

The `maxReconnects` budget is consumed per **resolved IP address** instead of per **logical server**. `serverPool.connectFailed(cur)` is called inside the loop over the addresses that a server's hostname resolves to — in both the initial connect loop (`tryToConnect` callers in `connectImpl`) and `reconnectImplConnect`. Since `NatsServerPool.connectFailed` increments `failedAttempts` and drops the server from the pool once it reaches `maxConnectAttempts`, a hostname resolving to N addresses exhausts the budget roughly N times faster than a single-address server.

## How to reproduce

1. Configure a client with `maxReconnects(M)` pointing at a single unreachable server whose hostname resolves to N > 1 IP addresses (common with round-robin DNS, Kubernetes headless services, or dual-stack A + AAAA records).
2. Let the connection fail and observe the number of connection attempts before the client gives up.

## Current behavior

Each failed attempt against each resolved address counts against the server's `maxReconnects` budget. With N resolved addresses the pool removes the server after only about `M / N` failed rounds — e.g. `maxReconnects(4)` against a hostname resolving to 3 addresses gives up after roughly one round of reconnect attempts instead of four.

## Expected behavior

One full round over all of a server's resolved addresses counts as **one** failed attempt against that server. With `maxReconnects(M)` the client performs M reconnect rounds regardless of how many addresses the hostname resolves to; the total number of socket-level attempts is `N × M` (plus the initial round).

## Fix

Move the `serverPool.connectFailed(cur)` call (and the `failList` bookkeeping in the connect path) out of the per-resolved-address loop, so it runs once after a whole round over the server's addresses has failed. It is guarded by `!isConnected() && !isClosed()` so a success or a close mid-round doesn't record a failure.

## Why the unit test proves the fix

`MaxReconnectResolvedIpsTest` removes both the network and DNS from the equation so the attempt count is fully deterministic:

* A custom `NatsServerPool` overrides `resolveHostToIps` to return a fixed list of addresses for the configured hostname — simulating multi-address DNS without any actual resolution.
* A custom `DataPort` counts every `connect` invocation and always throws `IOException` — every attempt fails, so the client must walk through its entire reconnect budget.
* `Nats.connectReconnectOnConnect` makes the initial connect share the reconnect budget and exhaust it synchronously, so by the time the call returns a `CLOSED` connection all attempts have happened.

`testMaxReconnectNotDividedByResolvedIpCount` uses `maxReconnects = 4` with 3 resolved addresses and asserts **exactly** `3 × (4 + 1) = 15` connect attempts. Before the fix, `connectFailed` fired once per address per round (3× per round), so the budget of 4 was exhausted after about one round and the count came out far below 15 — the assertion fails on the old code and passes with the fix. The strict equality also proves the fix doesn't over-count: more than 15 would mean the budget stopped being honored.

`testSingleIpBaselineUnchanged` runs the same scenario with 1 resolved address and asserts `4 + 1 = 5` attempts, proving the single-address behavior (where per-address and per-server counting coincide) is unchanged by the fix.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -281,14 +281,16 @@ class NatsConnection implements Connection {
                 timeTraceLogger.trace("setting status to disconnected");
                 updateStatus(Status.DISCONNECTED, resolved, cur);
 
-                failList.add(cur);
-                serverPool.connectFailed(cur);
-
                 String err = connectError.get();
 
                 if (this.isAuthenticationError(err)) {
                     this.serverAuthErrors.put(resolved, err);
                 }
+            }
+
+            if (!isConnected() && !isClosed()) {
+                failList.add(cur);
+                serverPool.connectFailed(cur);
             }
         }
 
@@ -482,7 +484,6 @@ class NatsConnection implements Connection {
                     return;
                 }
 
-                serverPool.connectFailed(cur);
                 String err = connectError.get();
                 if (this.isAuthenticationError(err)) {
                     if (err.equals(this.serverAuthErrors.get(resolved))) {
@@ -490,6 +491,10 @@ class NatsConnection implements Connection {
                     }
                     serverAuthErrors.put(resolved, err);
                 }
+            }
+
+            if (!isConnected() && !isClosed()) {
+                serverPool.connectFailed(cur);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


scottf:
@stasimus Your fix looks good. Can you please update your branch.

scottf:
@claude review

stasimus:
updated

scottf:
@stasimus claude called out `CONNECT_ATTEMPTS` please review that and make changes if appropriate

stasimus:
@scottf I did apply minor change for suggestion: added `@ResourceLock` to both tests so they stay serialized even if parallelism is enabled
  later." But since the annotation is one line per test and eliminates the risk outright, applying it seemed better than acknowledging it.

scottf (approved):
LGTM

PR #1597, opened 2026-07-14
ScheduledTask: survive tick exceptions and report them to ErrorListener

stasimus:
## Problem

When any exception is generated inside `io.nats.client.support.ScheduledTask#run`, it stays invisible — without any logging imprints — and it stops the scheduled task from being rescheduled afterward, silently losing all upcoming schedules.

## Expected behaviour

1. Log the error.
2. Keep the scheduled trigger for next tasks if the exception is recoverable.

## Proof that it won't be rescheduled

`ScheduledFutureTask.run()` calls `runAndReset()`:

```java
protected boolean runAndReset() {
    if (state != NEW || !RUNNER.compareAndSet(this, null, Thread.currentThread()))
        return false;
    boolean ran = false;
    int s = state;
    try {
        Callable<V> c = callable;
        if (c != null && s == NEW) {
            try {
                c.call();           // note: result is NOT recorded
                ran = true;
            } catch (Throwable ex) {
                setException(ex);   // state: NEW -> COMPLETING -> EXCEPTIONAL
            }
        }
    } finally { ... }
    return ran && s == NEW;
}
```

- On success, `c.call()` returns, `ran = true`, and — unlike plain `run()` — the result is never stored, so the state stays `NEW`. That's the "reset": a `FutureTask` is reusable only while its state is `NEW`, and `runAndReset` deliberately avoids consuming it. Returns `true`.
- On exception, the `catch (Throwable ex)` calls `setException(ex)`, which irreversibly transitions the state to `EXCEPTIONAL` (`FutureTask` states are one-way; there's no path back to `NEW`). `ran` stays `false`, so it returns `false`. The exception is now stored in the future — retrievable via `get()`, never rethrown to the worker thread.

Then `ScheduledFutureTask.run()` consumes that boolean:

```java
else if (super.runAndReset()) {
    setNextRunTime();
    reExecutePeriodic(outerTask);   // only on true!
}
```

So the suppress-forever behavior is just: exception → `runAndReset()` returns `false` → the task is silently not re-enqueued, and since its state is now `EXCEPTIONAL` it couldn't run again even if it were (the `state != NEW` guard at the top would bail). Nothing is logged, no thread dies — the throwable sits in the future waiting for a `get()` that, in the fixed-rate pattern, nobody ever calls.

### DIFF
--- a/src/main/java/io/nats/client/impl/MessageManager.java
+++ b/src/main/java/io/nats/client/impl/MessageManager.java
@@ -155,7 +155,8 @@ abstract class MessageManager {
                         if (sinceLast > alarmPeriodSettingNanos.get()) {
                             handleHeartbeatError();
                         }
-                    })
+                    },
+                    conn::reportScheduledTaskException)
             );
 
             // since we just scheduled, reset this otherwise it may alarm too soon
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -648,13 +648,14 @@ class NatsConnection implements Connection {
                                 // it's running in a thread, there is no point throwing here
                             }
                         }
-                    });
+                    }, this::reportScheduledTaskException);
                 }
 
                 long cleanMillis = this.options.getRequestCleanupInterval().toMillis();
 
                 if (cleanMillis > 0) {
-                    cleanupTask = new ScheduledTask(scheduledExecutor, cleanMillis, () -> cleanResponses(false));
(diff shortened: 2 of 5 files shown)


scottf:
It is true that the task won't reschedule. This matches the behavior for:

`ScheduledExecutorService.scheduleAtFixedRate`

```
* If any execution of the task
* encounters an exception, subsequent executions are suppressed.
```

So if the user's `run()`, which can only throw a runtime exception fails, that's the user's problem and they should have handled it and logged their own issue.

```
public interface Runnable {
    public abstract void run();
}
```

Issue #1596, opened 2026-07-13
Ordered consumer create failures throw unrelated exception


ghost:
### Observed behavior

```
java.lang.IllegalStateException: This subscription is inactive.
	at io.nats.client.impl.NatsSubscription.unsubscribe(NatsSubscription.java:183)
	at io.nats.client.impl.NatsJetStreamSubscription.unsubscribe(NatsJetStreamSubscription.java:30)
	at io.nats.client.impl.NatsJetStreamImpl._createConsumerUnsubscribeOnException(NatsJetStreamImpl.java:142)
	at io.nats.client.impl.NatsJetStream.createSubscription(NatsJetStream.java:507)
	at io.nats.client.impl.NatsConsumerContext.subscribe(NatsConsumerContext.java:126)
	at io.nats.client.impl.NatsMessageConsumer.doSub(NatsMessageConsumer.java:158)
	at io.nats.client.impl.NatsMessageConsumer.<init>(NatsMessageConsumer.java:59)
	at io.nats.client.impl.NatsConsumerContext.consume(NatsConsumerContext.java:342)
	at io.nats.client.impl.NatsConsumerContext.consume(NatsConsumerContext.java:304)
	at io.nats.client.impl.NatsOrderedConsumerContext.consume(NatsOrderedConsumerContext.java:133)
```

### Expected behavior

Actual create failure exception is thrown

### Server and client version

nats.java 2.26.0

### Steps to reproduce

Create an ordered consumer that times out.

scottf:
I get the create part. Not sure about the complex filter part. Is this repeatable? Is there any server information, stream config

scottf:
## Verdict

The reported symptom is real and is **already fixed on `main` / in 2.26.1** by commit `cb97bc8a` ("Better exception raising on create consumer", 2026-07-14 — the day after this issue was filed). It has no test and no release note, which is probably why this issue is still open.

The root cause underneath the symptom is a **connection close racing an in-flight consumer-create**, and there are three related soft spots that `cb97bc8a` does not cover. Details in the following comments.

## For the reporter

Two things, @jpsugar-flow:

1. **Upgrade to 2.26.1+.** The `IllegalStateException` no longer masks the real failure.
[...]

PR #1598, opened 2026-07-14
Better exception raising on create consumer

scottf:
When creating a consumer fails and is set to try to unsubscribe, if the unsubscribe fails, it's exception was raised instead of the original exception.

Addresses https://github.com/nats-io/nats.java/issues/1596


scottf:
@claude review

piotrpio (approved):
LGTM!

PR #1601, opened 2026-07-22
Address forceReconnectImpl reader/writer stop race

scottf:
forceReconnectImpl tears the socket down with reader.stop(false) + an async port close + a 100 ms join, then proceeds to reconnectImpl() regardless of whether that join succeeded. Three facts combine into a race:

reader.stop(false) sets running=false but does NOT shutdownInput() — the reader thread stays blocked in dataPort.read(...); only the port close wakes it, and that runs on another thread.
The stopped future completes only when run() actually returns, so the .get(100ms) is really "wait up to 100 ms for the reader thread to die" — and on a busy executor / TLS socket it can miss.
tryToConnect re-joins the reader only if (reader.isRunning()) (V3: core/.../NatsConnection.java:516), which is now false — so it does NOT wait for a still-alive reader before calling reader.start(...) on the same instance.
A reader that outlives the 100 ms window can then, once the async close finally unblocks it: throw IOException, see running flipped back true by the new start(), call handleCommunicationIssue on the now-healthy connection (spurious full reconnect), and — via its finally { running.set(false) } — stomp the shared flag the freshly started reader loops on. Timing-dependent, worse under executor pressure and TLS. See the audit for the full trace.

### DIFF
--- a/src/main/java/io/nats/client/impl/NatsConnection.java
+++ b/src/main/java/io/nats/client/impl/NatsConnection.java
@@ -350,6 +350,13 @@ class NatsConnection implements Connection {
         try {
             updateStatus(Status.DISCONNECTED);
 
+            // Stop the reader and writer BEFORE the port is closed below. stop(false) sets running=false
+            // without shutdownInput, so when the close unblocks a socket-blocked read the reader treats
+            // the resulting IOException as an expected shutdown rather than a communication issue that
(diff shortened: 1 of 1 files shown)


scottf:
@claude review

mtmk (approved):
LGTM

PR #1602, opened 2026-07-29
Instructions for Building a Canary

scottf:
A canary is a workflow to do something like test against this library. This is a non-production document, not intended for the average user, just documentation.


mtmk (approved):
LGTM

PR #1612, opened 2026-08-11
Fix Test Flappers

scottf:


philpennock (approved):
This still seems prone to flakiness, isn't there a fake clock which is appropriate to use here, tied into the thing being tested?

scottf:
> This still seems prone to flakiness, isn't there a fake clock which is appropriate to use here, tied into the thing being tested?

I think it's better, but will watch it to see if it's still flaky.

PR #1614, opened 2026-08-18
Fix pending queue lookup race

zileongggg:
# Summary
Fixes a race condition in jnats pending-queue accounting reported in [Synadia support ticket #2038](https://support.synadia.com/hc/en-us/requests/2038?brand_id=360006118134).
NatsConsumer#getPendingMessageCount() and getPendingByteCount() read getMessageQueue() twice. If a subscription is invalidated between those reads, the second call can return null, causing an NPE on the connection reader thread:
```
NatsConnectionReader.gatherMessageData
  → NatsConsumer.hasReachedPendingLimits
  → NatsJetStreamSubscription.getPendingMessageCount
 ```
 
This can surface as IOException: Gather Message Data and trigger an unnecessary disconnect/recovery cycle.

# Changes
- Cache the ConsumerMessageQueue reference before reading its message count or byte size.
- Return 0 when the queue is no longer available.
- Add regression coverage for a queue that becomes unavailable immediately after its first lookup.

### CODE REVIEW

**Comment by scottf** [comment]
**Copied from the original submitter.**

Stack trace from the affected system:

```
NATS exception occurred while connecting to URL tls://10.242.160.129:4222,tls://10.242.160.130:4222,tls://10.242.160.131:4222: Gather Message Data
java.io.IOException: Gather Message Data
    at io.nats.client.impl.NatsConnectionReader.gatherMessageData(NatsConnectionReader.java:395)
    at io.nats.client.impl.NatsConnectionReader.run(NatsConnectionReader.java:177)
    at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
    at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
    at java.base/java.lang.Thread.run(Thread.java:1474)
Caused by: java.lang.NullPointerException: Cannot invoke "io.nats.client.impl.ConsumerMessageQueue.sizeInBytes()" because the return value of "io.nats.client.impl.NatsConsumer.getMessageQueue()" is null
    at io.nats.client.impl.NatsConsumer.getPendingByteCount(NatsConsumer.java:94)
    at io.nats.client.impl.NatsJetStreamSubscription.getPendingByteCount(NatsJetStreamSubscription.java:30)
    at io.nats.client.impl.NatsConsumer.hasReachedPendingLimits(NatsConsumer.java:146)
    at io.nats.client.impl.NatsConnection.deliverMessage(NatsConnection.java:1841)
    at io.nats.client.impl.NatsConnectionReader.gatherMessageData(NatsConnectionReader.java:375)
    ... 6 common frames omitted
```

**Comment by scottf** [comment]
**Copied from the original submitter.**

We dug a bit deeper into the jnats library and have the following findings. We suspect that the disconnection is caused by a racing condition, as 1 thread is trying to access the incoming variable in NatsSubscription via getPendingMessageCount(), while another thread sets the incoming variable to null via invalidate().

**Thread 1 - NatsConnectionReader**

I believe this thread starts when the connection to NATS is established. The suspected call chain is as follows:

1. `NatsConnectionReader.start()` will start a background thread to read messages from the subscription `connection.getReaderExecutor().submit(this, Boolean.TRUE);`
2. `NatsConnectionReader.run()` will call `gatherMessageData()`
3. `gatherMessageData` calls `NatsConnection.deliverMessage()`
4. Before delivering the message, it will check whether the consumer has reached the pending limits by calling `NatsConsumer.getPendingMessageCount()`

**Thread 2 - NIO Application calls KeyValue.keys()**

1. Our application will call `KeyValue.keys()` periodically before performing a certain action. Current interval per call is 5 seconds.
2. `KeyValue.keys()` will call the `visitSubject()` method from `NatsFeatureBase`.
3. `NatsFeatureBase` has a `finally { sub.invalidate(); }` block, which sets the incoming variable in `NatsSubscription` to null.

**Hypothesis**

Thread 1 Step 4 is having a race condition with Thread 2 Step 3.

The offending class is suspected to be `NatsConsumer`:

```java
public long getPendingMessageCount() {
    return this.getMessageQueue() != null ? this.getMessageQueue().length() : 0;
}
```

scottf:
Replaced by #1615

PR #1619, opened 2026-08-25
Add a Windows build workflow

scottf:
Runs the full test suite on windows-latest against a nats-server built from the tip of main, on every push to main. Windows has never had CI coverage here, and several of the server bugs found through the Java tests over the years were specific to the Windows environment.

The shared synadia-io/workflows install action is bash and unix-only, so this workflow installs the server itself rather than making that action cross platform. It builds nats-server.exe and hands the absolute path to the test harness through the nats_server_path environment variable, which NatsRunnerUtils.getResolvedServerPath() reads - that keeps PATH and PATHEXT out of the picture on a platform where an extensionless binary is not resolvable as a command.

Java 8 only, matching the other builds. No jacoco, coveralls, javadoc or publishing: those are covered on Linux and would only add runtime and failure surface. Test reports upload as an artifact on failure, since triaging Windows failures from raw console output is painful.


mtmk (approved):
LGTM

PR #1623, opened 2026-09-03
Fix test to respect server 2.15 change to max_consumers behavior

scottf:
The CI uses the latest (head) server. 2.15 has a change where creating a consumer when a stream does not set the max_consumers limits the number of consumers to 1000. This fixes a test that specifically creates more than 1024 consumer to test the paging ability of the js management getConsumers call.


MauriceVanVeen (approved):
LGTM

philpennock (approved):

PR #1625, opened 2026-09-05
Fix drain() future completing true when the drain timed out

### LINKED ISSUE #1616: drain() future completes true even when the drain timed out
`Subscription.drain(timeout)` returns a future that completes with whether the drain succeeded. For subscriptions it always completes `true`, including when the drain timed out with messages still queued. Dispatchers are not affected.

## Mechanism

`NatsConsumer.drain(Duration)` runs its wait loop, then cleans up, then completes the future:

```java
while (NatsSystemClock.nanoTime() - startTime < timeoutNanos && !Thread.interrupted()) {
    if (this.isDrained()) {
        break;
    }
    Thread.sleep(1);
}

this.cleanUpAfterDrain();
...
finally {
    tracker.complete(this.isDrained());
}
```

`NatsSubscription.cleanUpAfterDrain()` calls `connection.invalidate(this)`, which nulls `incoming`. So the `isDrained()` inside the `finally` is always evaluated against a subscription whose queue is already gone, and:

```java
boolean isDrained() {
    return isDraining() && this.getPendingMessageCount() <= 0;
}
```

`getPendingMessageCount()` returns `-1` once the queue is gone, so the comparison is true no matter how much was left unread. The loop's own `isDrained()` calls are fine — those run before the cleanup — but their result is discarded.

This is long-standing rather than new. Before #1615 the accessor returned `0` for a missing queue instead of `-1`, so the comparison was `0 == 0` and the outcome was identical. #1615 changed the sentinel and kept the behaviour deliberately, to avoid flipping the contract as a side effect of a race fix.

Dispatchers are unaffected because `NatsDispatcher.incoming` is `final` and never nulled, so their count stays real.

## Suggested fix

Capture the answer while the queue still exists, and complete the future with that:

```java
boolean drained = false;
try {
    while (...) {
        if (this.isDrained()) {
            drained = true;
            break;
        }
        Thread.sleep(1);
    }
    this.cleanUpAfterDrain();
}
...
finally {
    tracker.complete(drained);
}
```

An explicit flag is better than re-deriving it from queue state, since the queue is deliberately torn down before the future completes.

## Why this needs its own change

It flips a public contract. Callers who today see `true` from every drain will start seeing `false` when a drain genuinely times out — which is the point, but it can surface as new failures in code that was silently relying on the old answer. It wants its own release note, and `DrainTests` should grow a case that drains with an unreachable server or a too-short timeout and asserts `false`.

arimu1:
Fixes #1616.

`Subscription.drain(timeout)` is supposed to complete its future with whether the drain actually finished. For subscriptions it currently completes `true` even when the wait timed out with messages still queued. Dispatchers are not affected.

## Why

`NatsConsumer.drain()` waits, then calls `cleanUpAfterDrain()`, then completes the future with `isDrained()`. `NatsSubscription.cleanUpAfterDrain()` invalidates the subscription and nulls `incoming`, so `getPendingMessageCount()` becomes `-1` and `isDrained()` is true regardless of how much was left unread.

#1615 kept `<= 0` on `isDrained()` on purpose so that race fix would not also flip this contract. This change is the follow-up that actually records the wait result.

## Change

Capture `isDrained()` while the queue still exists, then clean up, then complete the future with the captured value.

This is a public-contract fix: callers that always saw `true` from a subscription drain will now see `false` when the drain genuinely timed out.

## Test

`DrainTests.testSubDrainTimeoutWithQueuedMessagesCompletesFalse` publishes two messages, does not read them, drains with a 1s timeout, and asserts the future completes `false`.

On current main that assertion fails (`true`). With this change it passes. Neighbor drain tests still pass, including successful subscription drain, dispatcher drain, connection drain, and the existing connection-drain timeout case (`testSlowAsyncDuringDrainCanBeInterrupted`).

Ran locally: `DrainTests`, `NatsConsumerPendingTests`, `SlowConsumerTests` (JBR 17, nats-server 2.14.6).


scottf:
@claude review

scottf:
The code is correct as is. The PR version of the code locks in the future at the end of the timeout instead of waiting for it to be actually completed.

